PR #31, opened 2022-05-25
Add monkey patch for Cross-Origin Resource Policy Internal Check

blu25:
This PR introduces a monkey patch that modifies the CORP internal check to not short circuit allow for same-origin URLs if the request's destination is a fenced frame. This will ensure that fenced frames always act like cross-origin frames, regardless of the origin of its src attribute.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -181,4 +181,4 @@ Where it currently states
 
 Modify it to say
 
-> If origin is <a>same origin</a> with response's <a for=response>URL</a>'s <a for=url>origin</a>, and [=request/destination=] is not a fenced frame, then return allowed.
+> If origin is <a>same origin</a> with response's <a for=response>URL</a>'s <a for=url>origin</a>, and [=request/destination=] is not "<code>fencedframe</code>", then return <b>allowed</b>.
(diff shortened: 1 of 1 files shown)


domfarolino (changes requested):

spec.bs
```diff
@@ -165,3 +165,17 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
...
+
+<h3 id=coep>Patching Cross-Origin Resource Policy Internal Check</h3>
+
+In [[!Fetch]]'s <a href="https://fetch.spec.whatwg.org/#cross-origin-resource-policy-internal-check">cross-origin resource policy internal check</a> steps, in the step
+
+> 5. Switch on policy:
+
+Where it currently states
+
+> If origin is same origin with response’s URL’s origin, then return allowed.
```
domfarolino, line 177: Can we make sure to carry over any of the linking here? Like "same origin" and "URL's origin" have links in the fetch spec. Let's try and preserve them here.

domfarolino (changes requested):

domfarolino (approved):

spec.bs
```diff
@@ -165,3 +168,17 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
...
+
+> 5. Switch on policy:
+
+Where it currently states
+
+> If origin is <a>same origin</a> with response's <a for=response>URL</a>'s <a for=url>origin</a>, then return <b>allowed</b>.
+
+Modify it to say
+
+> If origin is <a>same origin</a> with response's <a for=response>URL</a>'s <a for=url>origin</a>, and [=request/destination=] is not a fenced frame, then return allowed.
```
domfarolino, line 184: > is not a fenced frame

Can you wrap "fenced frame" in quotes, or whatever is appropriate for destinations? I think we use quotes and <code> or something, like https://fetch.spec.whatwg.org/#request-destination-script-like

domfarolino (changes requested):
One more change, otherwise it looks good.

spec.bs
blu25: Looks like the convention is "<code>destination</code>", so I'll update this to match.

domfarolino (approved):

PR #32, opened 2022-05-25
Add monkey patch to add "fencedframe" request destination

blu25:
Add a new "fencedframe" request destination that has the same exact properties as the "iframe" request destination.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -36,6 +36,11 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: reflect; url: common-dom-interfaces.html#reflect
     text: width; url: embedded-content-other.html#attr-dim-width
     text: height; url: embedded-content-other.html#attr-dim-height
+spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
+  type: dfn
+    text: non-subresource requests; url: #non-subresource-request
+    text: navigation requests; url: #navigation-request
+    text: RequestDestination enum; url: #dom-requestdestination
 </pre>
 
 <style>
@@ -166,8 +171,14 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
 
-<h3 id=coep>New fencedframe [=request=] [=request/destination=]</h3>
+<h3 id=new-request-destination>New fencedframe [=request=] [=request/destination=]</h3>
 
-Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, <code>fencedframe</code>. It will have the initiator <code>""</code>, the CSP directive <code>child-src</code>, and the features <code>HTML's &lt;fencedframe&gt;</code>.
+Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, "<code>fencedframe</code>". It will have the [=request/initiator=] "", the CSP directive <code>fenced-frame-src</code>, and the features HTML's <code>&lt;fencedframe&gt;</code>.
+
+Add "<code>fencedframe</code>" to the list of [=non-subresource requests=] and to the list of [=navigation requests=].
+
+Add "<code>fencedframe</code>" to the [=RequestDestination enum=].
+
+
 
 Inherit the same properties and rules that the <code>iframe</code> destination currently has. 
--- a/spec.bs
+++ b/spec.bs
@@ -165,3 +165,9 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 <h3 id=dimension-attributes>Dimension attributes</h3>
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
+
+<h3 id=coep>New fencedframe [=request=] [=request/destination=]</h3>
+
+Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, <code>fencedframe</code>. It will have the initiator <code>""</code>, the CSP directive <code>child-src</code>, and the features <code>HTML's &lt;fencedframe&gt;</code>.
+
+Inherit the same properties and rules that the <code>iframe</code> destination currently has. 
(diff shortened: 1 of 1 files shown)


domfarolino:
You'll need to follow the instructions on https://labs.w3.org/repo-manager/pr/id/WICG/fenced-frame/32 to link your account up with the W3C. Could you go ahead and do that?

spec.bs
```diff
@@ -165,3 +165,9 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 <h3 id=dimension-attributes>Dimension attributes</h3>
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
+
+<h3 id=coep>New fencedframe [=request=] [=request/destination=]</h3>
+
+Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, <code>fencedframe</code>. It will have the initiator <code>""</code>, the CSP directive <code>child-src</code>, and the features <code>HTML's &lt;fencedframe&gt;</code>.
+
```
domfarolino, line 172: `<code>child-src</code>` should I think be `<code>fenced-frame-src</code>`. We'll need a monkeypatch section for this as well

spec.bs
```diff
@@ -165,3 +165,9 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 <h3 id=dimension-attributes>Dimension attributes</h3>
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
+
+<h3 id=coep>New fencedframe [=request=] [=request/destination=]</h3>
```
domfarolino, line 169: id=coep doesn't seem quite right here

spec.bs
domfarolino, line 172: I think "HTML's" should be outside of `<code>`

spec.bs
```diff
@@ -165,3 +165,9 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
...
+
+Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, <code>fencedframe</code>. It will have the initiator <code>""</code>, the CSP directive <code>child-src</code>, and the features <code>HTML's &lt;fencedframe&gt;</code>.
+
+Inherit the same properties and rules that the <code>iframe</code> destination currently has.
```
domfarolino, line 173: It would be good to be specific here. Maybe hunt around for usages of this and summarize why we need a new destination here

domfarolino (changes requested):

spec.bs
```diff
@@ -165,3 +165,9 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 <h3 id=dimension-attributes>Dimension attributes</h3>
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
+
+<h3 id=coep>New fencedframe [=request=] [=request/destination=]</h3>
+
+Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, <code>fencedframe</code>. It will have the initiator <code>""</code>, the CSP directive <code>child-src</code>, and the features <code>HTML's &lt;fencedframe&gt;</code>.
```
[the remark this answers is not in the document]
blu25: Looking at the convention, it looks like the quotes are usually outside of code blocks. So I think we can either just omit the code tag as you said and keep the "", or we can say "It will have no initiator". Whichever you think would be better.

spec.bs
blu25: Would we want both? fenced-frame-src seems similar to the [frame-src directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src), but that's not mentioned anywhere in the [fetch spec](https://fetch.spec.whatwg.org/). And I think if other frame types (like iframes) listen to the child-src directive, it would make sense for fenced frames to listen to it as well.

spec.bs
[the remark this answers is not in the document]
domfarolino: Having empty quotes sounds good

spec.bs
domfarolino: Ah I see. OK maybe just `child-src` here is fine, especially since we don't use frame-src here in the fetch spec. That is a little surprising to me though. I guess we're listing the closest-but-most-general directive that applies.

blu25: And we should still monkey-patch a fenced-frame-src directive somewhere as part of a different PR?

spec.bs
```diff
@@ -165,3 +171,17 @@ The {{HTMLFencedFrameElement/src}} IDL attribute must [=reflect=] the respective
 <h3 id=dimension-attributes>Dimension attributes</h3>
 
 This section details monkeypatches to [[!HTML]]'s <a href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes">Dimension attributes</a> section. This section will be updated to include <{fencedframe}> in the list of elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
+
+<h3 id=new-request-destination>New fencedframe [=request=] [=request/destination=]</h3>
+
+Every distinct element should have its own request destination to allow for special handling when making requests (fenced frame request behavior deviates from iframe behavior in enough ways to justify having a separate destination). Update the list of <a href="https://fetch.spec.whatwg.org/#concept-request-destination">associated destinations</a> to include a new entry, "<code>fencedframe</code>". It will have the [=request/initiator=] "", the CSP directive <code>fenced-frame-src</code>, and the features HTML's <code>&lt;fencedframe&gt;</code>.
```
domfarolino, line 177: Can we just link to https://github.com/whatwg/fetch/blob/main/fetch.bs#L1445 directly instead of using the `<a>`?

spec.bs
domfarolino: Yes. looks like you're taking care of that in another CL so I'll resolve this.

spec.bs
```diff
@@ -38,6 +38,11 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: height; url: embedded-content-other.html#attr-dim-height
 spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
+    text: non-subresource requests; url: #non-subresource-request
+    text: navigation requests; url: #navigation-request
```
domfarolino, line 42: already exported https://github.com/whatwg/fetch/blob/main/fetch.bs#L1896 (please check to see if there are any others like this here too)

spec.bs
```diff
@@ -38,6 +38,11 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    text: RequestDestination enum; url: #dom-requestdestination
```
domfarolino, line 43: Can you just use the IDL syntax to grab the destination? https://tabatkins.github.io/bikeshed/#autolink-wrappers

domfarolino (commented):
Looking good, almost there

domfarolino (approved):

PR #40, opened 2022-06-10
Revert "Add monkey patch for Cross-Origin Resource Policy Internal Check"

blu25:
Reverts WICG/fenced-frame#31

This is not necessary since the potential 1-bit leak that this is patching is not an issue. See this document for reasoning why:
https://docs.google.com/document/d/1wQTnfpksl8i-UxQ-1eoFvKtYIgv7RVCNnM42YbJVEYQ/edit?usp=sharing


domfarolino:
Can you make the doc link public?

PR #51, opened 2022-09-21
Use spec-prod instead of a Makefile to build the spec.

jyasskin:
This generally produces better error messages than curl.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/fenced-frame/pull/51.html" title="Last updated on Sep 21, 2022, 8:03 PM UTC (413a229)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/51/b829f69...jyasskin:413a229.html" title="Last updated on Sep 21, 2022, 8:03 PM UTC (413a229)">Diff</a>


domfarolino:
Looks like this PR deletes the Makefile? What is the preferred way to build the spec locally?

jyasskin:
I could undelete the Makefile if people are actively using it. I'd always install Bikeshed locally and run `bikeshed` or `bikeshed watch`.

domfarolino:
Hmm, I think `make remote` is pretty useful for people that don't have bikeshed installed, plus we still have the Makefiles in many WHATWG specs ([DOM](https://github.com/whatwg/dom/blob/main/Makefile#L4), [Fetch](https://github.com/whatwg/fetch/blob/main/Makefile#L4), [URL](https://github.com/whatwg/url/blob/main/Makefile#L4), etc) so it seems like good practice to keep it around?

jyasskin:
Works for me. I've restored it, with just the `make ci` target removed.

jyasskin:
(Part of the reason the WHATWG specs have Makefiles is that they have the COMMIT-SHA text macro, so raw `bikeshed` invocations actually fail. But I'm still fine with having Makefiles if the people actively developing a spec want them.)

.github/workflows/build.yml
```diff
@@ -1,22 +1,18 @@
...
-    - master
+  pull_request: {}
   push:
     branches:
     - master
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
domfarolino, line 10: Btw, can we just stick with latest or is there a reason to version this?

domfarolino (approved):

.github/workflows/build.yml
jyasskin: If a newer ubuntu is in some way incompatible with what spec-prod is doing, I think we want to upgrade intentionally. This matches the example in https://w3c.github.io/spec-prod/, and if spec-prod ever changes to need a newer OS, they can update their own version number to make sure we adopt both changes at the same time.

domfarolino (approved):

Issue #48, opened 2022-07-28
Proposal for changes to fenced frames urns/attributes


gtanzer:
Previously, we talked about urn:uuids mapping to URLs. But this is a bit imprecise, because there was actually other metadata associated with those URLs, like reporting metadata for FLEDGE `reportEvent()`, budget metadata for sharedStorage, etc.

This proposal makes this explicit, saying that urn:uuids map to a set of _attributes_, one of which is the `src` attribute. And we can add additional attributes like `width` and `height` that allow consumer APIs to have more control over and flexibility with information flow for their particular use case. See [this document](https://docs.google.com/document/d/1bTuFdm7_bdGAlBT8KBWIFg8Kdh7bOV1fE-zMQGAewyY/edit) for the full proposal.

Here is an existing issue about how the proposal affects FLEDGE: https://github.com/WICG/turtledove/issues/312

We welcome feedback. 🙂

gtanzer:
As a follow-up to the above proposal:

Previously, we conceptualized opaque identifiers as a means to map only a `src`. Therefore, it made sense to set them with the `src` attribute, and have them be strings (urn:uuids).

Given that these opaque identifiers have become and are continuing to become more expansive (describing more about the resulting fenced frame’s characteristics than just the resource URL), we don’t think it makes as much sense anymore to set them using the `src` attribute. Therefore, we are considering setting these opaque identifiers— which represent a “configuration”, “intent”, “set of properties”, or some other name subject to bikeshedding—using a new member of HTMLFencedFrameElement which is a WebIDL object. Because this new property need not be a string, we are also considering making the opaque identifier the WebIDL object’s reference (where some of the fields are opaque, as privacy requires for each use case).

For some examples of how this might look in practice:

```
// FLEDGE.
let frame = document.querySelector('fencedframe');
let config = await navigator.runAdAuction({ width: 100, height: 200, … });
frame.width = config.contentWidth;
frame.height = config.contentHeight;
frame.config = config;

// sharedStorage
let frame = document.querySelector('fencedframe');
let intent = await sharedStorage.selectURL({ width: 100, height: 200 },
[
  {url: '/url1', ...},
  {url: '/url2', ...}
]);
frame.width = config.contentWidth;
frame.height = config.contentHeight;
frame.config = config;

// Direct use (mostly for testing).
let frame = document.querySelector('fencedframe');
frame.config = new FencedFrameConfig({ url: 'https://www.example.com/' });
```

[...]

PR #56, opened 2022-12-08
Update explainer to talk about configs and use cases, rather than urn:uuids and modes

gtanzer:
Integrate proposed changes from https://github.com/WICG/fenced-frame/issues/48, and in relation to https://github.com/WICG/turtledove/issues/312


explainer/README.md
```diff
@@ -85,15 +85,10 @@ Since fenced frames are embedded frames, they also behave like iframes in many w
 
 
 ```
-<fencedframe src="demo_fenced_frame.html"></fencedframe>
+var fencedframe = document.createElement('fencedframe');
```
domfarolino, line 88: nit: const :)

explainer/README.md
```diff
@@ -85,15 +85,10 @@ Since fenced frames are embedded frames, they also behave like iframes in many w
...
+fencedframe.config = FencedFrameConfig({'src': 'demo_fenced_frame.html'});
```
domfarolino, line 89: This has to be `new FencedFrameConfig(...`

domfarolino, line 89: Right now the [ctor only takes a URL](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/fenced_frame_config.idl;l=16;drc=d99163d64d43b85a9efcf0110a0cfa81886b4616), so maybe we should reflect that here?

Or if we really want to be able to pass in a dictionary of things, we need to make that implementation change now, before we roll any of these APIs out to the wild. If that's our intention, then we can leave this here and make an APIFreeze2-blocking task to update the IDL any and tests ASAP.

explainer/README.md
```diff
@@ -135,18 +130,18 @@ This discussion assumes that third-party cookies, like all other third party sto
...
 If the process of getting an ad in the page is complex or expensive, there needs to be a way to ensure that the resulting ad will actually end up in the page before the expensive process begins.
 
-A static API method will be introduced to the HTMLFencedFrameElement class to check this. No fenced frame will be created when calling this API, and it can be invoked before actually attempting to create a fenced frame. The API will return a boolean, true if an opaque-ads fenced frame would be able to load in the caller's context, false if not.
+A static API method will be introduced to the HTMLFencedFrameElement class to check this. No fenced frame will be created when calling this API, and it can be invoked before actually attempting to load a fenced frame config. The API will return a boolean, true if the config would be able to load in the caller's context, false if not.
 
 ##### Example usage
 
 ```
-HTMLFencedFrameElement.canLoadOpaqueURL();
+HTMLFencedFrameElement.canLoadConfig(config);
```
domfarolino, line 144: Right now the API is still `canLoadOpaqueURL()`. Should we make a reference to that, saying that `canLoadConfig()` will replace it in the future, and in the meantime both will exist?

explainer/README.md
```diff
@@ -180,11 +175,11 @@ More about security mechanisms are detailed in:
 The fenced frame’s main goal is to improve privacy by disallowing communication with the embedder. There are however some attributes that might need to be shared between the two and their privacy impact needs to be carefully considered and mitigated, if possible. Some of these attributes are:
 
 
-*   **Initial size and resize:** To avoid the size attribute being used to communicate user identifying information from the embedding context to the fenced frame, this will be limited to only a few values. E.g. Some of the values that are relevant for ads for the "opaque-ads" mode. We are also considering allowing some of these sizes to be flexible based on the viewport width. Note that since size is a channel, these ads cannot be resized by the publisher. 
+*   **Initial size and resize:** The API that generates a fenced frame config can pick the initial size that the fenced frame document sees, subject to whatever restrictions it deems necessary for its privacy model. If the initial size is fixed, then any changes the embedder makes to the `width` and `height` attributes of the `<fencedframe>` will not be reflected inside the fenced frame.
```
domfarolino, line 178: I might just say "any changes the embedder makes to the width and height of the `<fencedframe>` will not [...]" instead of calling out the attributes, since the attributes are just one of many ways to resize the element

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
+# Fenced frame configs
+
+
+# Introduction
+
+For use cases involving APIs that access cross-site data, we need to be able to load a fenced frame with content determined by the API without revealing information about the content to the embedding context. For example, with interest-based ads in [FLEDGE](https://github.com/WICG/turtledove), the winning ad that's returned from the auction auction depends on the user's cross-site interest group data, which we don't want to expose to the site that calls the auction. This document proposes a web-platform way of loading content into a fenced frame using an opaque object.
```
domfarolino, line 6: "auction auction" should just be "auction"

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+# Proposed solution
+
+`<fencedframe>` has an attribute `config`, rather than `src`. APIs like FLEDGE return a WebIDL object called `FencedFrameConfig`, which has a series of fields that specify the behavior desired by the API (e.g. the ad url, width and height as seen from within the fenced frame, etc.). When the embedder stores this object into the `config` attribute, the fenced frame loads the context accordingly.
```
domfarolino, line 11: It's really just a JS object, so I'd probably go for "return a `FencedFrameConfig` object defined by [our WebIDL](https://wicg.github.io/fenced-frame/#fenced-frame-config-interface)". The spec is obviously not finished, but that's the link to use if we wanted to add one (I think it's a good idea).

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
...
+  
+In order to hide information as described above, the browser _redacts_ `FencedFrameConfig` before sending it to the embedder. This means that certain fields which are sensitive, like the ad url, are replaced with an enum value `opaque`. The embedder may see whether there is a value defined for that field, but not what the value is. Likewise, when the embedder requests that a config be loaded into the fenced frame, the browser is responsible for looking up the config in a data structure in order to access the unredacted information.
```
domfarolino, line 13: "enum value" => "string", since that's all it is in JS, and all that developers would ever care about

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+```
+navigator.runAdAuction(myAuctionConfig).then((auctionWinnerConfig) => {
+  // auctionWinnerConfig value e.g.
+  // FencedFrameConfig {
+  //   'src': opaque ('ad.com/foo' internally)
+  //   ...
+  // }
+  var adFrame = document.createElement('fencedframe');
```
domfarolino, line 28: nit: const

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
...
+  adFrame.setAttribute(config,auctionWinnerConfig);
```
domfarolino, line 29: Two issues here:
  1. You have a comma accessor instead of a period
  2. `setAttribute()` won't actually work here since the `config` attribute is not a content attribute (string), but _only_ a WebIDL attribute, which are only settable via JS. You can tell this because we don't have `[Reflect]` in our WebIDL: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/html_fenced_frame_element.idl;l=10;drc=d99163d64d43b85a9efcf0110a0cfa81886b4616

explainer/interaction_with_content_security_policy.md
```diff
@@ -32,7 +32,7 @@ Note that this is not applicable to opaque urls which are discussed in the next
 
 # **Opaque URL**
 
-The first mode of fenced frames is going to be for the [opaque ads use cases](https://github.com/shivanigithub/fenced-frame/blob/master/explainer/modes.md#opaque-ads) and since the url of the fenced frame needs to be hidden from the embedding page, we cannot apply the frame-src directive directly to the mapped url, since CSP reporting can reveal to the embedding page if a fenced frame was successfully created or not. This would lead to the embedding page deriving the possible origins of the opaque ad, going against the isolation guarantees of this mode.
+The first use cases for fenced frames are going to be for [opaque ads](https://github.com/shivanigithub/fenced-frame/blob/master/explainer/use_cases.md#opaque-ads) and since the url of the fenced frame needs to be hidden from the embedding page, we cannot apply the frame-src directive directly to the mapped url, since CSP reporting can reveal to the embedding page if a fenced frame was successfully created or not. This would lead to the embedding page deriving the possible origins of the opaque ad, going against the isolation guarantees of this mode.
```
domfarolino, line 35: Let's fix up the `shivanigithub` link here. Also we can make it relative if you'd like.

explainer/fenced_frame_config.md
dmdabbs: The FF Opaque Source [explainer](https://github.com/WICG/fenced-frame/blob/master/explainer/opaque_src.md) example shows the auction result promise resolving to an opaque url...

```
navigator.runAdAuction(myAuctionConfig).then((auctionWinnerUrl) => {
  // auctionWinnerUrl value e.g.
  // urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6
  var adFrame = document.createElement('fencedframe');
  adFrame.setAttribute(src,auctionWinnerUrl);
});
```

One chooses this path 'upstream' by returning a simple `'render': url` from `generateBid()`.


And the means to opt into the new way is to return 

`'render': {url: renderUrl, size: {width: renderWidth, height: renderHeight}},`

and then use 

```
navigator.runAdAuction(myAuctionConfig).then((auctionWinnerConfig) => {
  const adFrame = document.createElement('fencedframe');
  adFrame.config = auctionWinnerConfig;
});
```

Is that correct?

domfarolino: > and then use
> `navigator.runAdAuction(myAuctionConfig).then((auctionWinnerConfig) => { [.....]`

Actually the way you'll get FLEDGE's promise to resolve a `FencedFrameConfig` (instead of a URN) is by having `resolveToConfig: true` _inside_ the `myAuctionConfig` (that you pass into `runAdAuction()`). That's how you communicate your preference to deal in terms of configs (instead of URNs) to FLEDGE. This behavior isn't enabled yet though, so if you do it today you'll still get a URN. In fact, even once the behavior is enabled, it will only be rolled out gradually to more users for testing, so if you pass in `resolveToConfig` there will be a transition period where FLEDGE _may_ still give you a URN. Long-term, FLEDGE will always resolve to a `FencedFrameConfig` with or without `resolveToConfig: true` in the auction configuration.

The reason all of this is not super clear yet is because I have not made the PR to the FLEDGE repository which describes exactly how to take advantage of this, but I'll be doing that soon.

----

Regarding your first question:

> And the means to opt into the new way is to return
> `'render': {url: renderUrl, size: {width: renderWidth, height: renderHeight}},`

I am not sure about this (not as familiar with FLEDGE). @gtanzer do you know?

gtanzer: Yes, the combination of those things is correct. In order to hook into the new size behavior in FLEDGE, you will need to  declare ad sizes in the interest group declaration, make a bid that includes both url and size, put `resolveToConfig` in the auction config, and then store the winning config in `.config` as @dmdabbs's first example shows. (You can't/don't need to modify a config that gets returned from FLEDGE; it already has everything baked in.)

gtanzer: Actually, it may not be necessary to use configs to hook into the FLEDGE size changes: we would just need to do some extra implementation work to make urns work. We'll get back to you on that.

dmdabbs: An IG ad's size metdata may be [percentages](https://github.com/WICG/turtledove/pull/417/commits/d63d8e13537e697c93c52f7929da3cc06cb9ce4e?short_path=d65ba97#diff-d65ba9778fe3af46de3edfce2266b5b035192f8869280ec07179963b81f4e624), 

>...screen dimension coordinates (e.g. 100sw or 100sh)

The penny hasn't dropped for me how the internal size (I forget your terminology) is set and the external (page-visible dimensions). 

Also how pubs/sellers deal with the URN->Newfangled Config rollout/transition.

explainer/README.md
gtanzer: Changed the constructor to pass a URL only.

explainer/fenced_frame_config.md
gtanzer: @domfarolino Re the original comment, updated to `adFrame.config = auctionWinnerConfig;`

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
...
+## Turtledove Example
+
+When the SSP JS invokes the Turtledove API to run the ad auction, it gets back the `FencedFrameConfig` as the result, which is then used for rendering the fenced frame. This `FencedFrameConfig` has an opaque `src`, which maps to an actual ad url which is part of the interest group.
+
+
+```
+navigator.runAdAuction(myAuctionConfig).then((auctionWinnerConfig) => {
+  // auctionWinnerConfig value e.g.
+  // FencedFrameConfig {
+  //   'src': opaque ('ad.com/foo' internally)
```
domfarolino, line 25: Could you wrap "opaque" in quotes, since that is how it will be represented in the platform?

explainer/README.md
domfarolino: We discussed maybe removing this section from the explainer for now. Do you still plan on doing that?

explainer/fenced_frame_config.md
```diff
@@ -0,0 +1,35 @@
+# Fenced frame configs
+
+
+# Introduction
```
domfarolino, line 4: nit: I might make this a `##`, along with `Proposed solution`. And then `## Turtledove Example` would be `### Turtledove Example`.

domfarolino (approved):
Generally LGTM with a few comments.

explainer/README.md
gtanzer: I semi-reverted the changes here. Now it describes the same HTMLFencedFrameElement.canLoadOpaqueURL() function, but as "would a fenced frame config with an opaque mapped url be able to load in this context"?

We can fully replace it when the replacement is fully determined/implemented.

PR #67, opened 2023-04-07
Elaborate on FencedFrameConfig

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/67.html" title="Last updated on Apr 17, 2023, 7:17 PM UTC (208b004)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/67/123ccb0...208b004.html" title="Last updated on Apr 17, 2023, 7:17 PM UTC (208b004)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -687,36 +687,38 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 <div algorithm>
   The <dfn method for=Fence>reportEvent(<var ignore>event</var>)</dfn> method steps are:
 
-  1. Let |instance| be this [=traversable navigable=]'s [=fenced frame config instance=].
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node
+     navigable=]'s [=navigable/traversable navigable=]'s [=fenced frame config instance=].
 
   1. If |instance| is null, then return.
 
   1. If |instance|'s [=fenced frame config instance/mapped url=] is null, then the behavior is
      unspecified.
 
-  1. If this [=document=]'s [=origin=] and |instance|'s
+  1. If the [=relevant settings object=]'s [=origin=] and |instance|'s
      [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
 
   1. If |instance|'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
 
-  1. If |event| is a `DOMString`, run [=report a private aggregation event=] with |event| and
+  1. If |event| is a {{DOMString}}, run [=report a private aggregation event=] with |event| and
      |instance|'s [=fenced frame config instance/fenced frame reporter=].
 
-  1. If |event| is a `FenceEvent`, run [=report an event=] with |event| and |instance|'s
+  1. If |event| is a {{FenceEvent}}, run [=report an event=] with |event| and |instance|'s
      [=fenced frame config instance/fenced frame reporter=].
 </div>
 
 <div algorithm>
   The <dfn method for=Fence>setReportEventDataForAutomaticBeacons(<var ignore>event</var>)</dfn> method steps are:
 
-  1. Let |instance| be this [=traversable navigable=]'s [=fenced frame config instance=].
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node
+     navigable=]'s [=navigable/traversable navigable=]'s [=fenced frame config instance=].
 
   1. If |instance| is null, then return.
 
   1. If |instance|'s [=fenced frame config instance/mapped url=] is null, then the behavior is
      unspecified.
 
-  1. If this [=document=]'s [=origin=] and |instance|'s
+  1. If the [=relevant settings object=]'s [=origin=] and |instance|'s
      [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
 
   1. If |instance|'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
@@ -728,14 +730,15 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 <div algorithm>
   The <dfn method for=Fence>getNestedConfigs()</dfn> method steps are:
 
-  1. Let |instance| be this [=traversable navigable=]'s [=fenced frame config instance=].
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node
+     navigable=]'s [=navigable/traversable navigable=]'s [=fenced frame config instance=].
 
   1. If |instance| is null, then return.
 
   1. If |instance|'s [=fenced frame config instance/mapped url=] is null, then the behavior is
      unspecified.
 
-  1. If this [=document=]'s [=origin=] and |instance|'s
+  1. If the [=relevant settings object=]'s [=origin=] and |instance|'s
      [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
 
   1. If |instance|'s [=fenced frame config instance/nested configs=] is null, then return.
--- a/spec.bs
+++ b/spec.bs
@@ -289,8 +289,10 @@ elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension a
 
 <h3 id=configuration-map>Configuration mapping</h3>
 
-Each [=traversable navigable=] has a <dfn for="traversable navigable" export>urn configuration
+Each [=traversable navigable=] has a <dfn for="traversable navigable" export>fenced frame config
 mapping</dfn>, a [=map=] mapping [=urn uuids=] to [=fenced frame config=] [=structs=].
+
+TODO: Set up the mapping so that it can be removed from the spec when urn iframes are gone?
 
 Note: This mapping is consulted during [=navigate|navigation=], and written to by what we
 colloquially refer to as *URN-generating APIs* or *config-generating APIs*, that generate both [=urn
@@ -323,14 +325,6 @@ Issue: Fix the "indistinguishable" IDL bug with the unions above.</span>
 Each {{FencedFrameConfig}} has a <dfn for=fencedframeconfig>url</dfn>, which is a string, initially
 null.
 
-Each {{FencedFrameConfig}} has a <dfn for=fencedframeconfig>urn uuid</dfn>, which is a [=urn uuid=]
-string, initially null.
-
-Note: A {{FencedFrameConfig}}'s [=fencedframeconfig/urn uuid=] is a non-web-exposed unique ID that
-serves as the key to the <{fencedframe}> element's [=fenced navigable container/fenced navigable=]'s
-<span class=XXX>URN config mapping</span>, which maps [=urn uuids=] to [=fenced frame config=]
-[=structs=] (that represent the web-exposed {{FencedFrameConfig}} object).
-
 A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/items=]:
 
 * <dfn for="fenced frame config">mapped url</dfn>, optional<url * boolean> (boolean is visibility to embedder)
@@ -342,12 +336,6 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
 * <dfn for="fenced frame config">fenced frame reporter</dfn>, optional<reporter * boolean * boolean>: TODO make sure reporters handles component ads correctly
 * <dfn for="fenced frame config">exfiltration budget metadata</dfn>, optional<(origin * double * boolean) * boolean * boolean>
 * <dfn for="fenced frame config">nested configs</dfn>, optional<(sequence of configs) * boolean * boolean>
-
-A <dfn export>fenced frame config mapping</dfn> is a map where each entry has the following structure:
-* The key is a <dfn for="fenced frame config mapping">urn uuid</dfn>, a [=urn uuid=] string
-* The value is a fenced frame config
-
-TODO: Set up the mapping so that it can be removed from the spec when urn iframes are gone?
 
 A <dfn export>fenced frame config instance</dfn> is a struct with the following [=struct/items=]:
 * <dfn for="fenced frame config instance">mapped url</dfn>
--- a/spec.bs
+++ b/spec.bs
@@ -321,9 +321,9 @@ At a high level, in order to store a [=fenced frame config=] in the
 [=traversable navigable/fenced frame config mapping=], one must first store a pending config, and
 then turn the pending config into a finalized config. Those procedures are as follows:
 
-<div algorithm="store pending config">
-  In order to store a [=fenced frame config=] into the
-  [=fenced frame config mapping/pending fenced frame config mapping=], run these steps:
+<div algorithm>
+  In order to <dfn export>store a [=fenced frame config=] into the
+  [=fenced frame config mapping/pending fenced frame config mapping=]</dfn>, run these steps:
 
   1. If the size of the [=fenced frame config mapping/pending fenced frame config mapping=] + the
      size of the [=fenced frame config mapping/finalized fenced frame config mapping=] meets or
@@ -334,9 +334,9 @@ then turn the pending config into a finalized config. Those procedures are as fo
      [=fenced frame config=].
 </div>
 
-<div algorithm="store finalized config">
-  In order to store a pending [=fenced frame config=] at a [=urn uuid=] into the
-  [=fenced frame config mapping/finalized fenced frame config mapping=], run these steps:
+<div algorithm>
+  In order to <dfn export>store a pending [=fenced frame config=] at a [=urn uuid=] into the
+  [=fenced frame config mapping/finalized fenced frame config mapping=]</dfn>, run these steps:
 
   1. If the [=urn uuid=] is not present in the
      [=fenced frame config mapping/pending fenced frame config mapping=], return failure.
@@ -347,9 +347,9 @@ then turn the pending config into a finalized config. Those procedures are as fo
      [=urn uuid=] and the value [=fenced frame config=].
 </div>
 
-<div algorithm="look up config">
-  In order to look up a [=fenced frame config=] at a [=urn uuid=] in the
-  [=traversable navigable/fenced frame config mapping=] for navigation, run these steps:
+<div algorithm>
+  In order to <dfn export>look up a [=fenced frame config=] at a [=urn uuid=] in the
+  [=traversable navigable/fenced frame config mapping=] for navigation</dfn>, run these steps:
 
   1. If the [=urn uuid=] is present in the
      [=fenced frame config mapping/pending fenced frame config mapping=], wait until it is no
@@ -563,8 +563,8 @@ A <dfn export>fenced frame config instance</dfn> is a struct with the following 
     :: the [=fenced frame config=]'s [=fenced frame config/embedder shared storage context=]
 </div>
 
-<div algorithm=redaction>
-  In order to redact a [=fenced frame config=], run these steps:
+<div algorithm>
+  In order to <dfn export>redact a [=fenced frame config=]</dfn>, run these steps:
 
   1. TODO
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -287,16 +287,49 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
 attributes</a> section. This section will be updated to include <{fencedframe}> in the list of
 elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension attributes apply to.
 
-<h3 id=configuration-map>Configuration mapping</h3>
+<h3 id=fenced-frame-config-map>Fenced frame config mapping</h3>
 
 Each [=traversable navigable=] has a <dfn for="traversable navigable" export>fenced frame config
-mapping</dfn>, a [=map=] mapping [=urn uuids=] to [=fenced frame config=] [=structs=].
+mapping</dfn>. At a high level, this object maps [=urn uuids=] to [=fenced frame config=]
+[=structs=].
 
 Note: This mapping is consulted during [=navigate|navigation=], and written to by what we
 colloquially refer to as *URN-generating APIs* or *config-generating APIs*, that generate both [=urn
 uuids=] and [=fenced frame configs=] for use in navigating <{fencedframe}> and <{iframe}> elements.
 See for example, the <a href=https://wicg.github.io/turtledove/>FLEDGE</a> and <a
 href=https://wicg.github.io/shared-storage/>Shared Storage</a> specifications.
+
+Each [=traversable navigable/fenced frame config mapping=] has a
+<dfn for="fenced frame config mapping" export>maximum number of configs</dfn>, which is
+implementation-defined. The [=maximum number of configs=] may be a nonnegative number or infinity.
+
+Note: It is important to specify the behavior of [=maximum number of configs=] because its
+semantics can interact with config-generating APIs in a privacy sensitive way.
+
+The [=traversable navigable/fenced frame config mapping=] actually consists of two submappings:
+<dl export dfn-for="fenced frame config mapping">
+  : <dfn>pending fenced frame config mapping</dfn>
+  :: a [=map=] from [=urn uuids=] to tuples of [=fenced frame configs=] and TODO: callbacks
+
+  : <dfn>finalized fenced frame config mapping</dfn>
+  :: a [=map=] from [=urn uuids=] to [=fenced frame configs=]
+</dl>
+
+At a high level, in order to store a config in the
+[=traversable navigable/fenced frame config mapping=], one must first store a pending config, and
+then turn the pending config into a finalized config. Those procedures are as follows:
+
+<div algorithm="store pending config">
+  In order to store a [=fenced frame config=] into the [=pending fenced frame config mapping=], run these steps:
+
+  1. TODO
+</div>
+
+<div algorithm="store finalized config">
+  In order to store a [=fenced frame config=] into the [=finalized fenced frame config mapping=], run these steps:
+
+  1. TODO
+</div>
 
 <h3 id=fenced-frame-config-interface>The {{FencedFrameConfig}} interface</h3>
 
@@ -454,7 +487,7 @@ A <dfn export>fenced frame config instance</dfn> is a struct with the following 
   :: null, or an [=fencedframetype/embedder shared storage context=]
 </dl>
 
-<div algorithm=configinstantiation>
+<div algorithm="config instantiation">
   A [=fenced frame config instance=] is initialized from a [=fenced frame config=] as follows:
 
     : [=fenced frame config instance/mapped url=]
--- a/spec.bs
+++ b/spec.bs
@@ -292,8 +292,6 @@ elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension a
 Each [=traversable navigable=] has a <dfn for="traversable navigable" export>fenced frame config
 mapping</dfn>, a [=map=] mapping [=urn uuids=] to [=fenced frame config=] [=structs=].
 
-TODO: Set up the mapping so that it can be removed from the spec when urn iframes are gone?
-
 Note: This mapping is consulted during [=navigate|navigation=], and written to by what we
 colloquially refer to as *URN-generating APIs* or *config-generating APIs*, that generate both [=urn
 uuids=] and [=fenced frame configs=] for use in navigating <{fencedframe}> and <{iframe}> elements.
--- a/spec.bs
+++ b/spec.bs
@@ -383,18 +383,18 @@ then turn the pending config into a finalized config. Those procedures are as fo
   1. Let |nestedMapping| be the [=fenced frame config mapping/nested config mapping=] of the
      associated [=fenced frame config mapping=].
 
-  1. If |nestedMapping|[|urn|] does [=map/exist=], return its value.
-
   1. Let |pendingMapping| be the [=fenced frame config mapping/pending config mapping=] of the
      associated [=fenced frame config mapping=].
-
-  1. If |pendingMapping|[|urn|] does [=map/exist=], wait until it does not [=map/exist=].
-     <span class="XXX">TODO: Formalize waiting?</span>
 
   1. Let |finalizedMapping| be the [=fenced frame config mapping/finalized config mapping=] of
      the associated [=fenced frame config mapping=].
 
-  1. If |finalizedMapping|[|urn|] does [=map/exist=], return its value.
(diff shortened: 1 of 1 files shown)


domfarolino:
Per our earlier discussion, all that's left here is making all of the optional stuff into the correct "Null or [...]" language, and specifying the visibility type etc right?

gtanzer:
> Per our earlier discussion, all that's left here is making all of the optional stuff into the correct "Null or [...]" language, and specifying the visibility type etc right?

Yes, I believe so.

spec.bs
```diff
@@ -278,9 +278,11 @@ elements that the <code>[=width=]</code> and <code>[=height=]</code> dimension a
 
 <h3 id=configuration-map>Configuration mapping</h3>
 
-Each [=traversable navigable=] has a <dfn for="traversable navigable" export>urn configuration
+Each [=traversable navigable=] has a <dfn for="traversable navigable" export>fenced frame config
 mapping</dfn>, a [=map=] mapping [=urn uuids=] to [=fenced frame config=] [=structs=].
 
+TODO: Set up the mapping so that it can be removed from the spec when urn iframes are gone?
```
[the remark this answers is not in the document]
gtanzer: https://github.com/WICG/fenced-frame/issues/70

spec.bs
```diff
@@ -310,21 +343,189 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+
+    : [=fenced frame config instance/partition nonce=]
+    :: a random [=partition nonce=] (TODO)
+
+    : [=fenced frame config instance/embedder shared storage context=]
+    :: the [=fenced frame config=]'s [=fenced frame config/embedder shared storage context=]
+</div>
+
+<div algorithm=redaction>
+  In order to redact a [=fenced frame config=], run these steps:
```
domfarolino, line 525: I'd make this:

```html
<div algorithm>
  In other to <dfn export>redact a fenced frame config</dfn>, run these steps:
</div>
```

So that it can be referenced as a dfn from other specs.

spec.bs
```diff
@@ -310,21 +343,189 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+  : <dfn>fenced frame reporter</dfn> TODO: including automatic beacon info
+  :: null, or a struct containing a [=fencedframetype/fenced frame reporter=], a
+     [=visibility/visibility to embedder=], and a [=visibility/visibility to content=]
+
+  : <dfn>exfiltration budget metadata</dfn>
+  :: null, or a struct containing an [=fencedframetype/exfiltration budget metadata=], a
+     [=visibility/visibility to embedder=], and a [=visibility/visibility to content=]
+
+  : <dfn>nested configs</dfn>
+  :: null, or a struct containing a sequence of pairs of [=urn uuids] and [=fenced frame configs=]
```
domfarolino, line 469: missing `=` after `urn uuids`

spec.bs
```diff
@@ -276,17 +276,50 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+then turn the pending config into a finalized config. Those procedures are as follows:
+
+<div algorithm="store pending config">
+  In order to store a [=fenced frame config=] into the [=pending fenced frame config mapping=], run these steps:
+
+  1. TODO
+</div>
+
+<div algorithm="store finalized config">
+  In order to store a [=fenced frame config=] into the [=finalized fenced frame config mapping=], run these steps:
```
domfarolino, line 318: This and the above algorithm should follow the advice in my comment for "redacting" a config.

spec.bs
```diff
@@ -310,21 +343,189 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+of fields that can be loaded into <{fencedframe}> elements and that specifies the resulting
+environments. Fenced frame configs can only be constructed by web platform APIs, not initialized
+or modified arbitrarily. Their fields also contain "visibilities" for different entities, which
+dictate whether the field should be "redacted" when inspected by particular execution contexts.
+Each time a fenced frame config is loaded into a <{fencedframe}> element, it is instantiated as a
+new "fenced frame config instance", which governs that particular context.
+
+We now establish some preliminary types:
+
+A <dfn export for=fencedframeconfig>visibility</dfn> is either
```
domfarolino, line 365: Will anyone ever directly reference this dfn instead of the two embedder- and content-specific ones? If not, maybe we can just define both types as having the generic values defined below.

gtanzer: There may be places in the redaction spec where we want to phrase things generally in terms of visibilities. But outside of the FF spec, I don't think so.

spec.bs
gtanzer: I changed it to have `<dfn export>redact a [=fenced frame config=]</dfn>`, lmk if the link inside the definition is a problem.

spec.bs
```diff
@@ -276,17 +276,109 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
 
 Note: This mapping is consulted during [=navigate|navigation=], and written to by what we
 colloquially refer to as *URN-generating APIs* or *config-generating APIs*, that generate both [=urn
 uuids=] and [=fenced frame configs=] for use in navigating <{fencedframe}> and <{iframe}> elements.
 See for example, the <a href=https://wicg.github.io/turtledove/>FLEDGE</a> and <a
 href=https://wicg.github.io/shared-storage/>Shared Storage</a> specifications.
 
+A <dfn>fenced frame config mapping</dfn> has three submappings:
+
+<dl export dfn-for="fenced frame config mapping">
```
domfarolino, line 292: Do all of these need to be exported (i.e., directly usable by other specifications), or can we reference them indirectly through the algorithms?

spec.bs
```diff
@@ -276,17 +276,109 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+  :: a [=map=] whose [=map/keys=] are [=urn uuids=] and whose [=map/values=] are [=fenced frame configs=]
+
+  : <dfn>finalized config mapping</dfn>
+  :: a [=map=] whose [=map/keys=] are [=urn uuids=] and whose [=map/values=] are [=fenced frame configs=]
+
+  : <dfn>nested config mapping</dfn>
+  :: a [=map=] whose [=map/keys=] are [=urn uuids=] and whose [=map/values=] are [=fenced frame configs=]
+</dl>
+
+Each [=fenced frame config mapping=] has a <dfn for="fenced frame config mapping" export>maximum
```
domfarolino, line 303: same wrt `export`

spec.bs
```diff
@@ -276,17 +276,109 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+  1. If |nestedConfigs| is null, return.
+
+  1. [=map/iterate|For each=] |urn| → |config| of |nestedConfigs|:
+
+     1. [=map/Set=] |nestedMapping|[|urn|] to |config|.
+</div>
+
+
+<div algorithm>
+  In order to <dfn export>look up a [=fenced frame config=] at a [=urn uuid=] in the
```
domfarolino, line 366: Can you follow the algorithm trend I did in https://github.com/WICG/fenced-frame/pull/67/commits/321f2a0a10304bba50790c8cc75c064a89a2fa0f with this one, especially regarding usage of infra primitives?

spec.bs
```diff
@@ -309,26 +401,242 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+
 Each {{FencedFrameConfig}} has a <dfn for=fencedframeconfig>url</dfn>, which is a string, initially
-null.
+null. TODO: flesh out {{FencedFrameConfig}}
 
-Each {{FencedFrameConfig}} has a <dfn for=fencedframeconfig>urn uuid</dfn>, which is a [=urn uuid=]
-string, initially null.
+A key feature of the <{fencedframe}> element is that web platform APIs can configure the behavior
+of the frame in a way that limits the ability of other execution contexts to modify or inspect this
+configuration, for security or privacy reasons. For example, the FLEDGE API (TODO:ref) performs
```
domfarolino, line 416: We should be able to link directly via `[FLEDGE](https://wicg.github.io/turtledove)`

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
+
+  1. If *instance* is null, then return.
+
+  1. If *instance*'s [=fenced frame config instance/mapped url=] is null, then the behavior is
+     unspecified.
+
+  1. If this [=document=]'s [=origin=] and *instance*'s
+     [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
+
+  1. If *instance*'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
```
domfarolino, line 679: Please use `|instance|` syntax inside algorithms for variables here and elsewhere in this PR (as we do in the rest of the spec).

spec.bs
```diff
@@ -309,26 +401,242 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+
+    : [=fenced frame config instance/partition nonce=]
+    :: a random [=partition nonce=] <span class=XXX>TODO: Specify what this means</span>
+
+    : [=fenced frame config instance/embedder shared storage context=]
+    :: the [=fenced frame config=]'s [=fenced frame config/embedder shared storage context=]
+</div>
+
+<div algorithm>
+  In order to <dfn export>redact a [=fenced frame config=]</dfn>, run these steps:
```
domfarolino, line 632: ditto, follow the trend in my previous commit wrt to signature and naming etc.

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
+     unspecified.
+
+  1. If this [=document=]'s [=origin=] and *instance*'s
+     [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
+
+  1. If *instance*'s [=fenced frame config instance/nested configs=] is null, then return.
+
+  1. Let *results* be a sequence of {{FencedFrameConfig}}s.
+
+  1. For each [=urn uuid=] *urn* and [=fenced frame config=] *config* in *instance*'s
```
domfarolino, line 724: Please continue the For each iteration map syntax that I used in my previous commit

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
+
+  1. If *instance*'s [=fenced frame config instance/mapped url=] is null, then the behavior is
+     unspecified.
+
+  1. If this [=document=]'s [=origin=] and *instance*'s
+     [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
+
+  1. If *instance*'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
+
+  1. If *event* is a `DOMString`, run [=report a private aggregation event=] with *event* and
```
domfarolino, line 681: Should be able to just use `[=DOMString=]` maybe? Or {{DOMString}}, not sure.

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
+     *instance*'s [=fenced frame config instance/fenced frame reporter=].
+
+  1. If *event* is a `FenceEvent`, run [=report an event=] with *event* and *instance*'s
```
domfarolino, line 684: Same, you should be able to use {{FenceEvent}}, per https://speced.github.io/bikeshed/#:~:text=idl%2C%20covering%20the%20dfn%20autolinks%20for%20WebIDL%20types%20(the%20%7B%7Bfoo%7D%7D%20shorthand)

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
+  1. If this [=document=]'s [=origin=] and *instance*'s
+     [=fenced frame config instance/mapped url=]'s origin are not [=same origin=], then return.
+
+  1. If *instance*'s [=fenced frame config instance/nested configs=] is null, then return.
+
+  1. Let *results* be a sequence of {{FencedFrameConfig}}s.
+
+  1. For each [=urn uuid=] *urn* and [=fenced frame config=] *config* in *instance*'s
+     [=fenced frame config instance/nested configs=]:
+     1. Redact *config* for the embedder (TODO:ref), into *redacted config*.
```
domfarolino, line 726: Can you replace TODO:ref with the appropriate reference (or if this was an easter-egg for _me_ to fill in let me know and I can help).

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
 
-  1. <span class=XXX>Fill this out!</span>
+  1. Let *instance* be this [=traversable navigable=]'s [=fenced frame config instance=].
+
+  1. If *instance* is null, then return.
+
+  1. If *instance*'s [=fenced frame config instance/mapped url=] is null, then the behavior is
+     unspecified.
+
+  1. If this [=document=]'s [=origin=] and *instance*'s
```
domfarolino, line 676: Here and elsewhere let's follow the pattern like https://wicg.github.io/turtledove/#:~:text=durationSeconds)%20method%20steps%20are%3A-,If%20this%27s%20relevant%20global%20object%27s%20associated%20Document,-is%20not%20allowed%20to%20use when it comes to accessing the current document.

spec.bs
```diff
@@ -358,19 +666,67 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
 <div algorithm>
   The <dfn method for=Fence>reportEvent(<var ignore>event</var>)</dfn> method steps are:
 
-  1. <span class=XXX>Fill this out!</span>
+  1. Let *instance* be this [=traversable navigable=]'s [=fenced frame config instance=].
```
domfarolino, line 669: We should be able to more formally access the traversable navigable via `Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=Document/node navigable=]'s [=navigable/traversable navigable=]` I think (here and any other places in this PR).

spec.bs
```diff
@@ -309,26 +401,242 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+  : <dfn>fenced frame reporter</dfn>
+  :: null, or a struct containing a [=fencedframetype/fenced frame reporter=], a
+     [=visibility/visibility to embedder=], and a [=visibility/visibility to content=]
+
+  : <dfn>exfiltration budget metadata</dfn>
+  :: null, or a struct containing an [=fencedframetype/exfiltration budget metadata=], a
+     [=visibility/visibility to embedder=], and a [=visibility/visibility to content=]
+
+  : <dfn>nested configs</dfn>
+  :: null, or a struct containing a sequence of [=fenced frame configs=] and a
```
domfarolino, line 522: Could we formalize this a bit? I read here that this is a struct-or-null, but in https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/67.html#store-nested-configs we try and iterate over `nested configs` which doesn't really make sense since this is a struct. It sounds like we want this to be:

`nested configs: {
    configs: map<urn, config>;
    visibility: visibility;
}`

in pseudo code? If that's right, let's give this a _member_ that is a map that we can explicitly reference when performing the iteration in the aforementioned link (and anywhere else, if applicable).

domfarolino (commented):
Getting close!

spec.bs
gtanzer: I think we can just use the algorithms. Removed the export.

spec.bs
gtanzer: Done.

spec.bs
gtanzer: Changing the algorithm names to match the models you gave.

spec.bs
gtanzer: I think we actually don't need visibility to content, like I suspected earlier, but let me finish fleshing out all the fields/redaction and see if that's still the case.

spec.bs
gtanzer: It gave me an error with node navigable in Document, so I changed it to:

`[=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node navigable=]'s [=navigable/traversable navigable=]`

spec.bs
gtanzer: Changed to the [=relevant settings object=] to match FLEDGE spec, but lmk if I'm misunderstanding what that means.

spec.bs
```diff
@@ -287,17 +287,118 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+
+<div algorithm>
+  To <dfn for="fenced frame config mapping" export>store a pending config</dfn> into a particular
+  [=fenced frame config mapping=] given a [=fenced frame config=] |config|, run these steps:
+
+  1. Let |pendingMapping| be the [=fenced frame config mapping/pending config mapping=] of the
+     associated [=fenced frame config mapping=].
+
+  1. If the [=map/size=] of the [=fenced frame config mapping/pending config mapping=] + the
+     size of the [=fenced frame config mapping/finalized config mapping=] meets or
```
domfarolino, line 337: nit: also refer to `[=map/size=]` here

domfarolino, line 337: s/meets or exceeds/>=

spec.bs
```diff
@@ -287,17 +287,118 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+  |nestedConfigs|, run these steps:
+
+  1. Let |nestedMapping| be the [=fenced frame config mapping/nested config mapping=] of the
+     associated [=fenced frame config mapping=].
+
+  1. If |nestedConfigs| is null, return.
+
+  1. [=map/iterate|For each=] |urn| → |config| of |nestedConfigs|:
+
+  1. [=map/Set=] |nestedMapping|[|urn|] to |config|.
```
domfarolino, line 376: Indent to make this appear in the loop

spec.bs
```diff
@@ -287,17 +287,118 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+  1. Let |pendingMapping| be the [=fenced frame config mapping/pending config mapping=] of the
+     associated [=fenced frame config mapping=].
+
+  1. Let |finalizedMapping| be the [=fenced frame config mapping/finalized config mapping=] of
+     the associated [=fenced frame config mapping=].
+
+  1. If |nestedMapping|[|urn|] [=map/exists=], return its value.
+
+  1. If |pendingMapping|[|urn|] [=map/exists=], wait until it does not [=map/exist=].
+     <span class="XXX">TODO: Formalize waiting?</span>
```
domfarolino, line 395: This is the piece that's intended to be called during navigation while in parallel, waiting for a pending config right?

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
 
 Issue: Fix the "indistinguishable" IDL bug with the unions above.</span>
 
+<div algorithm>
+  In order to <dfn export>lift a [=urn uuid=] and [=fenced frame config=] into a
```
domfarolino, line 425: Remind me what this will be used for? I'm not sure I understand what the intention is here (especially with the word "lift"). Is this what will be used by APIs that take a config from the "browser process" and generate the appropriate WebIDL object to expose to script?

domfarolino, line 425: Follow the conventions wrt naming and variables in the algorithm signature.

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+
+  : <dfn>partition nonce</dfn>
+  :: a [=partition nonce=]
+
+  : <dfn>embedder shared storage context</dfn>
+  :: null, or an [=fencedframetype/embedder shared storage context=]
+</dl>
+
+<div algorithm="config instantiation">
+  A [=fenced frame config instance=] is initialized from a [=fenced frame config=] as follows:
```
domfarolino, line 597: Let's have this algorithm explicitly return "a new fenced frame config instance, with:" _and then_ list the members below.

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+</div>
+
+<div algorithm>
+  In order to <dfn export> report a private aggregation event</dfn>, run these steps:
+
+  1. <span class=XXX>TODO: Fill this in</span>
+</div>
+
+<div algorithm>
+  In order to <dfn export> set automatic beacon data</dfn>, run these steps:
```
domfarolino, line 487: nit: These three algorithms have an extra space after `<dfn export>`

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+without the API caller knowing *which ad* won the auction or being able to manipulate the
+environment in which the ad loads.
+
+We achieve this using the concept of a "fenced frame config". A fenced frame config is a collection
+of fields that can be loaded into <{fencedframe}> elements and that specifies the resulting
+environments. Fenced frame configs can only be constructed by web platform APIs, not initialized
+or modified arbitrarily. Their fields also contain "visibilities" for different entities, which
+dictate whether the field should be "redacted" when inspected by particular execution contexts.
+Each time a fenced frame config is loaded into a <{fencedframe}> element, it is instantiated as a
+new "fenced frame config instance", which governs that particular context.
```
domfarolino, line 448: These two paragraphs are very informal and don't do a lot of linking or describe normative requirements. I'd recommend one of two things:
 1. Make this into a sort of "Introduction" subsection with the `*This section is non-normative*` disclaimer, and then make the following bits with the "preliminary types" into its own section called "Supporting concepts"
 2. Formalize these two paragraphs a bit and add a bunch of linking, and turn the "For example" stuff into its own example or note block.

I slightly prefer (1), but will leave it to you.

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+
+  : <dfn>nested configs</dfn>
+  :: null, or a struct containing a sequence of [=fenced frame configs=] and a
+     [=visibility/visibility to embedder=]
+
+  : <dfn>embedder shared storage context</dfn>
+  :: null, or an [=fencedframetype/embedder shared storage context=]
+</dl>
+
+A <dfn export>fenced frame config instance</dfn> is a struct with the following [=struct/items=]:
```
domfarolino, line 550: I think it would be good to have some sort of note or description about _who_ is responsible for converting a config into a config instance.

spec.bs
```diff
@@ -320,26 +421,243 @@ maps to an internal [=fenced frame config=] [=struct=].
...
+of fields that can be loaded into <{fencedframe}> elements and that specifies the resulting
+environments. Fenced frame configs can only be constructed by web platform APIs, not initialized
+or modified arbitrarily. Their fields also contain "visibilities" for different entities, which
+dictate whether the field should be "redacted" when inspected by particular execution contexts.
+Each time a fenced frame config is loaded into a <{fencedframe}> element, it is instantiated as a
+new "fenced frame config instance", which governs that particular context.
+
+We now establish some preliminary types:
+
+A <dfn export for=fencedframeconfig>visibility</dfn> is either "<dfn export for=visibility>
```
domfarolino, line 452: Not sure where this should go off the top of my head, but let's also mention somewhere in this section below that the creator of a config is responsible for passing in _everything_ that can go into it, and that there are no optional members, though they can be explicitly set to null. It could be a note, I'm just trying to make the flow clear somewhere. Maybe it would be good in the "Introduction"-style section that I mentioned above that we could add.

spec.bs
gtanzer: I was thinking that we might need an equivalent of RedactedFencedFrameConfig, but I don't think we do anymore. The only complication is that we might need to put the urn uuid back into the config struct.

spec.bs
gtanzer: Changed this into a different kind of TODO

spec.bs
gtanzer: Made it a non-normative intro.

spec.bs
gtanzer: Added this into the non-normative intro. Though if we are requiring [=fenced frame config=] to be constructed with explicit values for each field, maybe that should be somewhere normative.

spec.bs
```diff
@@ -369,19 +739,70 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
...
   The <dfn method for=Fence>reportEvent(<var ignore>event</var>)</dfn> method steps are:
 
-  1. <span class=XXX>Fill this out!</span>
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node
+     navigable=]'s [=navigable/traversable navigable=]'s [=fenced frame config instance=].
+
+  1. If |instance| is null, then return.
+
+  1. If |instance|'s [=fenced frame config instance/mapped url=] is null, then the behavior is
+     unspecified.
```
domfarolino, line 761: This is a little bit strange to me. What does our implementation do? Surely we don't just defer to C++ undefined behavior right? Can we bail out here? (same with the instances below).

spec.bs
```diff
@@ -287,18 +287,383 @@ href="https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimensi
...
+An <dfn export for=fencedframetype>exfiltration budget metadata</dfn> is a struct containing an
+<dfn export for="exfiltration budget metadata">origin</dfn>, which is an [=origin=]; and an <dfn
+export for="exfiltration budget metadata">amount to debit</dfn>, which is a non-negative valid
+floating point number.
+
+An <dfn export for=fencedframetype>exfiltration budget metadata reference</dfn> is a struct
+containing an <dfn export for="exfiltration budget metadata reference">origin</dfn>, which is an
+[=origin=]; and an <dfn export for="exfiltration budget metadata reference">amount to debit
+reference</dfn>, which is a mutable reference to a non-negative valid floating point number.
+<span class=XXX>TODO: are mutable references a thing in spec?</span>
```
domfarolino, line 483: Can this just be the number itself? Specs have the concept of a "pointer to foo" basically, but if we could make this just a plain ole number (instead of a pointer to a number elsewhere) that would be easiest. Is there a strong reason not to do this?

spec.bs
```diff
@@ -369,19 +749,74 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
 <div algorithm>
   The <dfn method for=Fence>reportEvent(<var ignore>event</var>)</dfn> method steps are:
 
-  1. <span class=XXX>Fill this out!</span>
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s [=node
+     navigable=]'s [=navigable/traversable navigable=]'s [=fenced frame config instance=].
+
+     <span class=XXX>This and the below references should point to an actual member on [=traversable
+     navigable=], not just the type.</span>
```
domfarolino, line 756: I put this as a TODO here that we should follow-up on quickly, but feel free to address this before we land if you want. Basically I think inside [this section](https://wicg.github.io/fenced-frame/#traversable-navigables) we should introduce a new member called "fenced frame config instance" or whatever is similar to what we have on `DocumentLoader` where this can live. That way here and elsewhere we have a concrete thing to initialize, assign, and address wherever we need.

domfarolino (approved):
Looking great, just a few small comments and I think we're good to go

spec.bs
```diff
@@ -681,6 +688,9 @@ maps to an internal [=fenced frame config=] [=struct=].
 
 Issue: Fix the "indistinguishable" IDL bug with the unions above.</span>
 
+Each {{FencedFrameConfig}} has a <dfn for=fencedframeconfig>url</dfn>, which is a [=URL=] or null,
```
gtanzer, line 691: The reason why I deleted this is: what does it mean for `url` to be "initially null"? All {{FencedFrameConfig}} constructors should specify the value of `url` (which might be null, but it's not like that's a default).

spec.bs
gtanzer: Agree. Where should it live? If we didn't have to deal with urn iframes I think it could just be attached to the fenced content navigable or whatever, but we probably have to attach it to the root.

gtanzer: That reminds me: this part about fenced frame config mappings is probably wrong for urn iframes (though it is wrong in the same way that our implementation is "wrong"):

```
Each [=traversable navigable=] has a <dfn for="traversable navigable" export>fenced frame config
mapping</dfn>, which is a new [=fenced frame config mapping=].
```

spec.bs
gtanzer: We have a CHECK because this case is currently unreachable. We could make [=mapped url=] not nullable for now, but then we might have to change it later.

spec.bs
gtanzer: It has to be a reference because when you perform a fence-escaping navigation, it subtracts this number from the budget and then sets it to 0 (through the reference, so that it applies to all instances created from the same config).

spec.bs
gtanzer: Is this covered by the non-normative blurb:

```
Each time a <{fencedframe}> navigates to a [=fenced frame config=], it is instantiated as a new
[=fenced frame config instance=], which governs that particular context inside the [=fenced
navigable container/fenced navigable=].
```

or do you mean some additional text right next to this declaration? I imagine the text would be basically identical to that quoted bit.

spec.bs
gtanzer: Oh, you added this back to fix the compile warning...

gtanzer:
@domfarolino Looks fine to merge to me. Still a little stuff to do in followups.

spec.bs
domfarolino: That's right. Maybe I'll just make it a URL that's not initially null, and that has to be provided upon construction.

spec.bs
domfarolino: I'll make this a spec assert, but if that's really a requirement for now then we should just make the member non-nullable and change that in the future if we ever support that.

spec.bs
domfarolino: No I think this covers it, I must've forgotten about that para during my review.

spec.bs
domfarolino: Where does "the budget" that we subtract _from_ ultimately live? Right now in our spec it seems like a closed loop where the config member is referenced by the instance, and the instance member is referenced by the instantiation algorithm.

spec.bs
domfarolino: So the fenced navigable is indeed a traversable navigable, per https://wicg.github.io/fenced-frame/#fenced-navigable-container-fenced-navigable. That's the root, essentially `content::Page` / `content::FencedFrame` so the member should probably live there, and iframes can climb the tree to reference it. That's basically what the implementation does, right?

domfarolino: Ah yeah... hmm OK I think for now we should make it a normal member of "navigable" (basically FTN) with some strong Note saying that once URN iframes are gone this will exclusively live on "traversable navigable".

PR #68, opened 2023-04-10
Add focusing changes + monkeypatches

blu25:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/68.html" title="Last updated on Apr 17, 2023, 12:08 PM UTC (2e331e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/68/4b6b5ec...2e331e6.html" title="Last updated on Apr 17, 2023, 12:08 PM UTC (2e331e6)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -78,9 +78,17 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: interaction.html
       text: activation notification; url: activation-notification
       text: consume user activation; url: consume-user-activation
+      text: activation; url: activation
+      text: click focusable; url: click-focusable
+      text: focusable area; url: focusable-area
+      text: sequential focus navigation; url: sequential-focus-navigation
+    urlPefix: infrastructure.html
+      text: immediately; url: immediately
     urlPrefix: nav-history-apis.html
       for: Window
         text: navigable; url: window-navigable
+    urlPrefix: interactive-elements.html
+      text: accesskey attribute command; url: using-the-accesskey-attribute-to-define-a-command-on-other-elements
 spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
   type: dfn
     text: structured header; url: #section-1
@@ -651,6 +659,49 @@ traversables</a> section.
   1. Return |navigables|.
 </div>
 
+<h3 id=focusing-steps-algorithm>Modifications to the focusing steps algorithm</h3>
+
+  Modify the [=focusing steps=] algorithm to take a new optional [=boolean=]
+  argument <dfn lt="focus-unfenced">unfenced</dfn> that defaults to false.
+
+  Add a new step after step 7 of the algorithm (that defines <var>old chain</var> and <var>new chain</var>) that reads:
+
+  8. If [=focus-unfenced|unfenced=] is set to false, <var>new chain</var>
+     contains a Document with a [=fenced navigable container/fenced navigable=]
+     <var>nav</var>, and <var>old chain</var> does not also contain
+     an entry with <var>nav</var>, then return.
+
+  Modify the user agent note after the algorithm steps to read:
+
+  <p>User agents must [=immediately=] run the [=focusing steps=] for a
+  [=focusable area=] or [=navigable=] <var>candidate</var> whenever the
+  user attempts to move the focus to <var>candidate</var>.</p>
+
+  Modify the action of the [=accesskey attribute command=] algorithm to be:
+
+  <ol>
+   <li>Run the [=focusing steps=] for the element with [=focus-unfenced|unfenced=] set to true.</li>
+
+   <li><span>Fire a <code data-x="event-click">click</code> event</span> at the element.</li>
+  </ol>
+
+  Modify the behavior when a user [=activation|activates=] a [=click focusable=] [=focusable area=] to be:
+
+  <p>When a user [=activation|activates=] a [=click focusable=] [=focusable area=],
+  the user agent must run the [=focusing steps=] on the [=focusable area=]
+  with <var ignore=''>focus trigger</var> set to "<code data-x="">click</code>" and
+  [=focus-unfenced|unfenced=] set to true.</p>
+
+  Modify step 6 of the [=sequential focus navigation=] algorithm to be:
+
+  6. If <var>candidate</var> is not null, then run the [=focusing steps=] for
+    <var>candidate</var> with [=focus-unfenced|unfenced=] set to true and return.
+
+  Modify step 10 of the [=hide popover algorithm=] to read:
+
+  10. 
+
+<h3 id=focus-chain-algorithm>Modifications to the focus chain algorithm</h3>
 
 <h3 id=supports-loading-mode>The \`<a http-header><code>Supports-Loading-Mode</code></a>\` HTTP response header</h3>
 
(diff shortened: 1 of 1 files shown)


domfarolino:
One question: Do we need to modify https://html.spec.whatwg.org/multipage/interaction.html#sequential-navigation-search-algorithm? I guess we'll catch this with the audit anyways, in which case we don't have to modify it in this PR, right?

spec.bs
```diff
@@ -651,6 +677,169 @@ traversables</a> section.
...
+  Modify step 10 of the [=hide popover algorithm=] to read:
+
+  <ol>
+    <li value="10">
+      <p>If <var>previouslyFocusedElement</var> is not null, then:</p>
+
+      <ol>
+      <li><p>Set <var ignore=''>element</var>'s [=previously focused element=] to null.</p></li>
+
+      <li><p>If <var ignore=''>focusPreviousElement</var> is true, then run the [=focusing steps=] for
```
domfarolino, line 772: This step is identical to the one in the spec, and so is the above. Does this patch actually do anything?

spec.bs
```diff
@@ -651,6 +677,169 @@ traversables</a> section.
...
+
+  Add a new step after step 8 (step 7 in the existing version) of the algorithm (that defines |old
+  chain| and |new chain|) that reads:
+
+  9. If [=focus-unfenced|unfenced=] is false, |new chain| [=list/contains=] an element |element|
+     whose [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
+     container/fenced navigable=], and |old chain| does not also [=list/contain=] |element|, then
+     return.
+
+     Note: This is how we bail-out early just before calling the [=focus update steps=], in the case
```
domfarolino, line 726: I added this `Note:`. Does it capture the intent correctly? If so (that is, if I understand what our aim is here), then I think the above step is not adequate. If you move focus from one element inside of a fenced frame to another element inside of a fenced frame, I think the logic in your step will trigger and then we'll mistakenly early-return, right? Does that logic make sense?

domfarolino:
I added some formatting and editorial improvements, I'll try and fix the conflicts since I caused them.

spec.bs
blu25: I forgot to add the unfenced part to this, good catch.

spec.bs
blu25: I think it should be checking for a document instead of an element. That should correctly handle the case of switching focus between 2 elements in a fenced frame.

Both focus chains will contain the fenced frame root's document at that point, and the function will not early return. If we switch focus from one fenced frame to another, now both of the fenced frame root documents will only exist in one of the focus chains, and the function will early return.

blu25:
> One question: Do we need to modify https://html.spec.whatwg.org/multipage/interaction.html#sequential-navigation-search-algorithm? I guess we'll catch this with the audit anyways, in which case we don't have to modify it in this PR, right?

I think we might need to modify it. This is operating under the assumption that "If candidate is a [navigable container](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable-container) with a non-null [content navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#content-navigable)" will not return anything for a fenced frame, so we'd need to add an extra case to do the same thing but with fenced navigables.

I am assuming that "starting point is a [navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable)" will be okay though, since a fenced navigable is still a navigable.

spec.bs
```diff
@@ -759,6 +785,169 @@ in the [[#nested-traversables-intro]].
...
+  <h4 id=focus-chain-changes>Changes to the [=focus chain=] algorithm</h4>
+
+  Modify step 3 of the while loop in the [=focus chain=] algorithm to read:
+  <ol>
+    <li value="3">
+      <p>If <var>currentObject</var> is a [=focusable area=], then set
+      <var>currentObject</var> to <var>currentObject</var>'s [=DOM anchor=]'s 
+      [=Node/node document=].</p>
+
+      <p>Otherwise, if <var>currentObject</var> is a <code>Document</code> whose
```
domfarolino, line 916: Please use {{Document}}.

spec.bs
```diff
@@ -759,6 +787,194 @@ in the [[#nested-traversables-intro]].
...
+  * User agents may focus one of those elements in the process, by running the
+    [=focusing steps=] for that element, and may change the scrolling position of the
+    document, or perform some other action that brings the element to the user's attention.
+    If these steps were invoked by user gesture, [=focusing steps=] can be called with
+    [=focus-unfenced|unfenced=] set to true. For elements that are
+    [=form-associated custom elements=], user agents should use their [=face validation anchor=]
+    instead, for the purposes of these actions.
+</div>
+
+  <h4 id=has-focus-steps-changes>Changes to the [=has focus steps=] algorithm</h4>
```
domfarolino, line 892: While these are somewhat useful, I think they end up polluting the sidebar contents, so may we can remove the `<h4>`s for each individual algorithm and instead just wrap all of these in `<div algorithm>`s to separate/identify them. Sound good?

spec.bs
```diff
@@ -759,6 +787,194 @@ in the [[#nested-traversables-intro]].
...
+  </dl>
+
+  <h4 id=currently-focused-area-changes>Changes to the [=currently focused area of a top-level traversable=] algorithm</h4>
+
+  Modify step 3 of the [=currently focused area of a top-level traversable=] algorithm to read:
+
+  3. While |candidate|'s [=focused area=] is either a [=navigable container=] with a non-null
+    [=navigable container/content navigable=] or a [=fenced navigable container=] with a non-null
+    [=fenced navigable container/fenced navigable=]: set |candidate| to the
+    [=navigable/active document=] of either that [=navigable container=]'s
```
domfarolino, line 939: I think the last couple lines here could be simplified with https://wicg.github.io/fenced-frame/#navigable-unfenced-parent that I added recently. Basically not only does a traversable navigable have the concept of an unfenced parent (member), but now there is also a `[=navigable/unfenced parent=]` _algorithm_ that conditionally takes advantage of the member.

blu25: This algorithm is traversing from the top-level down, not from the bottom up. What I really need is something like an "unfenced child" algorithm. Is that something that would be worth adding to the spec?

domfarolino: Yes I think that would be useful. Feel free to post a separate PR for that if you'd like, but no need to block this one.

spec.bs
```diff
@@ -759,6 +786,185 @@ in the [[#nested-traversables-intro]].
...
+    [=navigable container/content navigable=] or that [=fenced navigable container=]'s
+    [=fenced navigable container/fenced navigable=], whichever is non-null.
+</div>
+
+<div algorithm=sequential-focus-navigation-patch>
```
domfarolino, line 935: All of the sequential focus and navigation changes are unfenced by default it seems. I just want to double check: that's because there are only initiated with user actions, right? I think I'm going to land this right not but please respond here after just for posterity

domfarolino (approved):

spec.bs
blu25: That is correct.

https://docs.google.com/spreadsheets/d/1p7_LtfkB1BAFx_Iu_vQ9J0zLHgd2SaC5ih4ANfhYp0E/edit#gid=533819146&range=152:152

PR #71, opened 2023-04-11
Add scroll bubbling monkeypatches

blu25:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/71.html" title="Last updated on Jul 31, 2023, 4:54 PM UTC (ca2a489)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/71/fb89d8c...ca2a489.html" title="Last updated on Jul 31, 2023, 4:54 PM UTC (ca2a489)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -230,6 +230,10 @@ spec: permissions-policy; urlPrefix: https://w3c.github.io/webappsec-permissions
   type: dfn
     for: permissions policy
       text: inherited policy; url: inherited-policy
+spec: css2; urlPrefix: https://www.w3.org/TR/CSS21/visuren.html
+  type: dfn
+    for: css2
+      text: viewport; url: viewport
 </pre>
 
 <style>
@@ -2446,32 +2450,24 @@ algorithms to achieve the outcomes described in the above explanatory content.
 
 <h3 id=cssom-monkeypatch>CSSOM View</h3>
 
-The [[!CSSOM-VIEW]] specification defines scroll bubbling behavior through the [=scroll a target
-into view=] algorithm. Our specification modifies it to not allow scrolling a target into view if it
-is inside of a fenced frame. This is done because the {{Element/scrollIntoView()}} method is an
-event that can both be invoked from and observed by script, which can open a communication side
-channel between a <{fencedframe}> and its embedder.
+The [[!CSSOM-VIEW]] specification defines the {{Element/scrollIntoView()}} method that calls the
+[=scroll a target into view=] algorithm. This will not only scroll the {{Element}} or
+[=css2/viewport=] to make the target visible, but will also scroll [=tree/ancestor=]s if necessary
+to make the target visible, essentially causing the scroll to "bubble up". This means that
+{{Element/scrollIntoView()}} performed in a [=child navigable=] or [=fenced navigable
+container/fenced navigable=] can be observed by its embedder, allowing for collusion across a fenced
+frame boundary. This section patches {{Element/scrollIntoView()}} to prevent that collusion at the
+expense of some functionality.
 
 <div algorithm=scroll-target-into-view>
-  Modify the [=scroll a target into view=] algorithm to add 2 steps after step 1 that read:
+  Modify the [=scroll a target into view=] algorithm to add a step at the end of the algorithm that reads:
 
-  2. If |target|'s associated {{Document}}'s [=node navigable=]'s [=navigable/traversable
-     navigable=] is a [=fenced navigable container/fenced navigable=], terminate these steps.
-
-  3. If |target|'s associated {{Document}}'s [=Document/ancestor navigables=] [list/contains=] a
-     [=navigable=] whose [=navigable/traversable navigable=] is a [=fenced navigable
+  14. If <var ignore>scrolling box</var>'s associated {{Element}} or [=css2/viewport=]'s associated
+     {{Document}}'s [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
      container/fenced navigable=], terminate these steps.
+
+  Note: This allows scrolling to "bubble up" to a fenced frame boundary, but not cross it.
 </div>
-
-<div algorithm=run-the-scroll-steps>
-  Modify the [=Document/run the scroll steps=] algorithm to add a check before step 1.1 that reads:
-
-  1. If <var ignore>target</var> is a {{Document}} whose [=node navigable=]'s
-     [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
-     [=iteration/continue=].
-</div>
-
-Note: This ensures that scroll events do not affect fenced frames.
 
 <h2 id=security-and-privacy>Security & Privacy Considerations</h2>
 
--- a/spec.bs
+++ b/spec.bs
@@ -2446,11 +2446,11 @@ algorithms to achieve the outcomes described in the above explanatory content.
 
 <h3 id=cssom-monkeypatch>CSSOM View</h3>
 
-The [[!CSSOM-VIEW]] module defines scroll bubbling behavior through the [=scroll a target into
-view=] algorithm. Our specification modifies it to not allow scrolling a target into view if it is
-inside of a fenced frame. This is done because the {{Element/scrollIntoView()}} method is an event
-that can both be triggered from and observed by the web platform, which can open a side channel
-between a fenced frame and its embedder.
+The [[!CSSOM-VIEW]] specification defines scroll bubbling behavior through the [=scroll a target
+into view=] algorithm. Our specification modifies it to not allow scrolling a target into view if it
+is inside of a fenced frame. This is done because the {{Element/scrollIntoView()}} method is an
+event that can both be invoked from and observed by script, which can open a communication side
+channel between a <{fencedframe}> and its embedder.
 
 Modify the [=scroll a target into view=] algorithm to add 2 steps after step 1 that reads:
 
--- a/spec.bs
+++ b/spec.bs
@@ -2457,10 +2457,11 @@ to make the target visible, essentially causing the scroll to "bubble up". This 
 {{Element/scrollIntoView()}} performed in a [=child navigable=] or [=fenced navigable
 container/fenced navigable=] can be observed by its embedder, allowing for collusion across a fenced
 frame boundary. This section patches {{Element/scrollIntoView()}} to prevent that collusion at the
-expense of some functionality.
+expense of some utility.
 
 <div algorithm=scroll-target-into-view>
-  Modify the [=scroll a target into view=] algorithm to add a step at the end of the algorithm that reads:
+  Modify the [=scroll a target into view=] algorithm to add a step at the end of the algorithm that
+  reads:
 
   14. If <var ignore>scrolling box</var>'s associated {{Element}} or [=css2/viewport=]'s associated
      {{Document}}'s [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
--- a/spec.bs
+++ b/spec.bs
@@ -2242,7 +2242,7 @@ achieve the outcomes described in the above explanatory content.
 
 <h3 id=cssom-monkeypatch>CSSOM View</h3>
 
-The <a href=https://www.w3.org/TR/cssom-view-1>CSSOM View Module</a> defines scroll bubbling
+The [[!CSSOM-VIEW]] module defines scroll bubbling
 behavior through the [=scroll a target into view=] algorithm. Our specification modifies it to not
 allow scrolling a target into view if it is inside of a fenced frame. This is done because the
 {{Element/scrollIntoView()}} method is an event that can both be triggered from and observed by the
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -784,3 +788,25 @@ specification adds another value for fenced frames:
...
+
+2. If the <var ignore=''>target</var>-associated {{Document}}'s [=node navigable=]'s
+   [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=], terminate these steps.
+3. If the <var ignore=''>target</var>-associated {{Document}}'s [=Document/ancestor navigables=] contains
+   a [=navigable=] whose [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
+   terminate these steps.
+
+Modify the [=run the scroll steps=] algorithm to add a check after step 1.1 that reads:
+
+2. If <var ignore=''>target</var> is a {{Document}} whose [=node navigable=]'s
```
blu25, line 808: This matches the impl behavior that doesn't propagate scroll events into a fenced frame. Is this the expected behavior? I wrote this just to match what we currently have in impl, but if this isn't what the behavior is supposed to be, we should file a crbug to fix.

domfarolino:
Could you resolve the conflicts on this PR?

spec.bs
```diff
@@ -144,6 +144,10 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     text: structured header; url: #section-1
     for: structured header
       text: token; url: name-tokens
+spec: CSS; urlPrefix: https://www.w3.org/TR/cssom-view-1/
+  type: dfn
+    text: scroll a target into view; url: #scroll-an-element-into-view
```
domfarolino, line 149: Per https://dontcallmedom.github.io/webdex/s.html#scroll%20a%20target%20into%20view%40%40cssom-view%25%25dfn it seems this is exported?

spec.bs
```diff
@@ -144,6 +144,10 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
...
+    text: run the scroll steps; url: document-run-the-scroll-steps
```
domfarolino, line 150: Same here https://dontcallmedom.github.io/webdex/r.html#run%20the%20scroll%20steps%40%40Document%40dfn

spec.bs
```diff
@@ -2239,3 +2239,25 @@ achieve the outcomes described in the above explanatory content.
   /fenced-frame/permission-geolocation.https.html
   /fenced-frame/permission-notification.https.html
 </wpt>
+
+<h3 id=cssom-monkeypatch>CSSOM View</h3>
```
domfarolino, line 2243: In this section please describe why we cannot allow this behavior (aka what privacy concern we're avoiding by implementing this section of the spec)

spec.bs
```diff
@@ -2239,3 +2239,25 @@ achieve the outcomes described in the above explanatory content.
...
+   [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=], terminate these steps.
+3. If the <var ignore=''>target</var>-associated {{Document}}'s [=Document/ancestor navigables=] contains
+   a [=navigable=] whose [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
+   terminate these steps.
+
+Modify the [=Document/run the scroll steps=] algorithm to add a check before step 1.1 that reads:
+
+1. If <var ignore=''>target</var> is a {{Document}} whose [=node navigable=]'s
+   [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
+   continue.
```
domfarolino, line 2261: Below these algorithms, add a link to any web platform tests here as we do elsewhere in this spec.

blu25: We currently have browser tests that are not reflected in any WPT: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/scroll_into_view_browsertest.cc;l=873;drc=4a33cee8668c3ff5f07f9953c7abb921895b9275. Should we attempt to convert those into WPTs?

blu25: For now, since there's nothing to add, I'm going to resolve this conversation. I added something to the tracker to track migrating WPTs, and, once they're migrated, we can update the spec to include those tests.

spec.bs
```diff
@@ -2239,3 +2239,27 @@ achieve the outcomes described in the above explanatory content.
   /fenced-frame/permission-geolocation.https.html
   /fenced-frame/permission-notification.https.html
 </wpt>
+
+<h3 id=cssom-monkeypatch>CSSOM View</h3>
+
+The <a href=https://www.w3.org/TR/cssom-view-1>CSSOM View Module</a> defines scroll bubbling
```
domfarolino, line 2245: Let's use the `[[]]` syntax here as we do elsewhere in this specification

spec.bs
```diff
@@ -2239,3 +2239,27 @@ achieve the outcomes described in the above explanatory content.
...
+
+2. If the <var ignore=''>target</var>-associated {{Document}}'s [=node navigable=]'s
+   [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=], terminate these steps.
+3. If the <var ignore=''>target</var>-associated {{Document}}'s [=Document/ancestor navigables=] contains
+   a [=navigable=] whose [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
+   terminate these steps.
+
+Modify the [=Document/run the scroll steps=] algorithm to add a check before step 1.1 that reads:
+
+1. If <var ignore=''>target</var> is a {{Document}} whose [=node navigable=]'s
```
domfarolino, line 2261: Quotes here are not needed

domfarolino:
The wrapping on this PR seems pretty off. Could you fix it?

domfarolino:
Formatting is still off :( please check your work

blu25:
I got Rewrap working with the spec file, so hopefully the formatting issues shouldn't happen again.

spec.bs
```diff
@@ -2444,6 +2444,35 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+event that can both be invoked from and observed by script, which can open a communication side
+channel between a <{fencedframe}> and its embedder.
+
+<div algorithm=scroll-target-into-view>
+  Modify the [=scroll a target into view=] algorithm to add 2 steps after step 1 that read:
+
+  2. If |target|'s associated {{Document}}'s [=node navigable=]'s [=navigable/traversable
+     navigable=] is a [=fenced navigable container/fenced navigable=], terminate these steps.
+
+  3. If |target|'s associated {{Document}}'s [=Document/ancestor navigables=] [list/contains=] a
```
domfarolino, line 2461: Did you read what "ancestor navigables" does? It crawls "parent" navigables until it reaches null, just like the ["traversable navigable"](https://html.spec.whatwg.org/#nav-traversable) algorithm. So it should not be possible to reach this condition. Do you mean to basically use the "unfenced" ancestor navigables, which this PR might be a reason to introduce to this spec? Or is this just an incoherent step?

spec.bs
```diff
@@ -2444,6 +2444,35 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+
+<div algorithm=run-the-scroll-steps>
+  Modify the [=Document/run the scroll steps=] algorithm to add a check before step 1.1 that reads:
+
+  1. If <var ignore>target</var> is a {{Document}} whose [=node navigable=]'s
+     [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=],
+     [=iteration/continue=].
+</div>
+
+Note: This ensures that scroll events do not affect fenced frames.
```
domfarolino, line 2474: Please elaborate on this as I don't understand what it means. Does this PR make *all* scroll events impossible to trigger in a fenced frame? That seems wrong, since we don't want to impact any that I guess originated from the other side of the fence, right?

domfarolino (commented):
Please look at the fixes I've made in the last commit so that you do not make them again.

spec.bs
blu25: Resolving since this is no longer in the patch.

spec.bs
blu25: Resolving since this is no longer in the patch.

spec.bs
```diff
@@ -2444,6 +2448,27 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+<h3 id=cssom-monkeypatch>CSSOM View</h3>
+
+The [[!CSSOM-VIEW]] specification defines the {{Element/scrollIntoView()}} method that calls the
+[=scroll a target into view=] algorithm. This will not only scroll the {{Element}} or
+[=css2/viewport=] to make the target visible, but will also scroll [=tree/ancestor=]s if necessary
+to make the target visible, essentially causing the scroll to "bubble up". This means that
+{{Element/scrollIntoView()}} performed in a [=child navigable=] or [=fenced navigable
+container/fenced navigable=] can be observed by its embedder, allowing for collusion across a fenced
+frame boundary. This section patches {{Element/scrollIntoView()}} to prevent that collusion at the
+expense of some functionality.
```
domfarolino, line 2460: n/"functionality"/"utility"

spec.bs
```diff
@@ -2444,6 +2448,27 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+
+<div algorithm=scroll-target-into-view>
+  Modify the [=scroll a target into view=] algorithm to add a step at the end of the algorithm that reads:
```
domfarolino, line 2463: I think this could be wrapped to 100 characters tighter?

spec.bs
```diff
@@ -2444,6 +2448,27 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+
+  14. If <var ignore>scrolling box</var>'s associated {{Element}} or [=css2/viewport=]'s associated
```
domfarolino, line 2465: Do we need to do anything about step 13 of this algorithm, which seems to unconditionally crawl up to the "outer" (at least I think that's what it's doing) viewport and call "perform the scroll" again, recursively? *That* seems to be the source of the recursion, no?

blu25: The recursive part is the part at the start of the algorithm that says "means to run these steps for each ancestor element or viewport that establishes a scrolling box scrolling box".

Step 13 is the step that causes the element or viewport to do the actual scroll. Nothing up until that point actually does any scrolling (step 11 is misleading because it mentions aligning, but everything resolves to the `position` variable).

I also don't see any recursive calls in the steps themselves. Step 13 is calling the [perform a scroll](https://drafts.csswg.org/cssom-view-1/#perform-a-scroll) algorithm (which is [the algorithm also called](https://drafts.csswg.org/cssom-view-1/#ref-for-perform-a-scroll) when [a user scrolls with a trackpad or a touch drag](https://drafts.csswg.org/cssom-view-1/#example-vvanimation)), which is separate from the [scroll a target into view](https://drafts.csswg.org/cssom-view-1/#scroll-a-target-into-view) algorithm.

Because of that, I think what we currently have is correct.

spec.bs
```diff
@@ -2444,6 +2448,28 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
 
+<h3 id=cssom-monkeypatch>CSSOM View</h3>
+
+The [[!CSSOM-VIEW]] specification defines the {{Element/scrollIntoView()}} method that calls the
+[=scroll a target into view=] algorithm. This will not only scroll the {{Element}} or
+[=css2/viewport=] to make the target visible, but will also scroll [=tree/ancestor=]s if necessary
+to make the target visible, essentially causing the scroll to "bubble up". This means that
+{{Element/scrollIntoView()}} performed in a [=child navigable=] or [=fenced navigable
+container/fenced navigable=] can be observed by its embedder, allowing for collusion across a fenced
+frame boundary. This section patches {{Element/scrollIntoView()}} to prevent that collusion at the
```
domfarolino, line 2459: We're actually patching the "scroll a target into view" algorithm specifically, so let's mention that _that's_ what we're patching.

spec.bs
```diff
@@ -2444,6 +2448,28 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+{{Element/scrollIntoView()}} performed in a [=child navigable=] or [=fenced navigable container/
+fenced navigable=] can be observed by its embedder, allowing for collusion across a fenced frame
+boundary. This section patches the [=scroll a target into view=] algorithm to prevent that collusion
+at the expense of some utility.
+
+<div algorithm=scroll-target-into-view>
+  Modify the [=scroll a target into view=] algorithm to add a step at the end of the algorithm that
+  reads:
+
+  14. If <var ignore>scrolling box</var>'s associated {{Element}} or [=css2/viewport=]'s associated
```
domfarolino, line 2466: This reads slightly confusing. What you're trying to get at is whether either:
 - scrolling box's associated Element's node navigable's traversable navigable; or
 - scrolling box's associated viewport's associated Document's node navigable's traversable navigable

... is a fenced navigable container. That's the condition you're shooting for right? In that case, I'd strive to be a little more verbose here by being a little redundant as I am above. As it stands, now it ambiguously can read as if you're keying off of whether (a) the associated Element is a fenced navigable container, or (b) the associated viewport's associated Document's node navigable's traversable navigable..... is a fenced navigable container. But that seems wrong, because I don't think the associated element could ever be a fenced navigable container, because the associated element should be outside of the viewport I think (the viewport I think would be everything _inside_ of a fenced navigable container, and we'd only ever have an associated viewport when bubbling out?).

Let me know if that matches your understanding. In any case, I guess we should try and make it is unambiguous as possible, even at the cost of brevity.

spec.bs
```diff
@@ -2444,6 +2448,28 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+     {{Document}}'s [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
+     container/fenced navigable=], terminate these steps.
```
domfarolino, line 2468: "Terminate these steps" is a little unclear just because technically any single instance of this algorithm is indeed terminating after the end of these steps, because the recursion is implicit in the definition and not explicit in the spec steps. So I think we need to say something more aggressive here to combat the implicit recursion allowed for in the definition of this algorithm. Maybe ", let this be the last instance of this algorithm that stops any further recursive instances that would otherwise follow."

spec.bs
blu25: Yes. My intention is to say "If scrolling box’s associated ([Element](https://dom.spec.whatwg.org/#element) or [viewport](https://www.w3.org/TR/CSS21/visuren.html#viewport))'s associated [Document](https://dom.spec.whatwg.org/#document)'s...". I'll write each case out fully to remove any ambiguity.

domfarolino (approved):

spec.bs
```diff
@@ -2444,6 +2448,31 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+  reads:
+
+  14. If <var ignore>scrolling box</var>'s associated {{Element}}'s associated {{Document}}'s [=node
+     navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable container/fenced
+     navigable=], or if <var ignore>scrolling box</var>'s associated [=css2/viewport=]'s associated
+     {{Document}}'s [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
+     container/fenced navigable=], then let this be the last instance of this algorithm that stops
+     any further recursive instances that would otherwise follow.
+
+  Note: This allows scrolling to "bubble up" to a fenced frame boundary, but not cross it.
```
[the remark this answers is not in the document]
blu25: At the moment, no. But we have an item in the fenced frame tracker to migrate the browsertests over to WPTs.

PR #72, opened 2023-04-12
Add CSP and CSPEE patches

blu25:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/72.html" title="Last updated on Apr 24, 2023, 10:17 PM UTC (ddb74bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/72/4b6b5ec...ddb74bc.html" title="Last updated on Apr 24, 2023, 10:17 PM UTC (ddb74bc)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -874,11 +874,28 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h2 id="interaction-with-other-specs">Interactions with other specifications</h2>
+
+Due to the necessarily cross-cutting nature of the <{fencedframe}> element and its interactions with
+core concepts like [=navigable=] and [=browsing context group=], there are a number of
+specifications that rely on terms whose usages must be re-evaluated in light of this specification;
+this section houses the various changes that we propose to other specifications.
+
+<h3 id=prerendering-monkeypatch>Prerendering</h3>
+
+The <a href=https://wicg.github.io/nav-speculation/prerendering.html>Prerendering Revamped</a>
+specification defines [=navigable=]'s [=navigable/loading mode=] and the values it can take on. Our
+specification adds another value for fenced frames:
+
+: "`fencedframe`"
+:: This [=navigable=] is displaying a <{fencedframe}>'s content
+
 <h3 id=csp-integration>CSP Integration</h3>
+
 Fenced frames have a special relationship with Content Security Policies [[!CSP]]. They can only be
-created in secure contexts and require the embedder to give permissions to all sources (i.e.
-"`https:`", "`https://*:*`", or "`*`"). This is done to prevent using [[!CSP]] for cross-channel
-communication.
+created in [=secure contexts=] and require their [=fenced navigable container=]'s
+[=navigable/active document=] to to give permissions to all sources (i.e. "`https:`",
+"`https://*:*`", or "`*`"). This is done to prevent using [[!CSP]] for side-channel communication.
 
 <div algorithm=pre-request-check-csp>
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -865,6 +874,32 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h3 id=csp-integration>CSP Integration</h3>
```
domfarolino, line 877: I think this section might be best to appear under the "Interactions with other specifications" section instead of the "HTML Integration" section.

spec.bs
```diff
@@ -865,6 +874,32 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+Fenced frames have a special relationship with Content Security Policies [[!CSP]]. They can only be
+created in secure contexts and require the embedder to give permissions to all sources (i.e.
```
domfarolino, line 879: Link to secure context dfn

spec.bs
```diff
@@ -865,6 +874,32 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+"`https:`", "`https://*:*`", or "`*`"). This is done to prevent using [[!CSP]] for cross-channel
```
domfarolino, line 880: side-channel, not cross-channel

spec.bs
```diff
@@ -865,6 +874,32 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+communication.
+
+Add a step after step 2 of the [=frame-src pre-request check=] that says:
```
domfarolino, line 883: Put all of these in algorithm blocks so the variables don't leak out.

domfarolino:
Closing and re-opening to maybe get PR preview to work again...

spec.bs
```diff
@@ -880,3 +821,122 @@ specification adds another value for fenced frames:
...
+ * The [=string=] "`*`"
+
+See our <a
+href=https://github.com/WICG/fenced-frame/blob/master/explainer/interaction_with_content_security_policy.md#proposal---scheme-source-matching>CSP
+explainer</a> that describes this.
+
+<h4 id=csp-algorithms>Algorithms</h4>
+
+<div algorithm=pre-request-check-csp>
+  Add a step after step 2 of the [=frame-src pre-request check=] that says:
```
domfarolino, line 850: I get the impression that these algorithms are wrong. Why do we apply these checks only in the `frame-src` directive? What if there is no `frame-src` or `fenced-frame-src` directive supplied in a page's policy, but there is a `default-src` that provides a specific origin? In that case, it seems like the `default-src` directive would execute by the normal rules, instead of our special FF-only rules, thus introducing the communication channel that we're trying to avoid, right? Same goes for if `child-src` is supplied, but not `frame-src` or `fenced-frame-src`.

Let me know if you also think this is a problem, and if so maybe to fix this we need to amend *every* pre-request directive check in the `fenced-frame-src` fallback directive list to abide by the FF-specific rules in the event the request's destination is `fencedframe`. That's a pretty easy change but I want to see if you think I'm on the right path or missing something.

domfarolino, line 850: I agree with doing this after step 2 instead of after step 3. But one thing I want to double check is that _only_ handling the case where the supplied directive's policy _does not_ contain the allowed values is OK. For example what if the supplied policy _does_ contain an allowed policy like `https:` but also contains `'none'` (i.e., `Content-Security-Policy: fenced-frame-src https: 'none'`). Your modifications to the `frame-src` directive will result in CSP spec invoking https://w3c.github.io/webappsec-csp/#match-url-to-source-list, which seems to iterate over each source expression, and _if any_ of the match, we return "Matches". So I think there's no issue here, but I just want to double check that my reading matches yours.

spec.bs
```diff
@@ -880,3 +821,122 @@ specification adds another value for fenced frames:
...
+  Add a step after step 1 in the [=Is response to request blocked by context's required CSP?=]
+  algorithm that reads:
+
+  2. If <var ignore>context</var>'s [=required csp=] is not `null`, and <var ignore>request</var>
+    [=request/destination=] is "`fencedframe`", return "`Blocked`".
+</div>
+
+<h4 id=new-csp-directive>New fenced-frame-src [[!CSP]] [=directive=]</h4>
+
+Fenced frames are a different element from an iframe. Therefore, using the
```
domfarolino, line 879: Somewhere in this section we'll need to modify https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request right?

domfarolino (changes requested):

spec.bs
domfarolino: > if so maybe to fix this we need to amend every pre-request directive check in the `fenced-frame-src` fallback directive list to abide by the FF-specific rules in the event the request's destination is `fencedframe`

Upon further thought, maybe this is correct as-is, since the `child-src` and `default-src` pre-request checks both defer to _other_ directives (see their step 3) which in this case will defer to `frame-src` and `fenced-frame-src` pre-request checks which both are modified by this spec. Is that the intention?

blu25: https://w3c.github.io/webappsec-csp/#directive-fallback-list I think that's right.

Yes. The intention is to have both frame-src and fenced-frame-src properly gated for fenced frames. Since we don't want people to be able to bypass restrictions we're adding simply by changing the CSP directive.

blu25: Yes, that matches what I saw. Step 4.1 returns early once it finds one match, so I think this logic is sound.

domfarolino: I understand that I'm just making sure we're not missing anything by _not_ gating any other algorithms i.e., the more general `default-src`, `child-src`, or even the actual algorithms that do the matching. It looks like we're fine though.

domfarolino (approved):

PR #73, opened 2023-04-13
Local scheme navigation policy container patch

blu25:
When navigating to a URL with a local scheme (URLs like about:blank), iframes clone their policy container from the policy container of the navigation's initiating document. We don't want fenced frames to be able to do that, since that will tell the fenced frame information about a policy container that lives past a fenced frame boundary. This PR adds a patch to stop that inheritance from happening.

(Implementation detail) Note that with the Chromium's MPArch implementation of fenced frames, we get that behavior for free because the initiator frame token is not passed in to the policy container builder during the navigation request. However, we want to make sure that any other implementations also have this behavior, which is why this patch is needed.

The main algorithm that is patched is [determine navigation params policy container](https://html.spec.whatwg.org/multipage/browsers.html#determining-navigation-params-policy-container). It now takes a new "fenced" parameter that prevents cloning the initiator policy container if set to true. There are 3 places where this function is called:

- [Within the beginning navigation algorithm](https://html.spec.whatwg.org/multipage/browsing-the-web.html#beginning-navigation:determining-navigation-params-policy-container), which does not need a patch because the part of the algorithm where this is hit only applies to object and embed elements (it will never hit for fenced frames).
- [Within “create navigation params from a srcdoc resource”](https://html.spec.whatwg.org/multipage/browsing-the-web.html#populating-a-session-history-entry:determining-navigation-params-policy-container), which does not need a patch because srcdoc is not supported in fenced frames.
- [Within create navigation params by fetching](https://html.spec.whatwg.org/multipage/browsing-the-web.html#populating-a-session-history-entry:determining-navigation-params-policy-container-2), which **does** need a patch because this is the path that is taken when fetch is called to navigate a fenced frame.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/73.html" title="Last updated on Jun 22, 2023, 2:08 PM UTC (5b23c1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/73/fb89d8c...5b23c1d.html" title="Last updated on Jun 22, 2023, 2:08 PM UTC (5b23c1d)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -22,6 +22,22 @@ WPT Display: open
 
 <pre class="link-defaults">
 </pre>
+<pre class="biblio">
+{
+  "private-network-access": {
+    "authors": [
+      "Titouan Rigoudy"
+    ],
+    "href": "https://wicg.github.io/private-network-access",
+    "title": "Private Network Access",
+    "status": "CG-DRAFT",
+    "publisher": "WICG",
+    "deliveredBy": [
+      "https://wicg.io/"
+    ]
+  }
+}
+</pre>
 <pre class="anchors">
 urlPrefix: https://www.ietf.org/rfc/rfc4122.txt
   type: dfn; text: urn uuid
@@ -151,9 +167,6 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     text: structured header; url: #section-1
     for: structured header
       text: token; url: name-tokens
-spec: local-network-access; urlPrefix: https://wicg.github.io/local-network-access/
-  type: dfn
-    text: Integration with HTML; url: integration-html
 spec: permissions-policy; urlPrefix: https://w3c.github.io/webappsec-permissions-policy
   type: dfn
     text: ASCII-serialized policy directive; url: serialized-policy-directive
@@ -2335,31 +2348,16 @@ achieve the outcomes described in the above explanatory content.
 
 <h3 id=private-network-access-monkeypatch>Private network access</h3>
 
-Private network access is a security feature that allows you to control which apps on your device
-can access your local network. On creation, regular iframes can inherit their policy container from
-the navigation initiator, while fenced frames explicitly do not. While this prevents information
-from leaking from an embedder to a fenced frame, this results in a policy container being created
-that causes subresource requests to not do private network access checks. This section patches the
-local network access spec to have subresources be subject to those checks.
+[[Private-Network-Access]] is a security feature that allows you to control which web applications
+on can make [=local network requests=]. On creation, <{iframe}>s inherit their policy container from
+the navigation initiator, while <{fencedframe}>s explicitly do not. While this prevents information
+from leaking into a <{fencedframe}> from its embedder, it also results in a
+[=Document/policy container=] for {{Document}}s inside of a <{fencedframe}> being created that
+causes [=subresource requests=] to not do private network access checks. This section patches the
+[[Private-Network-Access]] specification to have subresources be subject to those checks.
 
-<div algorithm=pna-html-changes>
-  Modify the [=Integration with HTML=] section. Rewrite step 3 to read:
-
-  3. An additional step is added to the [=create a policy container from a fetch response=]
-    algorithm:
-
-    1. If |environment| is not null, |environment|'s [=environment/target browsing context=] is not
-      null, and |environment|'s [=environment/target browsing context=]'s [=browsing context/active
-      document=]'s [=node navigable=]'s [=navigable/traversable navigable=] is a [=fenced navigable
-      container/fenced navigable=], then set |result|'s [=IP address space=] to [=IP address
-      space/public|public=].
-
-      Otherwise, set |result|'s [=IP address space=] to <var ignore>response</var>'s
-      [=response/IP address space=].
-</div>
-
-<div algorithm=pna-clone-policy>
-  Modify the [=Integration with HTML=] section. Rewrite step 2 to read:
+<div algorithm=pna-html-clone-policy>
+  Modify the [[PRIVATE-NETWORK-ACCESS#integration-html]] section. Rewrite step 2 to read:
 
   2. Modify the definition of [=clone a policy container=] to read:
 
@@ -2372,6 +2370,21 @@ local network access spec to have subresources be subject to those checks.
         public|public=].
 
         Otherwise, set |clone|’s [=IP address space=] to |policyContainer|’s [=IP address space=].
+</div>
+
+<div algorithm=pna-html-response-policy>
+  Modify the [[PRIVATE-NETWORK-ACCESS#integration-html]] section. Rewrite step 3 to read:
+
+  3. An additional step is added to the [=create a policy container from a fetch response=]
+    algorithm:
+
+    1. If |environment| is not null and |environment|'s [=environment/target browsing context=]'s
+       [=browsing context/active document=]'s [=node navigable=]'s [=navigable/traversable
+       navigable=] is a [=fenced navigable container/fenced navigable=], then set |result|'s [=IP
+       address space=] to [=IP address space/public=].
+
+       Otherwise, set |result|'s [=IP address space=] to <var ignore>response</var>'s
+       [=response/IP address space=].
 </div>
 
 <div algorithm=pna-determine-params>
--- a/spec.bs
+++ b/spec.bs
@@ -35,6 +35,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: browsers.html
       text: check a navigation response's adherence to its embedder policy; url: check-a-navigation-response's-adherence-to-its-embedder-policy
       text: queue a cross-origin embedder policy inheritance violation; url: queue-a-cross-origin-embedder-policy-inheritance-violation
+      text: create a policy container from a fetch response; url: creating-a-policy-container-from-a-fetch-response
     urlPrefix: dom.html
       text: categories; url: concept-element-categories
       text: contexts in which this element can be used; url: concept-element-contexts
@@ -60,6 +61,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: initialize the navigable; url: initialize-the-navigable
       text: node navigable; url: node-navigable
       text: system visibility state; url: system-visibility-state
+      text: active document; url: active-document
       for: navigable
         text: active session history entry; url: nav-active-history-entry
         text: current session history entry; url: nav-current-history-entry
@@ -89,6 +91,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: nav-history-apis.html
       for: Window
         text: navigable; url: window-navigable
+    urlPrefix: webappapis.html
+      text: target browsing context; url:concept-environment-target-browsing-context
 spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: queue a cross-origin embedder policy CORP violation report; url: queue-a-cross-origin-embedder-policy-corp-violation-report
@@ -97,6 +101,13 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     text: structured header; url: #section-1
     for: structured header
       text: token; url: name-tokens
+spec: LNA; urlPrefix: https://wicg.github.io/local-network-access/
+  type: dfn
+    text: Integration with HTML; url: integration-html
+    text: IP address space; url: policy-container-ip-address-space
+    for: response
+      text: IP address space; url: response-ip-address-space
+    text: ip-address-space-public; url: ip-address-space-public
 </pre>
 
 <style>
@@ -880,3 +891,20 @@ specification adds another value for fenced frames:
 
 : "`fencedframe`"
 :: This [=navigable=] is displaying a <{fencedframe}>'s content
+
+<h3 id=local-network-access-monkeypatch>Local Network Access</h3>
+
+<div algorithm=pna-html-changes>
+  Modify the [=Integration with HTML=] section. Rewrite step 3 to read:
+
+  3. An additional step is added to the [=create a policy container from a fetch response=]
+    algorithm:
+
+    1. If |environment| is not null, |environment|'s [=target browsing context=] is not null, and
+      |environment|'s [=target browsing context=]'s [=active document=]'s [=node navigable=]'s
+      [=navigable/traversable navigable=] is a [=fenced navigable container/fenced navigable=], then
+      set |result|'s [=IP address space=] to [=ip-address-space-public|public=].
+
+      Otherwise, set |result|'s [=IP address space=] to <var ignore>response</var>'s
+      [=response/IP address space=].
+</div>
--- a/spec.bs
+++ b/spec.bs
@@ -2373,7 +2373,7 @@ achieve the outcomes described in the above explanatory content.
   /fenced-frame/permission-notification.https.html
 </wpt>
 
-<h3 id=private-network-access-monkeypatch>Private network access</h3>
+<h3 id=private-network-access-monkeypatch>Private Network Access</h3>
 
 [[Private-Network-Access]] is a security feature that allows you to control which web applications
 on can make [=local network requests=]. On creation, <{iframe}>s inherit their policy container from
@@ -2393,8 +2393,8 @@ causes [=subresource requests=] to not do private network access checks. This se
 
     An additional step is added to the algorithm:
 
-      1. If |fenced| is set to true, then set |clone|’s [=IP address space=] to [=IP address space/
-        public|public=].
+      1. If |fenced| is set to true, then set |clone|’s [=IP address space=] to [=IP address
+         space/public=].
 
         Otherwise, set |clone|’s [=IP address space=] to |policyContainer|’s [=IP address space=].
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -2377,7 +2377,7 @@ achieve the outcomes described in the above explanatory content.
 
 [[Private-Network-Access]] is a security feature that regulates when and how [=local network
 requests=] can be made. On [[HTML#creating-browsing-contexts|creation]], <{iframe}>s inherit their
-[=Document/policy container=] from their [=navigable container=]'s [=node document=], while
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1563,3 +1576,26 @@ specification adds another value for fenced frames:
 
 : "`fencedframe`"
 :: This [=navigable=] is displaying a <{fencedframe}>'s content
+
+<h3 id=local-network-access-monkeypatch>Local Network Access</h3>
+
+Local network access is a security feature that allows you to control which apps on your device can
+access your local network. Regular iframes inherit their policy container from the navigation
+initiator, while fenced frames explicitly do not. This results in a policy container being created
+that causes subresource requests to not do PNA checks. This section patches the PNA spec to have
```
domfarolino, line 1585: Let's replace "PNA" with `[[LNA]]`, or spell the whole name out.

spec.bs
```diff
@@ -1563,3 +1576,26 @@ specification adds another value for fenced frames:
 
 : "`fencedframe`"
 :: This [=navigable=] is displaying a <{fencedframe}>'s content
+
+<h3 id=local-network-access-monkeypatch>Local Network Access</h3>
+
+Local network access is a security feature that allows you to control which apps on your device can
+access your local network. Regular iframes inherit their policy container from the navigation
+initiator, while fenced frames explicitly do not. This results in a policy container being created
```
domfarolino, line 1584: > Regular iframes inherit their policy container from the navigation initiator

Is this true? When they are _created_ I _think_ that's the case, but otherwise their policy container is informed from the [response creation algorithm](https://html.spec.whatwg.org/C#creating-a-policy-container-from-a-fetch-response), right?

spec.bs
```diff
@@ -119,11 +121,22 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    text: queue a cross-origin embedder policy CORP violation report; url: queue-a-cross-origin-embedder-policy-corp-violation-report
 spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
   type: dfn
     text: structured header; url: #section-1
     for: structured header
       text: token; url: name-tokens
+spec: LNA; urlPrefix: https://wicg.github.io/local-network-access/
+  type: dfn
+    text: Integration with HTML; url: integration-html
+    text: IP address space; url: policy-container-ip-address-space
```
domfarolino, line 136: This is exported right? https://github.com/WICG/local-network-access/blob/main/index.src.html#L815.

In any case, this is malformed as this link is for the "new policy container IP address member". Either we should add the appropriate `for: policy container` line above (like you do below with `response`) or change the url to a different link. Or just get rid of it altogether if it is indeed exported as I think it is.

spec.bs
```diff
@@ -119,11 +121,22 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    for: response
+      text: IP address space; url: response-ip-address-space
```
domfarolino, line 138: Is exported https://github.com/WICG/local-network-access/blob/main/index.src.html#L632

domfarolino (commented):
First pass. Also it looks like there are merge conflicts now as a few things have shifted out from under.

spec.bs
blu25: Correct. This is just talking about the creation case. Updated the comment to explicitly specify that.

domfarolino:
Is there anything with PNA that needs to be addressed in the navigation case? Perhaps only for `about:blank` navigations but maybe more, I can't recall exactly (though I thought there was some work to be done in that area).

blu25:
> Is there anything with PNA that needs to be addressed in the navigation case? Perhaps only for `about:blank` navigations but maybe more, I can't recall exactly (though I thought there was some work to be done in that area).

From what I understand, we only need to make sure that fenced frames aren't inheriting the policy container's IP address space from its embedder. I don't think that happens during the fetching process.

Though, looking at this again, I'm also not entirely convinced that this where the gating should happen.

spec.bs
```diff
@@ -2016,3 +2015,27 @@ achieve the outcomes described in the above explanatory content.
   /fenced-frame/permission-geolocation.https.html
   /fenced-frame/permission-notification.https.html
 </wpt>
+
+<h3 id=local-network-access-monkeypatch>Local Network Access</h3>
```
domfarolino, line 2019: It looks like the spec is maybe being renamed back to PNA (see https://github.com/WICG/local-network-access/issues/91#issuecomment-1494704528 and https://github.com/WICG/local-network-access/issues/91#issuecomment-1540157575), can you update your PR accordingly?

domfarolino:
> From what I understand, we only need to make sure that fenced frames aren't inheriting the policy container's IP address space from its embedder. I don't think that happens during the fetching process.

For navigations to local schemes I think it certainly can, right? https://html.spec.whatwg.org/C#populating-a-session-history-entry:determining-navigation-params-policy-container-2. So I think this is on the right track.

We also need to prevent the inheritance for initial Document creation, but that looks like it's being done mostly in https://wicg.github.io/fenced-frame/#creating-browsing-contexts-patch, which we should probably allude to in your PR, since that takes care of the creation case.

spec.bs
```diff
@@ -2238,3 +2240,85 @@ achieve the outcomes described in the above explanatory content.
   /fenced-frame/permission-geolocation.https.html
   /fenced-frame/permission-notification.https.html
 </wpt>
+
+<h3 id=private-network-access-monkeypatch>Private network access</h3>
+
+Private network access is a security feature that allows you to control which apps on your device can
+access your local network. On creation, regular iframes inherit their policy container from the
+navigation initiator, while fenced frames explicitly do not. While this prevents information from
```
domfarolino, line 2248: Do they _always_ get it from the initiator? What part of the spec are you referencing. For example, https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=4933-4941;drc=41b3f66441ca114b9b75194a635b9db7b588b772 contradicts this.

blu25: Not always. I'm referring to https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context step 15.2 "Set document's [policy container](https://html.spec.whatwg.org/multipage/dom.html#concept-document-policy-container) to a [clone](https://html.spec.whatwg.org/multipage/browsers.html#clone-a-policy-container) of creator's [policy container](https://html.spec.whatwg.org/multipage/dom.html#concept-document-policy-container).".

Should this say "can inherit" instead?

spec.bs
```diff
@@ -95,6 +96,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       for: document state
         text: document; url: document-state-document
       text: checking if unloading is user-canceled
+      text: history policy container; url: document-state-history-policy-container
+      text: source policy container; url: source-snapshot-params-policy-container
```
domfarolino, line 100: should be under source snapshot params

spec.bs
```diff
@@ -95,6 +96,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       for: document state
         text: document; url: document-state-document
       text: checking if unloading is user-canceled
+      text: history policy container; url: document-state-history-policy-container
```
domfarolino, line 99: should be under document state

spec.bs
blu25: I'm going to change this to "can inherit" and resolve this conversation. Feel free to re-open if you think there's better wording for this.

domfarolino (approved):
This pretty much LGTM after my adjustments, % one question here: https://github.com/WICG/fenced-frame/pull/73#pullrequestreview-1446300154 that needs to be answered.

spec.bs
```diff
@@ -2357,3 +2372,88 @@ achieve the outcomes described in the above explanatory content.
...
+
+<h3 id=private-network-access-monkeypatch>Private Network Access</h3>
+
+[[Private-Network-Access]] is a security feature that regulates when and how [=local network
+requests=] can be made. On [[HTML#creating-browsing-contexts|creation]], <{iframe}>s inherit their
+[=Document/policy container=] from their [=navigable container=]'s [=node document=], while
+<{fencedframe}>s do not. While this prevents information from leaking into a <{fencedframe}> from
+its embedder, it also results in a [=Document/policy container=] for {{Document}}s inside of a
+<{fencedframe}> being created that causes [=subresource requests=] to not do private network access
+checks. This section patches the [[Private-Network-Access]] specification to make subresources be
```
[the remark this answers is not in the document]
blu25: From https://docs.google.com/document/d/1Fy9CC5w1-8C1s2YfRioeURUVN5I3u3jA5o_87kH9-fA/edit#heading=h.uirj6a11h4rr:

"This currently results in the use of a default PolicyContainer, for which the IP address space is `kUnknown`. Due to the existing carveout for `kUnknown`, this means fenced frames’ subresource fetches are not subject to PNA checks."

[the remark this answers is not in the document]
domfarolino: Sure but that sounds like an implementation detail that doesn't exist in the specs, since `kUnknown` is not a property in the spec and the default policy container value is always public: https://wicg.github.io/local-network-access/#policy-container-ip-address-space

spec.bs
```diff
@@ -2357,3 +2372,89 @@ achieve the outcomes described in the above explanatory content.
...
+
+<h3 id=private-network-access-monkeypatch>Private Network Access</h3>
+
+[[Private-Network-Access]] is a security feature that regulates when and how [=local network
+requests=] can be made. On [[HTML#creating-browsing-contexts|creation]], <{iframe}>s inherit their
+[=Document/policy container=] from their [=navigable container=]'s [=Node/node document=], while
+<{fencedframe}>s do not. While this prevents information from leaking into a <{fencedframe}> from
+its embedder, it also results in a [=Document/policy container=] for {{Document}}s inside of a
+<{fencedframe}> being created that causes [=subresource requests=] to not do private network access
+checks. This section patches the [[Private-Network-Access]] specification to make subresources be
```
domfarolino, line 2384: Also can you clarify "private network access checks" here? What does this mean? There is the normal preflight which conditionally now deals with PNA headers depending on whether the request is a local network request, is that what you mean? Or specifically the https://wicg.github.io/local-network-access/#local-network-access-check? Or both?

blu25:
After talking with Titouan offline, I think we need to be patching more than just PNA. I don't think we want to inherit any part of the policy container from a parent for local navigations, so I've rewritten the spec to reflect that. Instead of just setting the IP address space, now it will only inherit the policy container from the initiator if the navigation is a local navigation not in a fenced frame. I think this is simpler and what we want the behavior to be in the end. WDYT?

spec.bs
```diff
@@ -1264,6 +1263,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
 document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
 container=], across the fenced frame boundary.
 
+<h3 id=local-policy-container-monkeypatch>Local Requests Policy Container Patch</h3>
+
+When making local requests, <{iframe}>s inherit their [=Document/policy container=] from their
+[=navigable container=]'s [=Node/node document=]. If fenced frames were to do the same thing, that
+would allow information about the initiator's [=IP address space=] to leak across a fenced frame
+boundary. his section patches [=Document/policy container=] inheritance to close that leak.
```
domfarolino, line 1271: This seems like a typo

spec.bs
```diff
@@ -1264,6 +1263,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
 document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
 container=], across the fenced frame boundary.
 
+<h3 id=local-policy-container-monkeypatch>Local Requests Policy Container Patch</h3>
+
+When making local requests, <{iframe}>s inherit their [=Document/policy container=] from their
+[=navigable container=]'s [=Node/node document=]. If fenced frames were to do the same thing, that
```
domfarolino, line 1269: Can you link to `<{fencedframe}>` here?

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
 document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
 container=], across the fenced frame boundary.
 
+<h3 id=local-policy-container-monkeypatch>Local Requests Policy Container Patch</h3>
+
+When making local requests, <{iframe}>s inherit their [=Document/policy container=] from their
```
domfarolino, line 1348: "local requests" is ambiguous. Do you mean "local requests" as in https://wicg.github.io/local-network-access/#local-network-request, since this is a PNA-focused PR, or do you mean requests whose scheme is local: https://fetch.spec.whatwg.org/#local-scheme. I thought the latter, but the terminology you chose (but didn't link to) is very reminiscent of the former, but inconsistent with your usage of "is local" below, which you fail to link to.

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
...
+[=navigable container=]'s [=Node/node document=]. If <{fencedframe}>s were to do the same thing,
+that would allow information about the initiator's [=IP address space=] to leak across a fenced
```
domfarolino, line 1350: Certainly it would be the entire policy container that would leak, right? Not just the address space.

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
...
+frame boundary. This section patches [=Document/policy container=] inheritance to close that leak.
+
+<div algorithm=local-policy-inheritance-determine-params>
+  Modify the definition of the [=determine navigation params policy container=] algorithm to read:
```
domfarolino, line 1354: "definition of the" isn't needed, and is inconsistent with wording elsewhere

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
...
+
+  To determine navigation params policy container given a [=URL=] |responseURL|; four [=policy
+  container=]-or-nulls <var ignore>historyPolicyContainer</var>, |initiatorPolicyContainer|,
+  <var ignore>parentPolicyContainer</var>, and <var ignore>responsePolicyContainer</var>; and an
+  optional [=boolean=] |fenced| that defaults to false:
+
+  Rewrite step 3 to read:
+
+    3. If |responseURL| is local, |initiatorPolicyContainer| is not null, and |fenced| is set to
+       false, then return a [=clone a policy container|clone=] of |initiatorPolicyContainer|.
```
domfarolino, line 1364: Add a note below this algorithm to mention that we do not have to modify the `about:srcdoc` case (and please link to the definition correctly with the right formatting: https://html.spec.whatwg.org/#about:srcdoc) because fenced frames cannot navigate to srcdoc resources.

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
...
+  Rewrite step 23 (now step 24) to read:
+
+  24. Let <var ignore>resultPolicyContainer</var> be the result of [=determining navigation params
+      policy container=] given <var ignore>response</var>'s URL, <var ignore>entry</var>'s
+      [=document state=]'s [=document state/history policy container=], <var
+      ignore>sourceSnapshotParams</var>'s [=source snapshot params/source policy container=], null,
+      <var ignore>responsePolicyContainer</var>, and |fenced|.
+</div>
+
+Note: <{fencedframe}> [=policy container=] "inheritance" upon initial {{Document}} creation (and its
```
domfarolino, line 1385: Remove the scare quotes around inheritance

spec.bs
```diff
@@ -1340,6 +1343,49 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
...
+  Modify the definition of the [=determine navigation params policy container=] algorithm to read:
+
+  To determine navigation params policy container given a [=URL=] |responseURL|; four [=policy
+  container=]-or-nulls <var ignore>historyPolicyContainer</var>, |initiatorPolicyContainer|,
+  <var ignore>parentPolicyContainer</var>, and <var ignore>responsePolicyContainer</var>; and an
+  optional [=boolean=] |fenced| that defaults to false:
+
+  Rewrite step 3 to read:
+
+    3. If |responseURL| is local, |initiatorPolicyContainer| is not null, and |fenced| is set to
```
domfarolino, line 1363: We absolutely have to link to "is local" here. Please check to see if you are missing any other important links.

spec.bs
blu25: I mean local scheme navigations. I'll update the section.

spec.bs
```diff
@@ -1340,6 +1342,53 @@ Note: This is because we need to ensure that we do not leak <var ignore>creator<
 document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
 container=], across the fenced frame boundary.
 
+<h3 id=local-policy-container-monkeypatch>Local requests policy container patch</h3>
+
+When making [=local scheme=] requests, <{iframe}>s inherit their [=Document/policy container=] from
```
domfarolino, line 1347: This reads as if you're targeting subresource requests, but I believe this only applies to navigation requests, right? Hence all the changes to "navigation params" stuff below?

domfarolino (changes requested):

spec.bs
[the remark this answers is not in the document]
blu25: Resolving because this is an outdated comment.

spec.bs
blu25: Resolving because the spec is now more generalized and this comment is outdated.

spec.bs
blu25: I've updated this to say "clone their [policy container](https://html.spec.whatwg.org/multipage/dom.html#concept-document-policy-container) from the [policy container](https://html.spec.whatwg.org/multipage/dom.html#concept-document-policy-container) of the [Document](https://dom.spec.whatwg.org/#document) initiating the [navigation request](https://fetch.spec.whatwg.org/#navigation-request)". This matches what's happening in [determine navigation params policy container](https://html.spec.whatwg.org/multipage/browsers.html#determining-navigation-params-policy-container) with the _initiatorPolicyContainer_ piece.

domfarolino (approved):
Looks good, I've made some very small editorial adjustments but nothing major. Thanks for the big improvements.

PR #74, opened 2023-04-13
Add permissions patches

blu25:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/74.html" title="Last updated on Apr 30, 2023, 12:35 PM UTC (cad5853)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/74/0582b0e...cad5853.html" title="Last updated on Apr 30, 2023, 12:35 PM UTC (cad5853)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -359,7 +359,7 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
 The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, determines the [=fenced
 container policy=] that will be used when the [=Document/permissions policy=] for a {{Document}} in
 the <{fencedframe}>'s [=fenced navigable container/fenced navigable=] is initialized. Its value must
-be a [=serialized permissions policy=].
+be a [=serialized permissions policy=]. [[!PERMISSIONS-POLICY]]
 
 The IDL attribute <dfn attribute for=HTMLFencedFrameElement>allow</dfn> must [=reflect=] the
 respective content attribute of the same name.
--- a/spec.bs
+++ b/spec.bs
@@ -74,8 +74,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: cross-origin isolation mode; url: bcg-cross-origin-isolation
       for: cross-origin isolation mode
         text: none; url:cross-origin-isolation-none
-    urlPrefix: document-lifecycle.html
-      text: create and initialize a Document object; url: initialise-the-document-object
     urlPrefix: browsing-the-web.html
       text: create navigation params by fetching; url: create-navigation-params-by-fetching
       text: document state; url: she-document-state
@@ -153,6 +151,10 @@ spec: permissions-policy; urlPrefix: https://w3c.github.io/webappsec-permissions
     text: ASCII-serialized policy directive; url: serialized-policy-directive
     text: serialized permissions policy; url: serialized-permissions-policy
     text: supported features; url: supported-features
+    text: Define an inherited policy for feature in container at origin; url: define-inherited-policy-in-container
+    text: Is feature enabled in document for origin?; url: is-feature-enabled
+    text: declared policy; url: declared-policy
+    text: allowlist; url: allowlist
     for: permissions
       text: matches; url: matches
       text: permissions policy; url: permissions-policy
@@ -1782,42 +1784,24 @@ can only load if it has opted into all of the [=fenced frame config/effective en
 <h4 id=permissions-policy-patches>Algorithm patches</h4>
 
 <div algorithm=create-permissions-policy>
-  Rewrite step 1 of the [=Create a Permissions Policy for a navigable=] algorithm to read:
+  Modify the definition of [=Create a Permissions Policy for a navigable=]:
+
+  Given null or an [=element=] (|container|), an [=origin=] (|origin|), and an optional [=boolean=]
+  |fenced| that defaults to false, this algorithm returns a new Permissions Policy.
+
+  Rewrite step 1 to read:
 
   1. [=Assert=]: if not null, <var ignore>container</var> is either a [=navigable container=] or a
      [=fenced navigable container=].
-</div>
 
-<div algorithm=create-fenced-permissions-policy>
-  Create a new algorithm called <dfn>Create a Permissions Policy for a fenced navigable</dfn> in
-  [[!PERMISSIONS-POLICY]].
+  Rewrite step 4 to read:
 
-  Given null or an element (|container|) and an optional [=fencedframetype/exhaustive set of
-  permissions=] (|effective permissions|), this algorithm returns a new [=Document/permissions
-  policy=].
+  4. For each |feature| supported:
 
-  1. [=Assert=]: if not null, |container| is a [=fenced navigable container=].
+     1. Let |isInherited| be the result of running [=Define an inherited policy for feature in
+        container at origin=] on |feature|, |container|, |origin|, and |fenced|.
 
-  2. Let |inherited policy| be a new [=ordered map=].
-
-  4. If |effective permissions| is given, then [=list/For each=] |supported feature| of the UA's
-     [=supported features=]:
-
-       1. If |supported feature| [=list/exists=] in |effective permissions|, then set |inherited
-          policy|[|supported feature|] to "`Enabled`".
-
-          Otherwise, set |inherited policy|[|supported feature|] to "`Disabled`".
-
-  5. Let |policy| be a new [=permissions/Permissions policy=], with [=permissions policy/inherited
-     policy=] |inherited policy| and declared policy a new [=ordered map=].
-
-  6. Return |policy|.
-
-  Note: It is safe to naively set the frame's permissions to the |effective permissions|. By this
-  point, the [=Should navigation response to navigation request be blocked by Permissions Policy?=]
-  check will have verified that the fenced frame is not trying to get extra permissions that the
-  embedder didn't delegate.
-
+     2. Set <var ignore>inherited policy</var>[|feature|] to |isInherited|.
 </div>
 
 <div algorithm=allow-attribute-fenced-frame>
@@ -1843,45 +1827,18 @@ can only load if it has opted into all of the [=fenced frame config/effective en
   Modify the definition of [=Create a Permissions Policy for a navigable from response=] to read:
 
   Given null, a [=navigable container=], or a [=fenced navigable container=] (|container|), an
-  [=origin=] (|origin|), a [=response=] (<var ignore>response</var>), and null or an
-  [=fenced frame config instance/effective enabled permissions=] (|effective permissions|), this
-  algorithm returns a new [=Document/permissions policy=].
+  [=origin=] (|origin|), and a [=response=] (<var ignore>response</var>), this algorithm returns a
+  new [=Document/permissions policy=].
 
-  Modify step 1 of the algorithm to read:
-  1. If |container| is a [=fenced navigable container=], then let <var ignore>policy</var> be the
-     result of running
-     [=Create a Permissions Policy for a fenced navigable=] given |container| and
-     |effective permissions|.
-  
-     Otherwise, let <var ignore>policy</var> be the result of running [=Create a Permissions Policy
-     for a navigable=] given |container| and |origin|.
-</div>
+  Add a step before step 1 that reads:
 
-<div algorithm=shared-document-creation-changes>
-  Modify the [=create and initialize a Document object=] algorithm. Rewrite step 3 to read:
+  1. If |container| is a [=fenced navigable container=], let |fenced| be true.
 
-  3. Let <var ignore>permissionsPolicy</var> be the result of [=Create a Permissions Policy for a
-     navigable from response|creating a permissions policy from a response=] given
-     |navigationParams|'s [=navigable=]'s [=navigable container|container=], |navigationParams|'s
-     [=navigation params/origin=], |navigationParams|'s [=navigation params/response=], and
-     |navigationParams|'s [=navigable=]'s [=fenced frame config instance=]'s [=fenced frame config
-     instance/effective enabled permissions=].
-</div>
+     Otherwise, let |fenced| be false.
 
-<div algorithm=new-browsing-context-changes>
-  Modify the [=create a new browsing context and document=] algorithm. Rewrite step 7 to read:
-
-  7. If <var ignore>navigationParams</var>'s [=navigable=]'s [=navigable/traversable navigable=] is
-     a [=fenced navigable container/fenced navigable=], let |permissionsPolicy| be the result of
-     [=Create a Permissions Policy for a fenced navigable|creating a permissions policy for a fenced
-     navigable=] given |embedder| and null.
-
-     Otherwise, let |permissionsPolicy| be the result of [=Create a Permissions Policy for a
-     navigable|creating a permissions policy=] given |embedder| and
-     <var ignore>origin</var>.
-
-  Note: This change is made in addition to the changes to [=create a new browsing context and
-  document=] outlined in [[#creating-browsing-contexts-patch]].
+  Modify the now step 2 of the algorithm to read:
+  2. Let <var ignore>policy</var> be the result of running [=Create a Permissions Policy
+     for a navigable=] given |container|, |origin|, and |fenced|.
 </div>
 
 <div algorithm=attempt-populate-history-patches>
@@ -1898,19 +1855,19 @@ can only load if it has opted into all of the [=fenced frame config/effective en
   Create a new algorithm called <dfn>Should navigation response to navigation request be blocked by
   Permissions Policy?</dfn> in [[!HTML]].
 
-  Given a [=navigation params=] (|navigationParams|) and a [=navigable=] (|navigable|), this algorithm returns
-  `Blocked` or `Allowed`:
+  Given a [=navigation params=] (|navigationParams|) and a [=navigable=] (|navigable|), this
+  algorithm returns `Blocked` or `Allowed`:
 
   1. If |navigable| is not a [=fenced navigable container/fenced navigable=], then return `Allowed`.
 
   2. Let |origin| be |navigationParams|'s [=navigation params/origin=].
 
   3. Let |effective permissions| be the |navigable|'s [=navigable/fenced frame config instance=]'s
-     [=fenced frame config instance/effective permissions=].
+     [=fenced frame config instance/effective enabled permissions=].
 
   4. Let |permissions policy| be the result of [=Create a Permissions Policy for a navigable|
-     creating a permissions policy=] given |navigable|'s [=navigable container|container=] and
-     |origin|.
+     creating a permissions policy=] given |navigable|'s [=navigable container|container=],
+     |origin|, and <var ignore>fenced</var> set to true.
 
   5. Let |inherited policy| be |permissions policy|'s [=permissions policy/inherited policy=].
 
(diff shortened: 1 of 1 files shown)


blu25:
> Is this ready for my review right now or is it still in draft shape?

This is ready for your review, but is not ready to be taken out of draft. There are a couple TODOs still in the patches that I need your input on how to handle.

spec.bs
```diff
@@ -309,6 +342,10 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
   1. <span class=XXX>TODO</span>
 </div>
 
+The IDL attribute
+<dfn attribute for="fencedframe"><code data-x="dom-fencedframe-allow">allow</code></dfn>
+</dfn> must [=reflect=] the respective content attributes of the same name.
```
domfarolino, line 347: In this PR there is no content attribute. You'll need to add an attribute to this section: https://github.com/WICG/fenced-frame/blob/master/spec.bs#L229

spec.bs
```diff
@@ -108,6 +111,23 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+  type: dfn
+    text: queue a cross-origin embedder policy CORP violation report; url: queue-a-cross-origin-embedder-policy-corp-violation-report
+    text: main fetch; url: main-fetch
+    text: should request be blocked due to a bad port; url: block-bad-port
+spec: mixed-content; urlPrefix: https://w3c.github.io/webappsec-mixed-content/
+  type: dfn
+    text: should fetching request be blocked as mixed content; url: should-block-fetch
+spec: CSP; urlPrefix: https://w3c.github.io/webappsec-csp/
+  type: dfn
+    text: should request be blocked by Content Security Policy; url: should-block-request
```
domfarolino, line 130: This does not seem used?

spec.bs
```diff
@@ -1563,3 +1598,133 @@ specification adds another value for fenced frames:
...
 :: This [=navigable=] is displaying a <{fencedframe}>'s content
+
+<h3 id=permissions-policy-changes>Permissions Policies</h3>
+Permissions are granted to fenced frames through the FencedFrameConfig object, which defines the
+required permissions for the fenced frame to load. The fenced frame is only allowed to navigate if
+all of the required permissions are present in the context loading the fenced frame. When a fenced
+frame loads, it is assumed that the permissions checks passed, and the fenced frame will load with
+the list of permissions that were specified in the config object.
+
+Modify the [=Create a Permissions Policy for a navigable=] algorithm:
```
domfarolino, line 1609: What's up with this?

domfarolino:
> There are a couple TODOs still in the patches that I need your input on how to handle.

Can you be more specific with what you need help on?

spec.bs
```diff
@@ -1563,3 +1598,133 @@ specification adds another value for fenced frames:
...
+</div>
+
+<div algorithm=allow-attribute-fenced-frame>
+  Modify the [=The allow attribute of the iframe element=] section.
+
+  Rename it to "The allow attribute of the iframe and fencedframe element".
+
+  Rewrite the section to read:
+
+  &lt;iframe> and &lt;fencedframe> elements have an "allow" attribute, which contains an
```
domfarolino, line 1651: `<{iframe}>` and `<{fencedframe}>`, per https://speced.github.io/bikeshed/#metadata-markup-shorthands

domfarolino, line 1651: "allow" should be a real link to the associated attributes

spec.bs
```diff
@@ -1563,3 +1598,133 @@ specification adds another value for fenced frames:
...
+  3. Let <var ignore>permissionsPolicy</var> be the result of [=Create a Permissions Policy for a
+    navigable from response|creating a permissions policy from a response=] given
+    |navigationParams|'s [=navigable=]'s [=navigable container|container=], |navigationParams|'s
+    [=navigation params/origin=], |navigationParams|'s [=navigation params/response=], and
+    <span class=XXX>TODO: Populate this with the required permissions to load from the config in
+    the navigation params</span>.
+</div>
+
+<div algorithm=new-browsing-context-changes>
+  Modify the [=create a new browsing context and document=] algorithm. Rewrite step 7 to read:
```
domfarolino, line 1689: Let's briefly mention that these changes are made _in addition to_ the changes made in https://wicg.github.io/fenced-frame/#creating-browsing-contexts-patch (please use https://speced.github.io/bikeshed/#section-links to link to that section).

spec.bs
domfarolino, line 1651: Same here, and elsewhere if applicable, regarding the element and attribute links.

spec.bs
```diff
@@ -1563,3 +1598,133 @@ specification adds another value for fenced frames:
...
+  Rewrite step 1 of the algorithm to read:
+
+  1. Assert: if not null, |container| is either a [=navigable container=] or a
+    [=fenced navigable container=].
+
+  Add a new step after step 1 that reads:
+
+  2. Assert: if |container| is not a [=fenced navigable container=], |required features| is null.
+
+  Rewrite step 4 to read:
```
domfarolino, line 1628: Is this step 4 of the new algorithm (with the above changes) or of the existing algorithm not considering the changes above.

spec.bs
blu25: I think I just forgot to remove that line.

spec.bs
blu25: Step 4 of the existing algorithm. I'm not sure what the convention is for referring to steps after new steps were added before it in a patch.

spec.bs
```diff
@@ -1563,3 +1605,135 @@ specification adds another value for fenced frames:
...
+  [=Document/permissions policy=].
+
+  Modify step 1 of the algorithm to read:
+  1. Let <var ignore>policy</var> be the result of running
+    [=Create a Permissions Policy for a navigable=] given |container|, |origin|, and
+    |required features|.
+</div>
+
+<div algorithm=shared-document-creation-changes>
+  Modify the [=shared document creation infrastructure=] algorithm. Rewrite step 3 to read:
```
domfarolino, line 1684: Just a heads up this isn't actually an algorithm. If you look under this heading in the HTML Standard, it references the ["create and initialize a Document object"](https://html.spec.whatwg.org/C#initialise-the-document-object) algorithm, which is what we should be referencing.

domfarolino:
> I'm not entirely sure how to get access to the fenced frame config.

Let me make sure I understand: these are cases where you are effectively "inside" the fenced frame and you need to access i.e., the "fenced frame properties" implementation-equivalent object in the spec, right? (Basically what we have in DocumentLoader in our implementation inside the FF renderer?) In that case, I think you can do what we do in https://wicg.github.io/fenced-frame/#ref-for-fenced-frame-config-instance%E2%91%A2, including with the XXX box below until I fix that :/

Does that help?

spec.bs
```diff
@@ -1563,3 +1605,135 @@ specification adds another value for fenced frames:
...
+
+<div algorithm=main-fetch-patches>
+  Modify the [=main fetch=] algorithm. Modify step 19 to add a new bullet point to the algorithm
+  list:
+
+  * [=should response to request be blocked by Permissions Policy=]
+</div>
+
+<div algorithm=permissions-policy-block-request>
+  Create a new algorithm in the [[!permissions-policy]] spec, called <dfn export>should response to
```
domfarolino, line 1713: So this is the algorithm that sees whether or not a fenced frame has all of the permissions that the "config instance" (read: fenced frame properties) demands that it have, right? In that case, I think we should not have this be called by Fetch, but instead basically be one of the many algorithms that we call before we finalize the navigation, that can lead us down the error path. See for example: https://wicg.github.io/fenced-frame/#ref-for-attempt-to-populate-the-history-entry's-document, which basically references algorithms like https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document:~:text=Otherwise%2C%20if%20the%20result%20of%20checking%20a%20navigation%20response%27s%20adherence%20to%20%60X%2DFrame%2DOptions%60, whose placement I think we want to mimic with the `S-L-M` header checks and probably the permissions policy checks as well, especially since these are navigation-specific checks.

domfarolino:
Please feel free to mark all previous discussions has "resolved" with the GitHub UI just so I'm sure which ones still need attention vs which ones are indeed addressed.

spec.bs
```diff
@@ -601,6 +640,9 @@ A <dfn export>fenced frame config</dfn> is a struct with the following [=struct/
 
   : <dfn>embedder shared storage context</dfn>
   :: null, or an [=fencedframetype/embedder shared storage context=]
+
+  : <dfn>required permissions to load</dfn>
```
gtanzer, line 644: We don't need this field (yet). This is the point of [=fenced frame config/effective permissions=].

"required permissions to load" as a name doesn't capture the fact that it disables all other permissions anyway.

spec.bs
```diff
@@ -1701,3 +1748,147 @@ specification is printed below:
...
+  <{iframe/allow}>; <{fencedframe}>: <{fencedframe/allow}>), which contain an [=ASCII-serialized
+  policy directive=].
+
+  The allowlist for the features named in the attribute may be empty; in that case, the default
+  value for the allowlist is "`src`", which represents the origin of the URL in the iframe’s src
+  attribute, or the fencedframe's [=fenced frame config=].
+
+  When not empty, the <{iframe/allow}> attribute will result in adding an allowlist for each
+  recognized feature to the iframe element’s content navigable's container policy or the fencedframe
+  element's [=fenced navigable container/fenced navigable=]'s container policy, when it is
```
domfarolino, line 1816: All of these references to container policy should probably be updated to point to either the already-existing "container policy" or the new "fenced container policy" when appropriate. This all assumes that we should introduce the concept of a "fenced container policy" that's independent of "container policy", instead of just modifying the existing "container policy definition accordingly. I'll let you decide which is best to do.

spec.bs
```diff
@@ -1717,3 +1764,147 @@ specification is printed below:
...
+  Add a new step after step 1 that reads:
+
+  2. [=Assert=]: if |container| is not a [=fenced navigable container=], |required features| is not
+     given.
+
+  Rewrite step 4 to read:
+
+  4. If |container| is a [=fenced navigable container=], then for each |feature| supported:
+
+       1. If |feature| [=list/exists=] in |required features|, set |inherited policy|[|feature|] to
```
domfarolino, line 1804: I think this is broken as-is right now. Your later modifications to https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context pass in nothing as the `|required policies|` variable, but it seems that when `|container|` is a fenced navigable container, we expect `|required features|`. We should fix this and in the process decide whether:
 - The initial `about:blank` Document in a fenced frame should have _all_ policies manually disabled; or...
 - Or inherit them from "beyond the fence" as iframes would normally do

The first option seems like the safest, and is kinda not web-observable except for with extensions I guess. But I'd love to know what our implementation does.

spec.bs
```diff
@@ -1717,3 +1764,147 @@ specification is printed below:
...
+
+  8. Otherwise, if the result of [=Should navigation response to navigation request be blocked by
+    Permissions Policy?=] given |navigationParams|'s [=request=],
+    |navigationParams|'s [=response=], and <var ignore>navigable</var> is "`Blocked`", then set
+    |failure| to true.
+</div>
+
+<div algorithm=permissions-policy-block-request>
+  Create a new algorithm called <dfn>Should navigation response to navigation request be blocked by
+  Permissions Policy?</dfn> in [[!HTML]].
```
domfarolino, line 1885: Is the intention of this algorithm really to not consider any of the permissions policies in the response? My understanding is that we wanted to make sure that the required permissions were not blocked by either:
 1. A parent document's permissions that were inherited in the disabled state
 2. Response headers in the actual response

Do we not care about (2) above? Is the only piece of the response that we need, the origin?

spec.bs
```diff
@@ -1717,3 +1764,145 @@ specification is printed below:
...
+     instance=]'s [=fenced frame config instance/required permissions to load=].
+
+  3. Let |inherited policy| be |navigable|'s [=navigable/active document=]'s [=Document/permissions
+     policy=]'s [=inherited policy=].
+
+  4. Let |origin| be |response|'s [=response/url=]'s [=url/origin=].
+
+  5. [=list/For each=] |permission| of |required permissions to load|:
+
+    1. If |permission| does not exist in |inherited policy|, and |permission|'s
```
domfarolino, line 1901: My understanding is that `inherited policy` will be completely full of all supported features, so it should always exist, per https://w3c.github.io/webappsec-permissions-policy/#inherited-policy-for-a-feature.

Do you mean to check if `|inherited policy|[|permission|] is "Disabled"` instead?

spec.bs
blu25: I agree with doing option 1. That will match the behavior in our implementation once the required permissions CL is landed. If an empty permissions list is passed into PermissionsPolicy::CreateForFencedFrame, it will disable every feature.

spec.bs
blu25: Re: response permissions policies, [we have a WPT](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/fenced-frame/default-enabled-features-attribute-allow.https.html;l=25-35;drc=3ccc9b67e688ec5f98f6664385f0587fb724d307) to check that we can further restrict permissions after a fenced frame loads

Re: the wildcard, [our implementation](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=8367-8384;drc=dfb15f3bc92332c4f433ad9721ba1a16ce2595cc) matches what's in the spec.
`the policy must be either explicitly enabled for 'origin', or the policy must by default be enabled for all origins`

spec.bs
blu25: My vote is to stick with "container policy". The container policies themselves behave the exact same between fenced frames and iframes. The difference is how external algorithms handle them. So I think we'd just be making a duplicate of "container policy" with "fenced container policy".

spec.bs
```diff
@@ -1701,3 +1748,147 @@ specification is printed below:
...
+
+Permissions are granted to {{Document}} inside of <{fencedframe}>s through the
+{{FencedFrameConfig}} object and its associated internal [=fencedframeconfig/config=], which defines
+the permissions [=fenced frame config/required permissions to load|required=] for a <{fencedframe}>
+to navigate successfully. Specifically, a {{Document}} inside of a <{fencedframe}> can only load if
+it has opted into all of the [=fenced frame config/required permissions to load=].
+
+<h4 id=fenced-container-policies>Fenced container policies</h4>
+
+TODO: define <dfn>fenced container policy</dfn> here, which is current referenced by
```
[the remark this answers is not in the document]
blu25: I'm going to stick with the existing container policy definition. I don't think we need a new "fenced container policy" here.

spec.bs
```diff
@@ -1717,3 +1755,160 @@ specification is printed below:
...
+<div algorithm=permissions-policy-block-request>
+  Create a new algorithm called <dfn>Should navigation response to navigation request be blocked by
+  Permissions Policy?</dfn> in [[!HTML]].
+
+  Given a [=response=] (|response|) and a [=navigable=] (|navigable|), this algorithm returns
+  `Blocked` or `Allowed`:
+
+  1. If |navigable| is not a [=fenced navigable container/fenced navigable=], then return `Allowed`.
+
+  2. Let |origin| be |response|'s [=response/url=]'s [=url/origin=].
```
domfarolino, line 1888: We shouldn't get the response's origin, but rather the navigation params's origin (https://html.spec.whatwg.org/C#navigation-params-origin), as this will be the actual origin used for the document.

spec.bs
```diff
@@ -1731,3 +1773,212 @@ specification is printed below:
...
+instance=]'s [=fenced frame config instance/effective enabled permissions=] is consulted [=Should
+navigation response to navigation request be blocked by Permissions Policy?|during navigation=]. A
+<{fencedframe}> navigation can only succeed if the [=Document/permissions policy=] for the
+navigation's resulting {{Document}} has an [=permissions policy/inherited policy=] such that the
+[=inherited policy for a feature|inherited policy value=] is "`Enabled`" for each feature in the
+[=fenced frame config/effective enabled permissions=]. Otherwise the environment the <{fencedframe}>
+is embedded in is deemed unsuitable for the [=fenced frame config=], and the navigation is blocked.
+
+At the same time, to make sure that a <{fencedframe}>'s embedder does not directly influence content
+in the <{fencedframe}>'s based on that navigation's [=navigation params/origin=] (since the origin
```
gtanzer, line 1799: extra 's on <{fencedframe}>'s here

spec.bs
```diff
@@ -1731,3 +1773,212 @@ specification is printed below:
   /fenced-frame/cspee.https.html
   /fenced-frame/embedder-csp-not-propagate.https.html
 </wpt>
+
+<h3 id=permissions-policy-changes>Permissions Policies</h3>
+
+*This introductory sub-section is non-normative.*
+
+The [=policy-controlled features=] available to {{Document}}s inside of a <{fencedframe}> are
+determined exclusively by the {{FencedFrameConfig}} that the <{fencedframe}> navigates to.
```
gtanzer, line 1782: This is true right now, but maybe not true with future changes (or default mode, depending on how we spec it).

spec.bs
```diff
@@ -1731,3 +1773,212 @@ specification is printed below:
...
+   if the subframe's {{Document}} was [=same origin=] with its embedder, a check this specification
+   avoids for fenced frames, since the <{fencedframe}>s {{Document}}'s [=Document/origin=] is
+   derived from cross-site data. Therefore, we simply "fail close".
+
+ * If a [=policy-controlled feature|feature=] that [=list/exists=] in the [=fenced frame
+   config/effective enabled permissions=] has a [=default allowlist=] of <a for="default
+   allowlist">`'self'`</a> but either the \`<a http-header>`Permissions-Policy`</a>\` header served
+   on the <{fencedframe}>'s embedder *OR* the <{fencedframe/allow}> attribute sets that
+   [=policy-controlled feature|feature=]'s [=allowlist=] to [=the special value *=], then the
+   navigation inside the <{fencedframe}> will succeed, and the resulting {Document}} is [=allowed to
```
gtanzer, line 1831: `{Document}}` -> `{{Document}}`

spec.bs
domfarolino: Right, thanks. For now I'm just worried about spec'ing opaque stuff basically.

spec.bs
```diff
@@ -1731,3 +1766,235 @@ specification is printed below:
...
+
+  1. [=list/For each=] |effective permission| of |effective permissions|:
+
+    1. If |inherited policy|[|effective permission|] is "Disabled", return "`Blocked`".
+
+  1. Return "`Allowed`."
+</div>
+
+<div algorithm=define-inherited-policy-in-container-patches>
+  Modify the [$Define an inherited policy for feature in container at origin$] algorithm to
```
domfarolino, line 1957: Do we need to also modify this algorithm to "fence" the container policy look-up? Right now https://w3c.github.io/webappsec-permissions-policy/#define-inherited-policy-in-container looks at the container policy (which may contain 'none', `self', 'src`, '*', or an origin) and runs the "matches" algorithm there. I think we'll need to ensure that 'self', 'src', and "an origin" are "fenced" and fail right? Or is there a reason we don't have to do this?

domfarolino (approved):
I added some more fixes and editorial content. I think this is all ready to go % one single comment!

spec.bs
domfarolino: Filed https://github.com/WICG/fenced-frame/issues/82 so we can just land this PR now.

PR #80, opened 2023-04-25
Introduce navigable's config instance

domfarolino:
I think this PR makes things better and more explicit, but I believe things are still broken even after this. A few questions/points:

 - In the `Fence` methods, should it ever be possible for `instance` to be null, as we check for in step 2? I think the `instance` that we check when retrieving the `Fence` object in the first place (https://wicg.github.io/fenced-frame/#dom-window-fence > Step 2) should be a sufficient gate, such that we don't have to check for _instance_ anymore inside the methods, right? The way I envision the ideal world here, is that because we have to support URN iframes, we make the "instance" retriever in both the `window.fence` getter and the `fence.*` methods both traverse upwards until they find a navigable with a non-null _instance_ and both reference that instance. Does that sound right?

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/80.html" title="Last updated on Apr 25, 2023, 1:24 PM UTC (630d476)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/80/ff89560...630d476.html" title="Last updated on Apr 25, 2023, 1:24 PM UTC (630d476)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -706,6 +706,18 @@ A <dfn export>fenced frame config instance</dfn> is a struct with the following 
     :: |config|'s [=fenced frame config/embedder shared storage context=]
 </div>
 
+Each [=navigable=] has a <dfn for=navigable>fenced frame config instance</dfn>, which is a [=fenced
+frame config instance=] or null, initially null.
+
+Advisement: This [=navigable/fenced frame config instance=] should really exist on [=traversable
+navigable=], specifically a [=fenced navigable container/fenced navigable=], however until
+third-party cookies are <a
+href=https://w3ctag.github.io/web-without-3p-cookies/#introduction>deprecated</a>, this
+specification supports many of the <{fencedframe}> constructs on the <{iframe}> element. This
+requires that for the short term, a normal [=navigable container/content navigable=] be able to load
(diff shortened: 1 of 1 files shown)


gtanzer:
@domfarolino 

> The way I envision the ideal world here, is that because we have to support URN iframes, we make the "instance" retriever in both the window.fence getter and the fence.* methods both traverse upwards until they find a navigable with a non-null instance and both reference that instance. Does that sound right?

Yes, that sounds right. There would be some algorithm "Get the current fenced frame config instance" which is implemented as that for now, and later becomes simpler.

domfarolino:
OK, let's implement that in a PR just after this one to fix all of this mess. I'm just trying to land this for https://github.com/WICG/fenced-frame/pull/74.

gtanzer:
> In the Fence methods, should it ever be possible for instance to be null, as we check for in step 2? I think the instance that we check when retrieving the Fence object in the first place (https://wicg.github.io/fenced-frame/#dom-window-fence > Step 2) should be a sufficient gate, such that we don't have to check for instance anymore inside the methods, right?

And yes, this also sounds right.

gtanzer (approved):

PR #84, opened 2023-05-05
Spec config assignment and core #navigate algorithm changes

domfarolino:
This PR finally specs the `config` IDL attribute assignment, making it the only entrypoint into HTML's `#navigate` algorithm. This PR also adds patches to the navigate algorithm, including the invocation to _find a config_ which may wait arbitrarily in the navigation process until the config is finalized and we can override the navigation URL to the actual mapped URL.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/84.html" title="Last updated on May 5, 2023, 10:10 PM UTC (2fdf0cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/84/5af7b1a...2fdf0cf.html" title="Last updated on May 5, 2023, 10:10 PM UTC (2fdf0cf)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1800,10 +1800,6 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
          [=URL=] whose [=url/scheme=] is "`https`".
       1. Set |url| to |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
 
-      1. Let |config instance| be the result of [=instantiate a config|instantiating=] |config|.
-
-      1. Set |navigable|'s [=navigable/fenced frame config instance=] to |config instance|.
-
   <wpt>
     /fenced-frame/frame-navigation.https.html
   </wpt>
--- a/spec.bs
+++ b/spec.bs
@@ -68,6 +68,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: active session history entry; url: nav-active-history-entry
         text: current session history entry; url: nav-current-history-entry
         text: parent; url: nav-parent
+        text: ongoing navigation
       for: traversable navigable
         text: session history entries; url: tn-session-history-entries
       for: browsing context group
@@ -89,6 +90,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: replace; url: hh-replace
       for: document state
         text: document; url: document-state-document
+      text: checking if unloading is user-canceled
     urlPrefix: interaction.html
       text: activation notification; url: activation-notification
       text: consume user activation; url: consume-user-activation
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1767,23 +1760,62 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+         any subsequent embedder-initiated navigations, <span class=allow-2119>should</span> they
+         occur, by the usual mechanism that tracks the [=navigable/ongoing navigation=].
+
+      1. [=Assert=] |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=] is a
+         [=URL=] whose [=url/scheme=] is "`https`".
+      1. Set |url| to |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
+
+      1. Let |config instance| be the result of [=instantiate a config|instantiating=] |config|.
+
+      1. Set |navigable|'s [=navigable/fenced frame config instance=] to |config instance|.
```
gtanzer, line 1805: Does this handle pending/failed navigations properly? The algorithm this is patching seems equivalent to navigation start, and navigation commit is separate. Not sure if each pending embedder-initiated navigation gets its own navigable. (From the changes above, it seems like a fenced navigable is at least something that sticks around and is navigated multiple times.)

domfarolino: > Does this handle pending/failed navigations properly? The algorithm this is patching seems equivalent to navigation start, and navigation commit is separate.

Yeah this is the start of the navigation. After this we'll go into the fetching portion which is the other period of time where we wait for an arbitrarily long period of time. "Commit" is _something like_ after-fetch & after all of the navigation failure checks pass, we finally enter ["loading a document"](https://html.spec.whatwg.org/C#populating-a-session-history-entry:loading-a-document).

And yes a fenced navigable (and normal navigable for that matter) sticks around for as long as the element containing it is connected to the DOM. So my plan was to keep overwriting the navigable's config instance for each embedder-initiated navigation. I guess if the navigation were to fail, the config instance would just be hanging around; I suppose we can consider clearing it but I'm not sure it makes a difference either way.

gtanzer: But while the navigation is pending (or if it fails), what happens to the config instance of the existing context? In the implementation it's in the NavigationRequest, which I assume here is like `navigationParams`.

domfarolino: Eep. Yep I think you're absolutely right. Let me remove the step that assigns the instantiated config for now — I'll add it back in another PR that adjusts things later in the navigation flow.

gtanzer: This is what I meant in one of the spec task doc comments, that maybe we should give each embedder-initiated navigation its own fenced navigable. That seems very easy to spec correctly (be sure that everything is isolated across the navigations, even beyond the fenced frame config instance). But there might be other reasons why we don't want to do that.

gtanzer (approved):

spec.bs
domfarolino: Saw that, sorry I haven't responded yet. The idea is this: browsing context group would be the right primitive to lay things on top of that have to be truly isolated across navigations that swap BCGs. (This matters more with subsequent commits I'll make that force browsing context group _swaps_ within the fenced navigable for embedder-initiated navigations).

See for example https://github.com/whatwg/html/issues/8563 about `name` accidentally getting stored on navigable and persisting across BCG swaps when it probably shouldn't. I'll respond over on Google docs but maybe we should move the config instance to BCG (or browsing context while we have to support URN iframes).

PR #89, opened 2023-05-18
Add "on navigate callback"

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/89.html" title="Last updated on May 22, 2023, 5:38 PM UTC (fdae7dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/89/204e674...fdae7dd.html" title="Last updated on May 22, 2023, 5:38 PM UTC (fdae7dd)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -556,6 +556,8 @@ An <dfn export for=fencedframetype>interest group descriptor</dfn> is a struct w
   :: a [=string=]
 </dl>
 
+An <dfn export for=fencedframetype>on navigate callback</dfn> is a TODO.
+
 An <dfn export for=fencedframetype>exhaustive set of sandbox flags</dfn> is a [=sandboxing flag
 set=].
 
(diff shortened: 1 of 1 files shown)


gtanzer:
@domfarolino This is done modulo the type definition for "on navigate callback", which is a TODO for this PR (not for the future).

domfarolino:
Got it, thanks. I think the type can just be something like "these steps, which take a foo and a bar" (with the steps defined below). Or if we ever need to set this asynchronously we can just make this super vague like https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-load-resumption-steps.

gtanzer:
@domfarolino changed it to just say that the callback is "a series of steps".

domfarolino (approved):

PR #90, opened 2023-05-22
Move fenced frame config instance to browsing context

domfarolino:
This is a partial follow-up to the discussion in https://github.com/WICG/fenced-frame/pull/84#discussion_r1186531028. All this does is move the fenced frame config instance from navigable => browsing context, and updates the advisement text around it and all usages.

A follow-up to this will finalize the config instance assignment to happen at the time where a new browsing context group is chosen for the fenced frame for embedder-initiated navigations, so that the config instance is always wiped.


spec.bs
```diff
@@ -945,17 +945,16 @@ A <dfn export>fenced frame config instance</dfn> is a struct with the following
     :: |config|'s [=fenced frame config/embedder shared storage context=]
 </div>
 
-Each [=navigable=] has a <dfn for=navigable>fenced frame config instance</dfn>, which is a [=fenced
-frame config instance=] or null, initially null.
+Each [=browsing context=] has a <dfn for="browsing context">fenced frame config instance</dfn>,
```
gtanzer, line 948: What is the significance of [=browsing context=] here vs [=browsing context group=] above?

gtanzer (approved):
LGTM % question on BC vs BCG

spec.bs
gtanzer: OK, whatever

PR #100, opened 2023-06-02
Address issues with focus patches from spec review

blu25:
- Reword step 9 to make it clear that the starting point should be a navigable's parent OR a fenced navigable's unfenced parent.
- Reword step 9 to reference the unfenced parent algorithm instead of the traversable navigable's unfenced parent declaration.
- Fix formatting issues.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/100.html" title="Last updated on Jun 5, 2023, 8:04 PM UTC (75e4d0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/100/fb89d8c...75e4d0c.html" title="Last updated on Jun 5, 2023, 8:04 PM UTC (75e4d0c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1753,10 +1753,10 @@ the fenced frame boundary, which is a privacy leak. To avoid this, we effectivel
 
   Modify step 9 of the [=sequential focus navigation=] algorithm to read:
 
-  9. Otherwise, |starting point| is a [=focusable area=] in a [=child navigable=] or [=fenced
-     navigable container/fenced navigable=]. Set |starting point| to that [=child navigable=]'s
-     [=navigable/parent=] or [=fenced navigable container/fenced navigable=]'s [=navigable/unfenced
-     parent=] and return to the step labeled <i>loop</i>.
+  9. Otherwise, |starting point| is a [=focusable area=] whose [=Node/node document=]'s [=node
+     navigable=] is a [=child navigable=] or [=fenced navigable container/fenced navigable=]. Set
+     |starting point| to that [=node navigable=]'s [=navigable/unfenced parent=] and return to the
+     step labeled <i>loop</i>.
 </div>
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1737,25 +1738,25 @@ the fenced frame boundary, which is a privacy leak. To avoid this, we effectivel
...
+     [=focus-unfenced|unfenced=] set to true and return.
 
   Modify step 9 of the [=sequential focus navigation=] algorithm to read:
 
-  9. Otherwise, |starting point| is a [=focusable area=] in a [=child navigable=] or
-    [=fenced navigable container/fenced navigable=]. Set |starting point| to that
-    [=child navigable=] or [=fenced navigable container/fenced navigable=]'s
-    [=traversable navigable/unfenced parent=] and return to the step labeled <i>loop</i>.
+  9. Otherwise, |starting point| is a [=focusable area=] in a [=child navigable=] or [=fenced
+     navigable container/fenced navigable=]. Set |starting point| to that [=child navigable=]'s
```
domfarolino, line 1757: This first sentence is ambiguous, so let me try and peel it apart to understand. What if we're dealing with a child navigable _inside_ a fenced navigable? Like an iframe inside of a fenced frame. Should we be returning the iframe's parent navigable (i.e., the fenced frame root), or should we be going all the way up past the fenced frame to its embedder, and returning _that_ embedder navigable?

If it is the former, then I think my comment below applies.

Either way this sentence is ambiguous and we should clear it up by saying something like:

> Otherwise, |starting point| is a [=focusable area=] whose [=node/node document=]'s [=document/navigable=] is a [=child navigable=] or a [=fenced navigable container/fenced navigable=]

(I'm sure I don't have all the dfn references in that down perfectly, but you get the gist of it)

PR #101, opened 2023-06-05
Spec config instance setting and clearing semantics

domfarolino:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/101.html" title="Last updated on Jun 6, 2023, 12:14 PM UTC (81efe41)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/101/fb89d8c...81efe41.html" title="Last updated on Jun 6, 2023, 12:14 PM UTC (81efe41)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1337,11 +1337,18 @@ Each {{Window}} object has an associated <dfn for=Window>fence</dfn>, which is a
 
     * If |creator| is non-null and <var ignore>embedder</var> is not a <{fencedframe}> element,
       then:
-</div>
 
-Note: This is because we need to ensure that we do not leak <var ignore>creator</var>'s [=the
-document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
-container=], across the fenced frame boundary.
+  Note: This is because we need to ensure that we do not leak <var ignore>creator</var>'s [=the
+  document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
+  container=], across the fenced frame boundary.
+
+
+  Add a substep to step 4 of this algorithm (that was modified above) that reads:
+
+    4. Set <var ignore>browsingContext</var>'s [=browsing context/fenced frame config instance=] to
+       |creator|'s [=Document/browsing context=]'s [=browsing context/fenced frame config
+       instance=].
+</div>
 
 <h3 id=nested-traversables>Nested traversables</h3>
 
@@ -1792,8 +1799,6 @@ the fenced frame boundary, which is a privacy leak. To avoid this, we effectivel
 This section describes how the <{fencedframe}> element interacts with the ever-complicated process
 of navigation, which includes integration with various headers, isolation mechanisms, and policies.
 
-<span class=XXX>TODO: Describe how the [=fenced frame config instance=] is carried along the navigation request and is stored upon navigation commit, along with running the [=fenced frame config instance/on navigate callback=].</span>
-
 <h4 id=supports-loading-mode>The \`<a http-header><code>Supports-Loading-Mode</code></a>\` HTTP response header</h4>
 
 <i>This section is intended to monkeypatch the [[HTML]] Standard, however since the \`<a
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1989,6 +1992,71 @@ finalized.
...
+         [=Document/origin=];
+
+       * |navigationParams|'s [=navigation params/response=]'s
+         [=response/has-cross-origin-redirects=] boolean is true
+
+     Then set |navigationParams|'s [=navigable/active browsing context=]'s [=browsing context/fenced
+     frame config instance=] to null.
+
+     Note: This ensures that the {{Window/fence}} API return null for script running in the
+     {{Document}} created by this algorithm.
```
domfarolino, line 2057: @gtanzer I can't remember if this is the goal (to completely clear `window.fence`) or if we just want to clear *parts* of the config, so please let me know if this does not match the implementation.

Also if you could point me towards any WPTs demonstrating this behavior it would be great to link to them from here.

spec.bs
```diff
@@ -1989,6 +1997,80 @@ finalized.
...
+   persist through this kind of cross-{{Document}} same-[=browsing context group=] navigation, so
+   we manually clear it.
+
+ * <i>Same-origin inner-content-initiated navigations</i>: The [=fenced frame config instance=]
+   referenced in the immediately-preceding point persists through this kind of cross-{{Document}}
+   navigation. The patches below take no action to clear it, so it continues to be visible through
+   the {{Window/fence}} API.
+
+<div algorithm=create-and-initialize-document-patch>
+  Modify [[!HTML]]'s [=create and initialize a Document object=] algorithm to insert a step after
```
gtanzer, line 2033: "a step" -> "two steps"?

spec.bs
```diff
@@ -1989,6 +1997,80 @@ finalized.
...
+
+    1. Set |browsingContext|'s [=browsing context/fenced frame config instance=] to
+       |navigationParams|'s [=navigation params/fenced frame config instance=].
+
+  4. Otherwise, if |navigationParams|'s [=navigable/active browsing context=]'s [=browsing
+     context/fenced frame config instance=] is non-null, and either of the following conditions are
+     true:
+
+       * |navigationParams|'s [=navigation params/origin=] is not [=same origin=] with
+         |navigationParams|'s [=navigation params/navigable=]'s [=navigable/active document=]'s
```
gtanzer, line 2053: This should be whether it's same-origin with the config instance's mapped url, not the active document.

spec.bs
gtanzer: Liam recently changed it to clear parts of the config instance, but that was just motivated by the way permissions work in impl across processes. I don't think the spec needs to work like that.

The config instance semantics are:
* Embedder-initiated navigations unconditionally set the config instance for the fenced frame tree
* Future accesses to the config instance through window.fence APIs, etc. are gated by a check on whether the accessing origin is same-origin with the config instance's mapped url.

I'm not sure whether the latter bullet really should be spec'd as nulling the config instance. We still want the information associated with it.

The spec for reportEvent (https://wicg.github.io/fenced-frame/#dom-fence-reportevent) and other APIs already include this origin check.

spec.bs
```diff
@@ -1334,11 +1337,18 @@ Each {{Window}} object has an associated <dfn for=Window>fence</dfn>, which is a
...
-container=], across the fenced frame boundary.
+  Note: This is because we need to ensure that we do not leak <var ignore>creator</var>'s [=the
+  document's referrer|referrer=], [=Document/origin=], [=creator base url=], [=Document/policy
+  container=], across the fenced frame boundary.
+
+
+  Add a substep to step 4 of this algorithm (that was modified above) that reads:
+
+    4. Set <var ignore>browsingContext</var>'s [=browsing context/fenced frame config instance=] to
+       |creator|'s [=Document/browsing context=]'s [=browsing context/fenced frame config
```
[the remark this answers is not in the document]
domfarolino: Browsing contexts stay put while the many documents in them come and go through navigations and traversals. This is true until there is a browsing context group swap which only happens for (1) COOP/COEP for top-level traversables, and (2) embedder-initiated fenced frame navigations

spec.bs
domfarolino: > We still want the information associated with it.

Can you elaborate? I'm not sure what "information" is here, but it sounds like for cross-origin-to-config-instance iframes within fenced frames (or even fenced frames that have navigated cross-origin to their mapped URL at some point), the `window.fence` API should still be non-null and the APIs on it should still be exposed, its just that they should each decide whether to work or not based on the current origin. Is that right?

spec.bs
domfarolino: Hmm it sounds like the second half of this whole section should probably just be deleted given https://github.com/WICG/fenced-frame/pull/101#discussion_r1218282328. Do you agree?

spec.bs
gtanzer: Right, each API should decide whether or not to work based on the current origin. The fact that cross-origin frames "don't have" a config instance in the renderer implementation is just so compromised renderers can't see that information.

At a spec level, you just have a config instance that governs the entire fenced frame tree and ad hoc checks for each API that uses the data in it.

gtanzer (approved):

PR #106, opened 2023-06-24
Spec the beacon fetch integration

domfarolino:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/106.html" title="Last updated on Jun 24, 2023, 7:11 PM UTC (176cbb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/106/376d4d0...176cbb5.html" title="Last updated on Jun 24, 2023, 7:11 PM UTC (176cbb5)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -746,7 +746,7 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
 
   1. Let |destination url| be |destination map|[|eventType|].
 
-  1. Let |request| be a new [=request=] with the following:
+  1. Let |request| be a new [=request=] with the following properties:
 
     : [=request/method=]
     :: <code>`POST`</code>
--- a/spec.bs
+++ b/spec.bs
@@ -746,9 +746,51 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
 
   1. Let |destination url| be |destination map|[|eventType|].
 
-  1. Send a beacon with |eventData| to |destination url|.
+  1. Let |request| be a new [=request=] with the following:
 
-     Issue: Specify the [[FETCH]] integration for sending this request.
+    : [=request/method=]
+    :: <code>`POST`</code>
+
+    : [=request/URL=]
+    :: |destination url|
+
+    : [=request/header list=]
+    :: A new [=header list=] containing a [=header=] whose [=header/name=] is `"Content-Type"` and
+       [=header/value=] is `"text/plain"`.
+
+    : [=request/body=]
+    :: A [=body=] whose [=body/source=] is |eventData|.
+
+    : [=request/client=]
+    :: null
+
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -746,9 +746,51 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
...
+    : [=request/service-workers mode=]
+    :: `"all"`
+
+       Issue: The default is `"all"`, so we technically don't have to set anything here. We do in order
+       to remind ourselves that it might be more appropriate to skip service workers here, like some
+       [other
+       beacons](https://wicg.github.io/attribution-reporting-api/#ref-for-request-service-workers-mode).
+
+    : [=request/origin=]
+    :: <span class=XXX>Get the origin from somewhere, like the [implementation
```
domfarolino, line 776: This is the only thing that is blocking this PR I think. From [this](https://wicg.github.io/fenced-frame/#ref-for-send-a-beacon%E2%91%A0) entrypoint I can get an origin pretty easily and plumb it through. But I don't know enough about what's going on with https://wicg.github.io/fenced-frame/#finalize-a-reporting-destination or how to get an origin from there. Thoughts? Do we need to store it somewhere in the destination map or something?

spec.bs
```diff
@@ -746,9 +746,51 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
 
   1. Let |destination url| be |destination map|[|eventType|].
 
-  1. Send a beacon with |eventData| to |destination url|.
+  1. Let |request| be a new [=request=] with the following:
```
gtanzer, line 749: Is there supposed to be a noun after "following"?

gtanzer (approved):
LGTM, though I don't really know the details of how the request should be configured.

spec.bs
gtanzer: Ah, thought this would be in a followup.

I assume it needs to be stored in the pending event, like in the impl https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/fenced_frame/fenced_frame_reporter.cc;l=288;drc=fcf80d7b62173cb0f418679551f0dd8909cb79ab;bpv=1;bpt=1

gtanzer: https://wicg.github.io/fenced-frame/#fencedframetype-pending-event

so just another field here, and stored wherever the rest is currently stored

spec.bs
domfarolino: Meh, idk. Requests aren't formal structs, so I don't want to say "items". I guess I could put "properties".

spec.bs
domfarolino: OK we can probably tackle that separately then. I'll file a bug for this outside of this PR and we can unblock this one.

PR #111, opened 2023-08-18
Make FencedFrameConfig IDL objects serializable

blu25:
This PR:

- Makes `FencedFrameConfig` objects serializable and defines serializing/deserializing steps for it.
- Removes `config` as an attribute of a `FencedFrameConfig`, replacing it with a clone of the values it needs from the internal `Fenced Frame Config` struct.
- Passes in the `sharedStorageContext` directly to the `navigate` algorithm, which will read the `sharedStorageContext` from the IDL object to the `embedder shared storage context` in the `Fenced Frame Config` struct. This is done since we are no longer writing directly to the `Fenced Frame Config` struct in the `setSharedStorageContext()` algorithm.
- Modifies `getNestedConfigs()` to account for the changes to `FencedFrameConfig`.
- Update the step number for the `navigate` step that goes in parallel.
- Update the permissions policy introduction section to mention the internal `[=fenced frame config=]` struct instead of the `{{FencedFrameConfig}}` IDL object.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/111.html" title="Last updated on Aug 24, 2023, 5:33 PM UTC (4cfb07d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/111/7281202...4cfb07d.html" title="Last updated on Aug 24, 2023, 5:33 PM UTC (4cfb07d)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -456,9 +456,13 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
 
   1. Let |urn uuid| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/urn=].
 
+  1. Let |shared storage context| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/
+     sharedStorageContext=].
+
   1. [=Navigate=] |element|'s [=fenced navigable container/fenced navigable=] to |urn uuid| using
      |element|'s [=Node/node document=], with [=historyHandling=] set to "<a for="history handling
-     behavior">`replace`</a>" , and [=referrerPolicy=] set to <a>"`no-referrer`"</a>.
+     behavior">`replace`</a>", [=referrerPolicy=] set to <a>"`no-referrer`"</a>, and
+     sharedStorageContext set to |shared storage context|.
 
      Note: See [[#navigation-changes]] for the <{fencedframe}>-specific changes to the ordinary
      navigation flow.
@@ -524,7 +528,7 @@ to discern when a pending config is finalized, it is important that all visibili
 transparent fields do not change from the pending config to the finalized config, given that they
 can be inspected through {{FencedFrameConfig}}'s getters. Therefore, a {{FencedFrameConfig}} that
 is created and exposed to the web platform is effectively immutable even if its underlying
-[=fencedframeconfig/config=] is technically "pending", and will finish resolving completely later.
+[=fenced frame config=] is technically "pending", and will finish resolving completely later.
 
 Each [=fenced frame config mapping=] has a <dfn for="fenced frame config mapping">maximum number of
 configs</dfn>, which is implementation-defined. The [=fenced frame config mapping/maximum number of
@@ -893,7 +897,7 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
     :: a [=URL=]
 
     : <dfn for="mapped url">visibility</dfn>
-    :: a [=visibility=]
+    :: a [=fencedframeconfig/visibility=]
 
   : <dfn>container size</dfn>
   :: null, or a [=fencedframetype/size=]
@@ -904,7 +908,7 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
     :: a [=fencedframetype/size=]
 
     : <dfn for="content size">visibility</dfn>
-    :: a [=visibility=]
+    :: a [=fencedframeconfig/visibility=]
 
   : <dfn>interest group descriptor</dfn>
   :: null, or a [=struct=] with the following [=struct/items=]:
@@ -912,7 +916,7 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
     :: an [=fencedframetype/interest group descriptor=]
 
     : <dfn for="interest group descriptor">visibility</dfn>
-    :: a [=visibility=]
+    :: a [=fencedframeconfig/visibility=]
 
   : <dfn>on navigate callback</dfn>
   :: null, or a series of steps
@@ -923,7 +927,7 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
     :: an [=fencedframetype/exhaustive set of sandbox flags=]
 
     : <dfn for="effective sandbox flags">visibility</dfn>
-    :: a [=visibility=]
+    :: a [=fencedframeconfig/visibility=]
 
   : <dfn>effective enabled permissions</dfn>
   :: null, or a [=struct=] with the following [=struct/items=]:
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -524,7 +528,7 @@ to discern when a pending config is finalized, it is important that all visibili
 transparent fields do not change from the pending config to the finalized config, given that they
 can be inspected through {{FencedFrameConfig}}'s getters. Therefore, a {{FencedFrameConfig}} that
 is created and exposed to the web platform is effectively immutable even if its underlying
-[=fencedframeconfig/config=] is technically "pending", and will finish resolving completely later.
+[=fenced frame config=] is technically "pending", and will finish resolving completely later.
```
domfarolino, line 531: Hmm, well now a fenced frame element doesn't directly have a fenced frame config struct, so maybe we should make this more specific by referencing `the [=fenced frame config=] represented by the [=fencedframe/config=]'s [=fencedframeconfig/urn=]`. WDYT?

spec.bs
```diff
@@ -2097,6 +2144,9 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 </div>
 
 <div algorithm=navigate>
+  Modify the beginning sentence of [[HTML]]'s [=navigate=] algorithm to include an extra parameter:
```
domfarolino, line 2147: Instead of "beginning sentence", maybe "definition"

spec.bs
```diff
@@ -1314,8 +1341,28 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        : [=fencedframeconfig/containerWidth=]
+        :: null if |config|'s [=fenced frame config/container size=] is null, otherwise |config|'s
+           [=fenced frame config/container size=]'s [=size/width=]
+
+        : [=fencedframeconfig/containerHeight=]
+        :: null if |config|'s [=fenced frame config/container size=] is null, otherwise |config|'s
+           [=fenced frame config/container size=]'s [=size/height=]
+
+        : [=fencedframeconfig/contentWidth=]
+        :: null if |config|'s [=fenced frame config/content size=] is null, {{OpaqueProperty}} if
```
domfarolino, line 1356: We cant just use `{{OpaqueProperty}}`. See the other usages of this where we have to specify a value (which happens to be the only single value available).

spec.bs
```diff
@@ -1314,8 +1341,28 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+           |config|'s [=fenced frame config/content size=]'s [=content size/visibility=] is
+           [=visibility/opaque=], otherwise |config|'s [=fenced frame config/content size=]'s
+           [=size/width=]
+
+        : [=fencedframeconfig/contentHeight=]
+        :: null if |config|'s [=fenced frame config/content size=] is null, {{OpaqueProperty}} if
```
domfarolino, line 1362: same

spec.bs
```diff
@@ -1141,60 +1145,83 @@ element with arbitrary URLs *not* produced by config-generating APIs for develop
...
 </div>
 
-<div algorithm>
-  The <dfn method for=FencedFrameConfig>setSharedStorageContext(|contextString|)</dfn> method steps
-  are:
+<div algorithm="FencedFrameConfig deserializer">
+  Their
+  [=deserialization steps=], given |serialized|, |value|, and <var ignore> targetRealm</var> are:
+
+  1. Initialize |value|'s [=fencedframeconfig/urn=] attribute to |serialized|.`[[Urn]]`.
```
domfarolino, line 1209: I don't think there should be wrapped in backticks. Just normal text (escaping the brackets if needed?).

spec.bs
```diff
@@ -1141,60 +1145,83 @@ element with arbitrary URLs *not* produced by config-generating APIs for develop
...
+<div algorithm="FencedFrameConfig serializer">
+  {{FencedFrameConfig}} objects are [=serializable objects=]. Their [=serialization steps=], given
+  |value|, |serialized|, and |forStorage| are:
 
-  1. If [=this=]'s [=fencedframeconfig/config=]'s [=fenced frame config/content size=]'s [=content
-     size/visibility=] is "<a for=visibility>`transparent`</a>", return the [=fenced frame
-     config/content size=]'s [=content size/value=]'s [=size/width=].
+  1. If |forStorage| is true, then throw a {{DataCloneError}} {{DOMException}}.
+  
+  1. Set |serialized|.`[[Urn]]` to the value of |value|'s [=fencedframeconfig/urn=] attribute.
```
domfarolino, line 1185: What you're accessing here with `[=fencedframeconfig/urn=]` isn't an IDL attribute, it is just a normal member. So you don't have to add `to the value of |value|'s foo attribute`. You can just have `to |value|'s [=fencedframeconfig/foo=]`. Here and elsewhere.

spec.bs
blu25: That works for me.

spec.bs
blu25: They need to be escaped, but I can get away with just escaping the first bracket.

spec.bs
```diff
@@ -456,9 +456,13 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
...
   1. Let |urn uuid| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/urn=].
 
+  1. Let |shared storage context| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/
+     sharedStorageContext=].
+
   1. [=Navigate=] |element|'s [=fenced navigable container/fenced navigable=] to |urn uuid| using
      |element|'s [=Node/node document=], with [=historyHandling=] set to "<a for="history handling
-     behavior">`replace`</a>" , and [=referrerPolicy=] set to <a>"`no-referrer`"</a>.
+     behavior">`replace`</a>", [=referrerPolicy=] set to <a>"`no-referrer`"</a>, and
+     sharedStorageContext set to |shared storage context|.
```
domfarolino, line 465: Since your modification to #navigate added a normal positional optional parameter, you don't need to reference its name. Generally we only add linkable/named parameters to this algorithm (and reference those names in the calling of #navigate) if the argument is directly used by another algorithm: https://docs.google.com/document/d/1yxnzjRDVmAR5CC9GcAyY448lBD0u0E98eUEMHDhx1Dw/edit?disco=AAAAeXYly54.

It's hard to say how much we're sticking with that these days, but maybe we are? In any case, I'd recommend either dropping the specific name here (since it is linked), or make the parameter name that you added _linkable_, and then link to it here.

domfarolino (commented):
Pretty much lgtm % one nit!

domfarolino (approved):

PR #113, opened 2023-08-23
Spec reportEvent to custom destination URLs with macro substitution

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/113.html" title="Last updated on Sep 26, 2023, 4:20 PM UTC (29bc781)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/113/3b93e20...29bc781.html" title="Last updated on Sep 26, 2023, 4:20 PM UTC (29bc781)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -688,7 +688,7 @@ following [=struct/items=]:
 
 <dl export dfn-for="reporting destination info">
   : <dfn>reporting url map</dfn>
-  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=urls=]
+  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=URLs=]
 
   : <dfn>reporting macro map</dfn>
   :: null, or a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=strings=]
--- a/spec.bs
+++ b/spec.bs
@@ -812,7 +812,8 @@ A <dfn export for=fencedframetype>destination event</dfn> is either a
   1. Let |request| be a new [=request=] with the following properties:
 
     : [=request/method=]
-    :: <code>`POST`</code> if |event| is a [=fencedframetype/destination enum event=], otherwise <code>`GET`</code>.
+    :: <code>`POST`</code> if |event| is a [=fencedframetype/destination enum event=], otherwise
+       <code>`GET`</code>.
 
     : [=request/URL=]
     :: |destination url|
--- a/spec.bs
+++ b/spec.bs
@@ -683,11 +683,22 @@ A <dfn export for=fencedframetype>pending event</dfn> is a [=struct=] with the f
   :: a [=fencedframetype/destination event=]
 </dl>
 
+A <dfn export for=fencedframetype>reporting destination info</dfn> is a [=struct=] with the
+following [=struct/items=]:
+
+<dl export dfn-for="reporting destination info">
+  : <dfn>reporting url map</dfn>
+  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=urls=]
+
+  : <dfn>reporting macro map</dfn>
+  :: null, or a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=strings=]
+</dl>
+
 A <dfn export for=fencedframetype>fenced frame reporting map</dfn> is a [=map=] whose [=map/keys=]
 are {{FenceReportingDestination}}s and whose [=map/values=] are either:
  * [=lists=] of [=fencedframetype/pending events=] (which are used to represent events that need to
    be reported asynchronously, because the metadata has not been finalized yet); or
- * [=maps=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=urls=] (which are used
+ * [=fencedframetype/reporting destination infos=] (which are used
    to represent the actual metadata once it is finalized).
 
 Note: This representation is meant to allow config-generating APIs to reduce latency by resolving
@@ -704,21 +715,26 @@ be sent.
 
 <div algorithm>
   In order to <dfn export>finalize a reporting destination</dfn>, given a [=fencedframetype/fenced
-  frame reporting map=] |reporting map|, a {{FenceReportingDestination}} |destination|, and a
-  [=map=] |destination map| whose [=map/keys=] are [=strings=] and whose [=map/values=] are
-  [=urls=], run these steps:
+  frame reporting map=] |reporting map|, a {{FenceReportingDestination}} |destination|, a [=map=]
+  |destination map| whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=urls=], and
+  |macro map|, which is either null or a [=map=] whose [=map/keys=] are [=strings=] and whose
+  [=map/values=] are [=strings=], run these steps:
 
   1. [=Assert=] that |reporting map|[|destination|] is a [=list=] (i.e., that |destination|'s
      metadata has not yet been finalized).
 
   1. Let |pending event list| be |reporting map|[|destination|].
 
-  1. [=map/Set=] |reporting map|[|destination|] to |destination map|.
+  1. [=map/Set=] |reporting map|[|destination|] to a [=struct=] with the following [=struct/items=]:
+       : [=reporting destination info/reporting url map=]
+       :: |destination map|
+
+       : [=reporting destination info/reporting macro map=]
+       :: |macro map|
 
   1. [=list/For each=] |pending event| of |pending event list|:
 
-     1. [=Send a beacon=] with |destination map|, |pending event|'s [=pending event/eventType=],
-        and |pending event|'s [=pending event/eventData=].
+     1. [=Send a beacon=] with |destination map| and |pending event|'s [=pending event/event=].
 </div>
 
 A <dfn export for=fencedframetype>fenced frame reporting metadata</dfn> is a [=struct=] with the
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -750,19 +769,45 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
...
 
-  1. Let |destination url| be |destination map|[|eventType|].
+     1. Set |destination url| to |destination map|[|eventType|].
+
+  1. Otherwise:
+     1. Assert that |event| is a [=fencedframetype/destination URL event=].
+     1. Let |macro map| be |destination info|'s [=reporting destination info/reporting macro map=].
+     1. If |macro map| is null, return.
+     1. Set |destination url| to |event|.
+     1. TODO: Substitute macros from |macro map| into |destination url|.
```
gtanzer, line 804: @domfarolino Do you know of any existing spec for functionality like this? (take a map of string -> string, and substitute the keys -> values in another string without recursive substitution)

spec.bs
```diff
@@ -750,19 +769,45 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
...
-  |eventData|, run these steps:
+  In order to <dfn>send a beacon</dfn> with a [=fencedframetype/reporting destination info=]
+  |destination info| and a [=fencedframetype/destination event=] |event|, run these steps:
 
-  1. If |destination map|[|eventType|] does not [=map/exist=], return.
+  1. Let |destination url| be an empty [=string=].
+  1. If |event| is a [=fencedframetype/destination enum event=]:
+     1. Let |destination map| be |destination info|'s
+        [=reporting destination info/reporting url map=].
+     1. |destination map|[|eventType|] does not [=map/exist=], return.
```
qingxinwu, line 795: If ...

spec.bs
```diff
@@ -808,20 +854,32 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
...
-        `"seller"`.
+     1. If |metadata|'s [=fenced frame reporting metadata/direct seller is seller=] is true, set
+        |destination| to `"seller"`.
 
      1. Otherwise, set |destination| to `"component-seller"`.
 
-  1. Let |reporting map| be a reference to |reporter|'s [=fenced frame reporter/fenced frame
-     reporting metadata reference=]'s [=fenced frame reporting metadata/fenced frame reporting
-     map=].
+  1. If |event| is a [=fencedframetype/destination enum event=]:
```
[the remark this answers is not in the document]
gtanzer: Oops, thanks for catching

spec.bs
```diff
@@ -727,6 +740,12 @@ following [=struct/items=]:
 
   : <dfn>direct seller is seller</dfn>
   :: a boolean
```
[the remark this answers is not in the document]
gtanzer: might as well

spec.bs
```diff
@@ -1244,20 +1307,51 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+           {{FenceEvent/eventData}}:
+           1. [=exception/Throw=] a {{TypeError}}.
+        1. Let |destinationURL| be the result of running the [=URL parser=] on
+           {{FenceEvent/destinationURL}}.
+        1. [=exception/Throw=] a {{TypeError}} if any of the following conditions hold:
+           * |destinationURL| is failure;
+           * |destinationURL| [=url/scheme=] is not "`https`";
+        1. Run [=report an event=] using |instance|'s [=fenced frame config instance/fenced frame
+           reporter=] with {{FenceReportingDestination/buyer}} and a
+           [=fencedframetype/destination URL event=] that is event|'s
```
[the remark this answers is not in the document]
gtanzer: surprised it didn't give me a compile error

spec.bs
gtanzer: Fixed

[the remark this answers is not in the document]
gtanzer: Fixed.

gtanzer:
@qingxinwu Thanks for reviewing! Fixed all the comments.

spec.bs
domfarolino: Hmm, the only kind of substitution like this that comes to mind is when we substitute format specifiers in e.g., `console.log('My number is: %f', .435)`, which is in https://console.spec.whatwg.org/#formatter. That's a pretty manual process, and I don't know of any preexisting reusable utility for this unfortunately, so you'll probably just have to do it that way.

spec.bs
```diff
@@ -726,7 +739,13 @@ following [=struct/items=]:
   :: a [=fencedframetype/fenced frame reporting map=]
 
   : <dfn>direct seller is seller</dfn>
-  :: a boolean
+  :: a [=boolean=]
```
domfarolino, line 742: does this need an initial value, like the boolean below?

spec.bs
```diff
@@ -1211,18 +1267,26 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   };
 
   dictionary FenceEvent {
-    required DOMString eventType;
-    required DOMString eventData;
-    required sequence&lt;FenceReportingDestination&gt; destination;
+    // This dictionary has two mutually exclusive modes:
+    //
+    // When reporting to a preregistered destination (specified by enum), the following properties
```
domfarolino, line 1272: Can you add tighter line breaks here to accommodate the narrow IDL viewing window in the spec preview? (Same with below).

spec.bs
```diff
@@ -1211,18 +1267,26 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   };
 
   dictionary FenceEvent {
-    required DOMString eventType;
-    required DOMString eventData;
-    required sequence&lt;FenceReportingDestination&gt; destination;
+    // This dictionary has two mutually exclusive modes:
```
domfarolino, line 1270: I can't remember, is there a technical reason why we can't express this as a [union type](https://webidl.spec.whatwg.org/#dfn-union-type), and instead have to express this as a dictionary with these two implicit modes? If there is we should document it here, and if there is not we should try and follow-up with a web-compatible change to make this more idiomatic and less error-prone.

gtanzer: It was due to distinguishability. The issue isn't as clear in the spec because reportEvent for component ads hasn't been spec'd yet, which makes some of the required fields optional.

spec.bs
gtanzer: I think it's always specified, but might as well give it a default.

spec.bs
gtanzer: Factored it out into a TODO. The exact semantics of this are not super important right now.

gtanzer:
To be merged after https://github.com/WICG/fenced-frame/pull/117 , which creates [=fencedframeutil/substitute macros=].

domfarolino:
> To be merged after https://github.com/WICG/fenced-frame/pull/117 , which creates [=fencedframeutil/substitute macros=].

Sounds good, I think this mostly looks good so let's wait until that gets landed then.

domfarolino (approved):
@qingxinwu Does this look good to you?

spec.bs
```diff
@@ -732,7 +745,13 @@ following [=struct/items=]:
   :: a [=fencedframetype/fenced frame reporting map=]
 
   : <dfn>direct seller is seller</dfn>
-  :: a boolean
+  :: a [=boolean=], initially true
+
+  : <dfn>allowed reporting origins</dfn>
+  :: null or a [=list=] of [=origins=]
```
qingxinwu, line 751: nit: it would be good to briefly explain what this field means, and what null means for the field.

spec.bs
```diff
@@ -820,20 +877,32 @@ A <dfn export for=fencedframetype>fenced frame reporter</dfn> is a [=struct=] wi
...
+     1. If |metadata|'s [=fenced frame reporting metadata/direct seller is seller=] is true, set
+        |destination| to `"seller"`.
 
      1. Otherwise, set |destination| to `"component-seller"`.
 
-  1. Let |reporting map| be a reference to |reporter|'s [=fenced frame reporter/fenced frame
-     reporting metadata reference=]'s [=fenced frame reporting metadata/fenced frame reporting
-     map=].
+  1. If |event| is a [=fencedframetype/destination URL event=]:
+     1. If |event|'s [=url/origin=] is not [=same origin=] with any of the entries in
```
qingxinwu, line 894: IIRC, the implementation also checks if the event's origin is enrolled. Are we assuming that this same-origin check with `allowed reporting origins` covers that? If so, would be good to document at [line 750](https://github.com/WICG/fenced-frame/pull/113/files#diff-6f5a1d8263b0b0c42e2716ba5750e3652e359532647ac934c1c70086ae3ceddaR750) that all origins in [=fenced frame reporting metadata/allowed reporting origins=] are enrolled.

spec.bs
```diff
@@ -1350,20 +1425,57 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+           [=fencedframetype/destination URL event=] that is |event|'s
+           {{FenceEvent/destinationURL}}.
+
+     1. Otherwise:
+        1. If |event| does not have a {{FenceEvent/destination}} or |event| does not have a
+           {{FenceEvent/eventType}}:
+
+           1. [=exception/Throw=] a {{TypeError}}.
+
+        1. [=list/for each=] |destination| of |event|'s {{FenceEvent/destination}}:
```
qingxinwu, line 1455: nit: For each.  s/f/F

spec.bs
```diff
@@ -732,7 +745,13 @@ following [=struct/items=]:
   :: a [=fencedframetype/fenced frame reporting map=]
 
   : <dfn>direct seller is seller</dfn>
```
qingxinwu, line 747: I don't find where this field is set. It seems it always uses the default `true`. Is that expected?

spec.bs
qingxinwu: or maybe just say "passed from protected audience interest group's allowed reporting origins", with link to https://github.com/WICG/turtledove/blob/main/spec.bs#L3201.

spec.bs
gtanzer: It's set here:

https://wicg.github.io/turtledove/#:~:text=Set%20pendingConfig%E2%80%99s%20fenced%20frame%20reporting%20metadata%20to%20a%20struct%20with%20the%20following%20items%3A

spec.bs
gtanzer: Attestation checks are performed when sending the beacon, i.e. it uses an already present check on line 823.

spec.bs
qingxinwu: oh I see. Missed that, because clicking on the field cannot show usages like this. Should have done a search.
feel free to close this comment then.

spec.bs
qingxinwu: Ack. Forget about the "interesting" `optionally` part.

qingxinwu:
LGTM.

spec.bs
gtanzer: Added "An origin must be present in this list to be the destination of a [=fencedframetype/destination URL event=] report."

PR #115, opened 2023-08-24
Add is ad component field to fenced frame config.

xiaochen-z:
This is the first step to spec the ad component reportEvent change.

***
<a href="https://pr-preview.s3.amazonaws.com/xiaochen-z/fenced-frame/pull/115.html" title="Last updated on Nov 10, 2023, 7:07 PM UTC (8115f55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/115/15f4ac2...xiaochen-z:8115f55.html" title="Last updated on Nov 10, 2023, 7:07 PM UTC (8115f55)">Diff</a>


domfarolino:
How will this be used? I'm just trying to get a little context to review this, by understanding how it'll need to look in the future?

> the ad component reportEvent change.

... is a little vague.

xiaochen-z:
> How will this be used? I'm just trying to get a little context to review this, by understanding how it'll need to look in the future?
> 
> > the ad component reportEvent change.
> 
> ... is a little vague.

For ad component, event reporting works different than the main ad. This boolean field allows related event reporting algorithms to know whether this ad is an ad component or not. Please see [the explainer for ad component event reporting](https://github.com/WICG/turtledove/blob/main/Fenced_Frames_Ads_Reporting.md#support-for-ad-components).

domfarolino:
OK, could you include some of that context directly in the PR as a non-normative note, because that context is very useful? This seems to be the first time "component ads" are really referenced in the spec, so at the very least linking to that explainer content from the PR would be wonderful.

xiaochen-z:
Thanks. I've added a note on ad components and the special handling on event reporting.

domfarolino (approved):

gtanzer:
I'm not sure that this field is necessary in spec. I think it's used in the implementation only because the renderer can't access the entire frame tree with ancestors' fenced frame properties. But I might be misremembering.

domfarolino:
> I'm not sure that this field is necessary in spec. I think it's used in the implementation only because the renderer can't access the entire frame tree with ancestors' fenced frame properties. But I might be misremembering.

Should we abandon this or...? I'll trust people more familiar with the implementation these days to make the call.

xiaochen-z:
This field is used at [browser](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/render_frame_host_impl.cc;l=8794?q=is_ad_component&ss=chromium%2Fchromium%2Fsrc) to identify an ad component frame. If we do not mention this field in the spec, is it sufficient to provide a definition of ad component and then describe the algorithm like "if this is an ad component, then ..."?

gtanzer:
Actually maybe "is ad component" is necessary because you can (maybe not right now, but at some point) created nested fenced frames with different constructors, and you want to treat the actual ad component frames differently...

domfarolino:
So shall we land this @gtanzer?

gtanzer:
Sgtm, though maybe "is ad component" should have a default value of false.

xiaochen-z:
Added default values.

domfarolino:
Still look good @gtanzer ?

spec.bs
```diff
@@ -1026,6 +1036,9 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
 
   : <dfn>embedder shared storage context</dfn>
   :: null, or a [=string=]
+  
+  : <dfn>is ad component</dfn>
+  :: A [=boolean=]. Defaulting to false.
```
domfarolino, line 1041:
```suggestion
  :: A [=boolean=], initially false.
```

domfarolino (commented):
Ping @gtanzer

gtanzer:
still lgtm

PR #117, opened 2023-09-06
Spec deprecatedReplaceInURN

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/117.html" title="Last updated on Sep 22, 2023, 2:37 PM UTC (6600ce4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/117/cf06e42...6600ce4.html" title="Last updated on Sep 22, 2023, 2:37 PM UTC (6600ce4)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1222,6 +1222,64 @@ Each {{FencedFrameConfig}} has:
   1. Initialize |value|'s [=fencedframeconfig/contentHeight=] to |serialized|.\[[ContentHeight]].
 </div>
 
+To help with ease of adoption,
+<a href=https://github.com/WICG/turtledove/issues/286#issuecomment-1682842636>until 2026</a> we
+will support the API {{Window/navigator}}.{{Navigator/deprecatedReplaceInURN()}}, which allows you
+to substitute macros into the [=fenced frame config/mapped url=] corresponding to a given
+[=urn uuid=] or {{FencedFrameConfig}}.
+
+<pre class=idl>
+typedef (USVString or FencedFrameConfig) UrnOrConfig;
+
+partial interface Navigator {
+  undefined deprecatedReplaceInURN(
+    UrnOrConfig urnOrConfig, record&lt;USVString, USVString&gt; replacements);
+};
+</pre>
+
+<div algorithm>
+  To <dfn export for=fencedframeutil>substitute macros</dfn> with an [=ordered map=] with
+  [=string=] [=map/keys=] and [=string=] [=map/values=] |macros| into a [=string=] |string|, run
+  these steps:
+
+  1. TODO: [Spec this](https://github.com/WICG/fenced-frame/issues/116). Substitute the keys from
+     |macros| with the corresponding values into |string|, and return the new string. There should
+     be no recursive substitution.
+</div>
+
+<div algorithm>
+  The <dfn method for=Navigator method>deprecatedReplaceInURN(|urnOrConfig|, |replacements|)</dfn>
+  method steps are:
+
+  1. Let |urn| be null.
+
+  1. If |urnOrConfig| is a {{USVString}}, set |urn| to |urnOrConfig|.
+
+  1. Otherwise, set |urn| to |urnOrConfig|'s [=fencedframeconfig/urn=].
+
+  1. Let |mapping| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s
+     [=node navigable=]'s [=navigable/traversable navigable=]'s
+     [=traversable navigable/fenced frame config mapping=].
+
+  1. Let |config| be the result of [=fenced frame config mapping/finding a config=] in |mapping|
+     with |urn|.
+
+  1. If |config| is failure, [=exception/throw=] a {{TypeError}}.
+
+  1. [=map/For each=] |key| → _ of |replacements|:
+     1. If |key| does not [=string/start with=] "${" or "%%", [=exception/throw=] a {{TypeError}}.
+     1. If |key| does not [=string/end with=] "}" or "%%", [=exception/throw=] a {{TypeError}}.
+
+  1. Let |substitutedUrl| be the result of [=fencedframeutil/substituting macros=] with
+     |replacements| into |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
+
+  1. Set |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=] to |substitutedUrl|.
+
+  <wpt>
+    /fenced-frame/deprecated-config-apis.https.html
+  </wpt>
+</div>
+
 <h3 id=fence-interface>The {{Fence}} interface</h3>
 
 Several APIs specific to fenced frames are defined on the {{Fence}} interface.
--- a/spec.bs
+++ b/spec.bs
@@ -1223,10 +1223,10 @@ Each {{FencedFrameConfig}} has:
 </div>
 
 To help with ease of adoption,
-<a href=https://github.com/WICG/turtledove/issues/286#issuecomment-1682842636>until 2026</a> we
-will support the API {{Window/navigator}}.{{Navigator/deprecatedReplaceInURN()}}, which allows you
-to substitute macros into the [=fenced frame config/mapped url=] corresponding to a given
-[=urn uuid=] or {{FencedFrameConfig}}.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1222,6 +1222,64 @@ Each {{FencedFrameConfig}} has:
...
+[until 2026](https://github.com/WICG/turtledove/issues/286#issuecomment-1682842636) we will support
+the API {{Window/navigator}}.{{Navigator/deprecatedReplaceInURN()}}, which allows you to substitute
+macros into the [=fenced frame config/mapped url=] corresponding to a given [=urn uuid=] or
+{{FencedFrameConfig}}.
+
+<pre class=idl>
+typedef (USVString or FencedFrameConfig) UrnOrConfig;
+
+partial interface Navigator {
+  undefined deprecatedReplaceInURN(
```
domfarolino, line 1235: Implementation-wise [this returns a `Promise<T>`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ad_auction/navigator_auction.idl;l=41;drc=c242c647cd5854bfe9ebbc291a9bcb79a81c7ed3). Should we do the same?

spec.bs
```diff
@@ -1222,6 +1222,64 @@ Each {{FencedFrameConfig}} has:
...
+  The <dfn method for=Navigator>deprecatedReplaceInURN(|urnOrConfig|, |replacements|)</dfn>
+  method steps are:
+
+  1. Let |urn| be null.
+
+  1. If |urnOrConfig| is a {{USVString}}, set |urn| to |urnOrConfig|.
+
+  1. Otherwise, set |urn| to |urnOrConfig|'s [=fencedframeconfig/urn=].
+
+  1. Let |mapping| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s
```
domfarolino, line 1260: Just to double check, are we grabbing the right mapping? This API isn't meant to be called from _within_ a fenced frame, is it? If so then I think we need to grab the parent's mapping. I don't think that's the case but want to raise this just in case

spec.bs
```diff
@@ -1222,6 +1222,64 @@ Each {{FencedFrameConfig}} has:
...
+     [=node navigable=]'s [=navigable/traversable navigable=]'s
+     [=traversable navigable/fenced frame config mapping=].
+
+  1. Let |config| be the result of [=fenced frame config mapping/finding a config=] in |mapping|
```
domfarolino, line 1264: It seems wrong to do this synchronously, and seems to not match what the implementation does, since the implementation does this "in parallel" and returns a promise.

spec.bs
gtanzer: Oops, yup.

gtanzer: Or wait, I misread. It is indeed meant to be called from outside a fenced frame, so this access is right (like you thought). I copied this boilerplate from where Protected Audience constructs its configs.

spec.bs
gtanzer: Yeah, I should make this a promise...

gtanzer: Turned this into a promise, modeling after some parts of the Protected Audience spec. Lmk if it I didn't do it right.

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
   1. Initialize |value|'s [=fencedframeconfig/contentHeight=] to |serialized|.\[[ContentHeight]].
 </div>
 
+To help with ease of adoption,
```
domfarolino, line 1225: This should be in a note or otherwise "non-normative" disclaimed text.

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+[until 2026](https://github.com/WICG/turtledove/issues/286#issuecomment-1682842636) we will support
```
domfarolino, line 1226: This makes it sound like we're actually indeed removing the API in 2026. Are we? Or are we just "deprecating" it softly then?

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+
+  1. Let |urn| be null.
+
+  1. If |urnOrConfig| is a {{USVString}}, set |urn| to |urnOrConfig|.
+
+  1. Otherwise, set |urn| to |urnOrConfig|'s [=fencedframeconfig/urn=].
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |queue| be the result of [=starting a new parallel queue=].
```
domfarolino, line 1262: We don't need a parallel queue. We can just go `[=in parallel=]` for all of these steps instead.

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+
+  1. [=parallel queue/enqueue steps|Enqueue the following steps=] to |queue|:
+
+     1. If |urn| is TODO invalid, then [=queue a task=] to [=reject=] |p| with a "{{TypeError}}"
```
domfarolino, line 1266: Put the TODO in an XXX box

domfarolino, line 1266: Do we really need to do this in parallel? Does the implementation check this in the browser process, or does it immediately run this on the main thread and return a rejected promise? I think the difference can be observable.

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+
+     1. Let |mapping| be [=this=]'s [=relevant global object=]'s [=associated Document=]'s
+        [=node navigable=]'s [=navigable/traversable navigable=]'s
+        [=traversable navigable/fenced frame config mapping=].
+
+     1. Let |config| be the result of [=fenced frame config mapping/finding a config=] in |mapping|
+        with |urn|.
+
+     1. If |config| is failure, [=queue a task=] to [=resolve=] |p| with `undefined` and do not run
+        the remaining steps.
```
domfarolino, line 1277: Spec prose for this is usually "abort these steps".

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+
+     1. [=map/For each=] |key| → _ of |replacements|:
+        1. If |key| does not [=string/start with=] "${" or "%%", [=exception/throw=] a {{TypeError}}.
```
domfarolino, line 1280: Let's put these in `<code>` blocks instead of quotes.

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+typedef (USVString or FencedFrameConfig) UrnOrConfig;
+
+partial interface Navigator {
+  Promise&lt;undefined&gt; deprecatedReplaceInURN(
+    UrnOrConfig urnOrConfig, record&lt;USVString, USVString&gt; replacements);
+};
+</pre>
+
+<div algorithm>
+  To <dfn export for=fencedframeutil>substitute macros</dfn> with an [=ordered map=] with
```
domfarolino, line 1241: Does this have to be exported? Will anyone outside of the below algorithm us this?

spec.bs
```diff
@@ -1222,6 +1222,78 @@ Each {{FencedFrameConfig}} has:
...
+  [=string=] [=map/keys=] and [=string=] [=map/values=] |macros| into a [=string=] |string|, run
+  these steps:
+
+  1. <span class=XXX>TODO:</span> [Spec this](https://github.com/WICG/fenced-frame/issues/116).
+     Substitute the keys from |macros| with the corresponding values into |string|, and return the
+     new string. There should be no recursive substitution.
```
domfarolino, line 1247: "should" -> "must". I know this is a TODO, but if we're going to use words that technically have a normative meaning, we should use the most accurate ones.

spec.bs
gtanzer: It has been deprecated since it was added, hence the name. It should actually be removed in several years.

spec.bs
gtanzer: Yes, there is a Protected Audience spec PR that will use it:
https://github.com/WICG/turtledove/pull/801

and there are other unspec'd places in Protected Audience that will also need to use the same kind of macro substitution.

spec.bs
gtanzer: Moved the renderer-side checks out of the parallel block

domfarolino (approved):

PR #122, opened 2023-09-14
Update automatic beacon algorithm to send to all registered URLs

blu25:
Automatic beacons will now send to all URLs that registered with the reserved.top_navigation event type, regardless of what data was set when calling setReportEventDataForAutomaticBeacons(). However, automatic beacon data will only be included in the beacon if the beacon's destination is specified in the destination field when calling setReportEventDataForAutomaticBeacons().

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/122.html" title="Last updated on Sep 22, 2023, 3:15 PM UTC (d598d24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/122/bd99c82...d598d24.html" title="Last updated on Sep 22, 2023, 3:15 PM UTC (d598d24)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1449,16 +1449,18 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
   1. Let |beacon data| be |config|'s [=fenced frame config instance/fenced frame reporter=]'s
      [=fenced frame reporter/automatic beacon data=].
 
-  1. If |beacon data| is null, abort these steps.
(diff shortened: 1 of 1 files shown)


domfarolino:
Do the existing WPTs assert this behavior, or have we added any more that we could now pull into a `<wpt>` block?

blu25:
> Do the existing WPTs assert this behavior, or have we added any more that we could now pull into a `<wpt>` block?

There's one more that was just merged upstream. Added it to the list of automatic beacon WPTs.

PR #123, opened 2023-09-22
Add MediaDevices info to web platform integration.

VergeA:
Based on https://crbug.com/1307642 and corresponding discussion, we need to specify how deviceID from the MediaDevices interface is partitioned within fenced frames.

### DIFF
--- a/explainer/integration_with_web_platform.md
+++ b/explainer/integration_with_web_platform.md
@@ -93,7 +93,7 @@ The [MediaDevices interface](https://w3c.github.io/mediacapture-main/#mediadevic
 
 "To ensure stored identifiers are recognized, the identifier MUST be the same in Documents of the same origin in top-level traversables. In child navigables, the decision of whether or not the identifier is the same across documents, MUST follow the User Agent's partitioning rules for storage (such as localStorage), if any, to not interfere with mitigations for cross-site correlation."
 
-Fenced frames partition storage using a unique nonce, so that no two fenced frames can access the same partitioned storage. As a result, `deviceID` values will always be different within two fenced frames, even if their origin is the same.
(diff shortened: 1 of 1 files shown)


explainer/integration_with_web_platform.md
```diff
@@ -85,9 +85,16 @@ For more details, the implementation design doc can be found [here](https://docs
...
 
+## MediaDevices interface
+The [MediaDevices interface](https://w3c.github.io/mediacapture-main/#mediadevices) is able to list connected media devices via the [`enumerateDevices()`](https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices) method. The `deviceID` field in this method's output can potentially create consistent identifiers between same-origin frames embedded in different first-party sites. The [specification for `deviceID`](https://w3c.github.io/mediacapture-main/#dom-mediadeviceinfo-deviceid) states:
+
+"To ensure stored identifiers are recognized, the identifier MUST be the same in Documents of the same origin in top-level traversables. In child navigables, the decision of whether or not the identifier is the same across documents, MUST follow the User Agent's partitioning rules for storage (such as localStorage), if any, to not interfere with mitigations for cross-site correlation."
+
+Fenced frames partition storage using a unique nonce, so that no two fenced frames can access the same partitioned storage. As a result, `deviceID` values will always be different within two fenced frames, even if their origin is the same.
```
shivanigithub, line 96: "Fenced frames partition storage using a unique nonce, so that no other frame can access the same partitioned storage as a given fenced frame . As a result, `deviceID` values will always be different within two fenced frames and similarly the value in a fenced frame will always differ with that in other iframes/top-level frames, even if their origin is the same."

VergeA: Done.

shivanigithub (approved):
lgtm, thanks!

PR #125, opened 2023-09-28
Update permissions policy to match implementation

blu25:
The current [permisisons policy spec patches](https://wicg.github.io/fenced-frame/#permissions-policy-changes) correctly have the [load blocking algorithm](https://wicg.github.io/fenced-frame/#should-navigation-response-to-navigation-request-be-blocked-by-permissions-policy) where we won’t load a fenced frame if any of [the effective enabled permissions](https://wicg.github.io/fenced-frame/#fenced-frame-config-effective-enabled-permissions) are not set. The patches do not, however, handle setting the permissions policies correctly. The spec allows [the create a permissions policy algorithm](https://w3c.github.io/webappsec-permissions-policy/#create-for-navigable) to simply inherit from its embedder, which allows for information to leak across a fenced boundary.

This PR patches that so that the permissions that are set when a fenced frame loads are only the permissions in the "effective enabled permissions". This will ensure that all the permissions required to load are set (since that is checked at navigation time), but will not allow extra permissions to be enabled (including permissions set in the `allow` attribute of the fenced frame).

[See document for more information](https://docs.google.com/document/d/1xznk-FwL357FFW3f9NmzJfZRx4JHNOQlHzkN3JZCr5Y/edit?usp=sharing).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/125.html" title="Last updated on Nov 2, 2023, 4:46 PM UTC (f219fa2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/125/15f4ac2...f219fa2.html" title="Last updated on Nov 2, 2023, 4:46 PM UTC (f219fa2)">Diff</a>


spec.bs
```diff
@@ -2659,20 +2659,59 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+
+    1. If |effective permissions| [=list/contains=] |feature|, then set |inherited policy|[feature]
+       to "`Enabled`".
+       
+       Otherwise, set |inherited policy|[feature] to "`Disabled`".
+
+  7. Let |policy| be a new [=permissions policy=], with [=permissions policy/inherited policy=]
+     |inherited policy| and [=permissions policy/declared policy=] a new [=ordered map=].
+
+  8. Return |policy|.
```
domfarolino, line 2691: Number all of these `1.` like the rest of the spec

spec.bs
```diff
@@ -2737,11 +2782,15 @@ algorithms to achieve the outcomes described in the above explanatory content.
      creating a permissions policy$] given |navigable|'s [=fenced navigable container=], |origin|,
      and <var ignore>fenced</var> set to true.
 
-     Note: This is identical to the [=Document/permissions policy=] that will be created when the
-     navigation constructs the ultimate {{Document}} for this pending navigation. We create it now
-     and run tests on it since this is the appropriate time to determine if a navigation will fail,
-     and then throw it away. If the navigation succeeds, it will be recreated identically and
-     unconditionally installed on the {{Document}}.
+     Note: This is almost identical to the [=Document/permissions policy=] that will be created when
```
domfarolino, line 2785: This should link to https://w3c.github.io/webappsec-permissions-policy/#permissions-policy, not the `Document/` member variant

spec.bs
```diff
@@ -2737,11 +2782,15 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+     the navigation constructs the ultimate {{Document}} for this pending navigation (with the
+     [=Create a Permissions Policy for a fenced navigable=] algorithm). We create it now and run
+     tests on it since this is the appropriate time to determine if a navigation will fail, and then
+     throw it away. If the navigation succeeds, it will be recreated and unconditionally installed
```
domfarolino, line 2789: If the `allow` attribute contains `a, b, c, x, y, z`, but the config instance only contains `x, y, z`, then the permission policy created in _this_ algorithm will contain `a, b, c, x, y, z`, and we'll throw it away, right? Can we make it more clear that this permission policy has that characteristic? That's only implicitly alluded to via:

> However, the recreation will not include any additional enabled permissions that are not included in the [=fenced frame config/effective enabled permissions=],

But I think we can make it crystal clear and explicit by just saying early on in this paragraph that this, just like when it's called on iframes, will include all of the `allow` permissions, even those that are _not_ in the config. And then we can go into the second half of this paragraph where we talk about the fact that this is OK because once we actually create the real one, we just use the config's permissions, which does _not_ include the extra ones specified outside.

spec.bs
```diff
@@ -2659,20 +2659,59 @@ algorithms to achieve the outcomes described in the above explanatory content.
   or <{fencedframe}>'s [=container policy=], when it is constructed.
 </div>
 
+<div algorithm>
+  Create a new algorithm, called <dfn export>Create a Permissions Policy for a fenced
```
domfarolino, line 2663: Let's rename this to "Derive permissions policy directly from fenced frame config instance" or something super explicit like that, as discussed offline.

Also: does this need to be exported?

spec.bs
domfarolino, line 2785: nit: I'd link "created" to the "derive permissions policy directly from fenced frame config instance" and avoid the parenthesis "(with the [=foo=] algorithm)". I think this would be more clear.

spec.bs
```diff
@@ -2659,20 +2659,59 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+  navigable</dfn>.
+
+  Given null or an [=element=] (|container|), this algorithm returns a new Permissions Policy.
+
+  1. [=Assert=]: if not null, |container| is a [=fenced navigable container=].
```
domfarolino, line 2668: Do you think we can assert that `|container|` is never null? This algorithm seems to only have one caller which assures that the input container is non-null, right?

spec.bs
```diff
@@ -2659,20 +2659,59 @@ algorithms to achieve the outcomes described in the above explanatory content.
   or <{fencedframe}>'s [=container policy=], when it is constructed.
 </div>
 
+<div algorithm>
+  Create a new algorithm, called <dfn export>Create a Permissions Policy for a fenced
+  navigable</dfn>.
+
+  Given null or an [=element=] (|container|), this algorithm returns a new Permissions Policy.
```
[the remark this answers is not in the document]
blu25: Done. The "Create a Permissions Policy for a navigable from response" patch should also reference this permissions policy, rather than the Document/permissions policy that it's currently referencing.

One note is that the current "Create a Permissions Policy for a navigable" spec capitalizes "Permissions Policy", but that seems wrong.

domfarolino (approved):
Beautiful!

PR #127, opened 2023-10-03
Spec for attribution reporting integration

linnan-github:
This is migrated from the monkeypatches in attribution reporting API spec: https://wicg.github.io/attribution-reporting-api/#fenced-frame-monkeypatches

Also spec'ed the integration for automatic beacon.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/fenced-frame/pull/127.html" title="Last updated on Oct 12, 2023, 3:30 PM UTC (a7b975f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/127/ed8c323...linnan-github:a7b975f.html" title="Last updated on Oct 12, 2023, 3:30 PM UTC (a7b975f)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -251,9 +251,6 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
       text: event-source; url:eligibility-event-source
       text: navigation-source; url:eligibility-navigation-source
       text: unset; url:eligibility-unset
-    for: request
-      text: Attribution Reporting eligibility
-    text: process an attribution eligible response
 </pre>
 
 <style>
--- a/spec.bs
+++ b/spec.bs
@@ -794,7 +794,7 @@ A <dfn export for=fencedframetype>destination enum event</dfn> is a [=struct=] w
   : <dfn>attributionReportingEnabled</dfn>
   :: a [=boolean=]
 
-  : <dfn>contextOrigin</dfn>
+  : <dfn>attributionReportingContextOrigin</dfn>
   :: an [=origin=]
 </dl>
 
@@ -828,7 +828,7 @@ A <dfn export for=fencedframetype>destination event</dfn> is either a
 
     1. If |event|'s [=destination enum event/attributionReportingEnabled=] is true and the result
        of [=getting supported registrars=] is not [=list/is empty|empty=] and |event|'s
-       [=destination enum event/contextOrigin=] [=check if an origin is suitable|is suitable=]:
+       [=destination enum event/attributionReportingContextOrigin=] [=check if an origin is suitable|is suitable=]:
 
         1. If |event|'s [=destination enum event/type=] is `"reserved.top_navigation"`,
            set |attributionReportingEligibility| to "<code>[=eligibility/navigation-source=]</code>".
@@ -838,7 +838,7 @@ A <dfn export for=fencedframetype>destination event</dfn> is either a
         1. Set |processResponse| to these steps given a [=response=] |response|:
 
            1. Run [=process an attribution eligible response=] with |event|'s
-              [=destination enum event/contextOrigin=], |attributionReportingEligibility|, and |response|.
+              [=destination enum event/attributionReportingContextOrigin=], |attributionReportingEligibility|, and |response|.
 
         1. Set |useParallelQueue| to true.
 
@@ -1497,7 +1497,7 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
         1. Let |attributionReportingEnabled| be the result of determining whether |document| is
            [=allowed to use=] the "<code>{{PermissionPolicy/attribution-reporting}}</code>" feature.
 
-        1. Let |contextOrigin| be |document|'s [=node/context origin=].
+        1. Let |attributionReportingContextOrigin| be |document|'s [=node/context origin=].
 
         1. [=list/For each=] |destination| of |event|'s {{FenceEvent/destination}}:
 
@@ -1514,8 +1514,8 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
            : [=destination enum event/attributionReportingEnabled=]
            :: |attributionReportingEnabled|
 
-           : [=destination enum event/contextOrigin=]
-           :: |contextOrigin|
+           : [=destination enum event/attributionReportingContextOrigin=]
+           :: |attributionReportingContextOrigin|
 
   <wpt>
     /fenced-frame/fence-report-event.https.html
@@ -1716,7 +1716,7 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
        : [=destination enum event/attributionReportingEnabled=]
        :: |sourceSnapshotParams|'s [=source snapshot params/attribution reporting enabled=]
 
-       : [=destination enum event/contextOrigin=]
+       : [=destination enum event/attributionReportingContextOrigin=]
        :: |sourceSnapshotParams|'s [=source snapshot params/attribution reporting context origin=]
 
   1. If |beacon data|'s [=automatic beacon data/once=] is true, set |config|'s [=fenced frame
--- a/spec.bs
+++ b/spec.bs
@@ -1717,7 +1717,7 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
        :: |sourceSnapshotParams|'s [=source snapshot params/attribution reporting enabled=]
 
        : [=destination enum event/contextOrigin=]
-       :: |sourceSnapshotParams|'s [=source snapshot params/context origin=]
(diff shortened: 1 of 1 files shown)


linnan-github:
@domfarolino, Could you PTAL, thanks!

linnan-github:
@domfarolino Friendly ping, could you PTAL, thanks!

spec.bs
```diff
@@ -245,6 +245,12 @@ spec: css2; urlPrefix: https://www.w3.org/TR/CSS21/visuren.html
   type: dfn
     for: css2
       text: viewport; url: viewport
+spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-reporting-api/
+  type: dfn
+    for: eligibility
+      text: event-source; url:eligibility-event-source
+      text: navigation-source; url:eligibility-navigation-source
+      text: unset; url:eligibility-unset
```
domfarolino, line 253: nit: in all of these, let's add a space after the colon after `url:`, to match the other links here.

spec.bs
```diff
@@ -868,12 +902,16 @@ A <dfn export for=fencedframetype>destination event</dfn> is either a
...
+  1. [=Fetch=] |request| with [=fetch/processResponse=] being |processResponse| if it is not null
+     and [=fetch/useParallelQueue=] being |useParallelQueue|.
 
     Note: This algorithm can be invoked from while [=in parallel=] or while on a {{Document}}'s main
-    thread. To handle the [=in parallel=] invocation correctly, ordinarily we would invoke [=fetch=]
-    with [=fetch/useParallelQueue=] set to true, but since we don't pass in any of the <span
-    class=allow-2119>optional</span> [=response=]-handling algorithms, there is no need to do this.
+    thread. To handle the [=in parallel=] invocation correctly, we invoke [=fetch=]
+    with [=fetch/useParallelQueue=] set to true when the <span class=allow-2119>optional</span>
+    [=response=]-handling algorithm is used, otherwise there is no need to do this.
```
domfarolino, line 914: Can you link to any web platform tests here that rely on the response headers being processed by https://wicg.github.io/attribution-reporting-api/#process-an-attribution-eligible-response?

spec.bs
```diff
@@ -1667,12 +1733,22 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
 </div>
 
 <div algorithm="automatic beacon patch">
-  Modify [[HTML]]'s [=attempt to populate the history entry's document=] algorithm. In step 6,
-  substep 11, add a new step after step 5 that reads:
+  Modify [[HTML]]'s [=attempt to populate the history entry's document=] algorithm in the following ways.
+
+  First, add steps after step 5 that reads:
+
+  6. Let |attributionReportingEnabled| be the result of determining whether |navigable|'s
```
domfarolino, line 1740: This is still running in parallel, and before the new document has been installed in the target navigable/browsing context. So you're pulling the "old" origin and also checking `[=allowed to use=]` in the old document. Is that intended?

spec.bs
domfarolino, line 914: nit: `otherwise there is no need to do this even when this algorithm is running in parallel in other instances.`

spec.bs
```diff
@@ -808,6 +826,22 @@ A <dfn export for=fencedframetype>destination event</dfn> is either a
...
+       [=destination enum event/contextOrigin=] [=check if an origin is suitable|is suitable=]:
+
+        1. If |event|'s [=destination enum event/type=] is `"reserved.top_navigation"`,
+           set |attributionReportingEligibility| to "<code>[=eligibility/navigation-source=]</code>".
+
+        1. Otherwise, set |attributionReportingEligibility| to "<code>[=eligibility/event-source=]</code>".
+
+        1. Set |processResponse| to these steps given a [=response=] |response|:
+
+           1. Run [=process an attribution eligible response=] with |event|'s
```
domfarolino, line 840: Please make sure that https://wicg.github.io/attribution-reporting-api/#process-an-attribution-eligible-response can handle running in parallel. Right now it does not, because https://wicg.github.io/attribution-reporting-api/#obtain-and-deliver-a-debug-report "queues a task" to no explicit task source, which seems wrong. There may be more wrong with that algorithm, but that's at least my initial reaction, so following-up on that spec might be good.

spec.bs
linnan-github: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/attribution-reporting/
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/wpt_internal/attribution-reporting/
Most of the tests in these two suites rely on that.

spec.bs
linnan-github: Yeah, that's intended to check the initiator origin, therefore the old origin.

spec.bs
linnan-github: Thanks for catching this, I will file a PR to fix that in attribution reporting spec.

spec.bs
domfarolino: Ah, well the initiator origin may not be the "old document" origin, right? Any frame might have initiated this document. So for any navigation there are three possible origins:
 - The initiator origin
 - The target frame's old document origin
 - The target document's _new_ origin

Which do you want here?

spec.bs
domfarolino: Great, can you link to some of them via a block like this: https://github.com/WICG/fenced-frame/blob/master/spec.bs#L1660-L1666 ? Just the external/wpt ones will be linkable with that.

spec.bs
linnan-github: We want the initiator origin, and check [=allowed to use=] on the initiator frame. What's the proper way to get those?

domfarolino: You'll want to grab source snapshot params's fetch client's [origin](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin) for this. Although, I'm curious: what does the implementation do if that initiator document is already gone by the time we make it here?

spec.bs
linnan-github: Actually those wpts are for attribution reporting only, not fenced frame related, so I think it's better to link to ARA spec.

spec.bs
domfarolino: Or, I guess you'll want the "associated Document", not the "origin"

spec.bs
```diff
@@ -2392,14 +2381,20 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
   : <dfn for="source snapshot params">initiator fenced frame config instance</dfn>
   :: a [=fenced frame config instance=] or null, initially null.
 
   : <dfn for="source snapshot params">target fenced frame config</dfn>
   :: a [=fenced frame config=] or null, initially null.
 
+  : <dfn for="source snapshot params">attribution reporting enabled</dfn>
+  :: a [=boolean=].
+
+  : <dfn for="source snapshot params">context origin</dfn>
```
domfarolino, line 2395: Could we rename this to "attribution reporting context origin", since it is very specific to that spec?

spec.bs
domfarolino: OK, that sounds fine.

spec.bs
linnan-github: Updated, thanks!

spec.bs
```diff
@@ -2335,11 +2407,19 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
-  :: <var ignore>sourceDocument</var>'s [=browsing context=]'s [=browsing context/fenced frame
+  :: |sourceDocument|'s [=browsing context=]'s [=browsing context/fenced frame
      config instance=]
+
+  : [=source snapshot params/attribution reporting enabled=]
+  :: The result of determining whether |sourceDocument| is [=allowed to use=] the
+     "<code>{{PermissionPolicy/attribution-reporting}}</code>" feature
+
+  : [=source snapshot params/attribution reporting context origin=]
+  :: |sourceDocument|'s [=node/context origin=]
```
domfarolino, line 2421: From reading https://wicg.github.io/attribution-reporting-api/#node-context-origin, it really seems like the "context origin" is meant to be called/gathered in the context of a fenced frame's main document. But here, we're calling it on the _source document_, which means we're expecting the source document to be a fenced frame, which doesn't seem right. If you want this member to be set to the fenced frame's embedder document, I'd just grab `|sourceDocument|'s `[=Document/origin=]` directly. I'm not sure what the "context origin" is used for elsewhere.

domfarolino: Seems like this is mostly resolved now because https://github.com/WICG/attribution-reporting-api/pull/1070.

domfarolino (approved):

PR #130, opened 2023-10-23
Spec reserved.top_navigation_start and reserved.top_navigation_commit

blu25:
This PR does a few things to allow the new automatic beacon types:

- ~~Creates a new "valid automatic beacon event types" list that contains the start and commit beacon types.~~
-  Creates a new implicit enumerated type for the automatic beacon event types. This contains the start and commit beacon types.
- Creates a new "automatic beacon event" to be used for automatic beacons instead of "destination enum event", which contains the aforementioned implicit enumerated event type.
- Splits out automatic beacon data into its own struct.
- Turns the fenced frame reporter's automatic beacon data into a mapping to match the implementation.
- Passes "reserved.top_navigation_commit" as a parameter when calling the automatic beacon algorithm from its existing place at navigation commit.
- Also calls the automatic beacon algorithm at navigation start passing in "reserved.top_navigation_start".
- Removes "reserved.top_navigation".

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/130.html" title="Last updated on Nov 7, 2023, 4:43 PM UTC (e922439)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/130/dc24e72...e922439.html" title="Last updated on Nov 7, 2023, 4:43 PM UTC (e922439)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1746,16 +1746,16 @@ traversable=].
 <div algorithm="top_navigation_start beacon patch">
   Modify [[HTML]]'s [=navigate=] algorithm. Add a new step after step 4 that reads:
   
-  5. [=attempt to send an automatic beacon=] given <var ignore>sourceDocument</var>, <var
-     ignore>sourceSnapshotParams</var>, <var ignore>navigable</var>’s associated {{Document}}, and
-     "`reserved.top_navigation_start`".
+  5. [=Attempt to send an automatic beacon=] given <var ignore>sourceSnapshotParams</var>, <var
+     ignore>initiatorOriginSnapshot</var>, <var ignore>navigable</var>’s associated {{Document}},
+     and "`reserved.top_navigation_start`".
 </div>
 
 <div algorithm="top_navigation_commit beacon patch">
   Modify [[HTML]]'s [=attempt to populate the history entry's document=] algorithm. In step 6,
   substep 11, add a new step after step 5 that reads:
 
-  6. if <var ignore>failure</var> is false, then [=attempt to send an automatic beacon=] given <var
+  6. If <var ignore>failure</var> is false, then [=attempt to send an automatic beacon=] given <var
      ignore>sourceSnapshotParams</var>, <var ignore>entry</var>'s [=document state=]'s [=document
      state/initiator origin=], <var ignore>document</var>, and "`reserved.top_navigation_commit`".
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -762,8 +762,14 @@ following [=struct/items=]:
 </dl>
 
 An <dfn for=fencedframetype>automatic beacon event type</dfn> is either "<dfn
-for="automatic beacon event type">`reserved.top_navigation_start`</dfn>" or "<dfn
-for="automatic beacon event type">`reserved.top_navigation_commit`</dfn>".
+for="automatic beacon event type">`reserved.top_navigation_start`</dfn>", "<dfn
+for="automatic beacon event type">`reserved.top_navigation_commit`</dfn>", or "<dfn
+for="automatic beacon event type">`reserved.top_navigation`</dfn>".
+
+Advisement: <code>[=automatic beacon event type/reserved.top_navigation=]</code> is an earlier
+naming of <code>[=automatic beacon event type/reserved.top_navigation_commit=]</code>. While they
+both do the same thing, <code>[=automatic beacon event type/reserved.top_navigation=]</code> will be
+depreacted in the future and should not be used for new code.
 
 An <dfn for=fencedframetype>automatic beacon data</dfn> is a [=struct=] with the following
 [=struct/items=]:
@@ -1690,10 +1696,11 @@ A side effect of the fenced boundary model is that ads will lose the ability to 
 resulted in a successful navigation. This is because the page loaded from a top-level [=navigate|
 navigation=] originating from a fenced frame will not be allowed to report to the fenced frame that
 it loaded (through something like <code>window.[=Window/opener=]</code>). Instead, we introduce
-special event-level [=destination enum event/type|reporting type=]s, `reserved.top_navigation_start`
-and `resrved.top_navigation_commit`, which automatically sends an [=report an event|event-level
-beacon=] when a fenced frame initiates a successful [=navigate|navigation=] to a [=top-level
-traversable=].
+special event-level [=automatic beacon event/type|reporting type=]s, <code>[=automatic beacon event
+type/reserved.top_navigation_start=]</code> and <code>[=automatic beacon event
+type/reserved.top_navigation_commit=]</code>, which automatically sends an [=report an
+event|event-level beacon=] when a fenced frame initiates a successful [=navigate|navigation=] to a
+[=top-level traversable=].
 
 <div algorithm>
   To <dfn>attempt to send an automatic beacon</dfn> given a [=source snapshot params=]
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1728,13 +1743,21 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
   </wpt>
 </div>
 
-<div algorithm="automatic beacon patch">
+<div algorithm="top_navigation_start beacon patch">
+  Modify [[HTML]]'s [=navigate=] algorithm. Add a new step after step 4 that reads:
+  
+  5. [=attempt to send an automatic beacon=] given <var ignore>sourceDocument</var>, <var
```
domfarolino, line 1749: Capital "A"? For a new sentence?

domfarolino, line 1749: The signature for this algorithm does not take a sourceDocument as the first parameter.

spec.bs
```diff
@@ -1658,17 +1689,16 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
 A side effect of the fenced boundary model is that ads will lose the ability to know if a click
 resulted in a successful navigation. This is because the page loaded from a top-level [=navigate|
 navigation=] originating from a fenced frame will not be allowed to report to the fenced frame that
-it loaded (through something like <code>window.[=Window/opener=]</code>). Instead, we introduce a
-special event-level [=destination enum event/type|reporting type=], `reserved.top_navigation`, which
-automatically sends an [=report an event|event-level beacon=] when a fenced frame initiates a
-successful [=navigate|navigation=] to a [=top-level traversable=].
+it loaded (through something like <code>window.[=Window/opener=]</code>). Instead, we introduce
+special event-level [=destination enum event/type|reporting type=]s, `reserved.top_navigation_start`
+and `resrved.top_navigation_commit`, which automatically sends an [=report an event|event-level
```
domfarolino, line 1694: Can you link to these constant values, since they're now linkable?

spec.bs
```diff
@@ -1658,17 +1689,16 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
 A side effect of the fenced boundary model is that ads will lose the ability to know if a click
 resulted in a successful navigation. This is because the page loaded from a top-level [=navigate|
 navigation=] originating from a fenced frame will not be allowed to report to the fenced frame that
-it loaded (through something like <code>window.[=Window/opener=]</code>). Instead, we introduce a
-special event-level [=destination enum event/type|reporting type=], `reserved.top_navigation`, which
-automatically sends an [=report an event|event-level beacon=] when a fenced frame initiates a
-successful [=navigate|navigation=] to a [=top-level traversable=].
+it loaded (through something like <code>window.[=Window/opener=]</code>). Instead, we introduce
+special event-level [=destination enum event/type|reporting type=]s, `reserved.top_navigation_start`
```
domfarolino, line 1693: Right now the "reporting type" member is a string, but should it / could it be an [automatic beacon event type instead](https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/130.html#fencedframetype-automatic-beacon-event-type)? And please keep the code font when you link (see my last comment below in this review).

spec.bs
```diff
@@ -1728,13 +1758,22 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
   </wpt>
 </div>
 
-<div algorithm="automatic beacon patch">
+<div algorithm="top_navigation_start beacon patch">
+  Modify [[HTML]]'s [=navigate=] algorithm. Add a new step after step 4 that reads:
+  
+  5. [=Attempt to send an automatic beacon=] given <var ignore>sourceSnapshotParams</var>, <var
+     ignore>initiatorOriginSnapshot</var>, <var ignore>navigable</var>’s associated {{Document}},
+     and "[=automatic beacon event type/reserved.top_navigation_start=]".
```
domfarolino, line 1766: Use code format when linking to this (backticks I guess?)

spec.bs
```diff
@@ -1728,13 +1758,22 @@ successful [=navigate|navigation=] to a [=top-level traversable=].
...
+<div algorithm="top_navigation_commit beacon patch">
   Modify [[HTML]]'s [=attempt to populate the history entry's document=] algorithm. In step 6,
   substep 11, add a new step after step 5 that reads:
 
-  6. if <var ignore>failure</var> is false, then [=attempt to send an automatic beacon=] given <var
+  6. If <var ignore>failure</var> is false, then [=attempt to send an automatic beacon=] given <var
      ignore>sourceSnapshotParams</var>, <var ignore>entry</var>'s [=document state=]'s [=document
-     state/initiator origin=], <var ignore>document</var>, and "`reserved.top_navigation`".
+     state/initiator origin=], <var ignore>document</var>, and "[=automatic beacon event
+     type/reserved.top_navigation_commit=]".
```
domfarolino, line 1776: same as above

michal-kalisz:
This change appears to be very valuable. However, it breaks backward compatibility (discontinues support for the automatic beacon event type: reserved.top_navigation). When the change will be merged and available in chrome, some users will have an older version of the browser (requiring reserved.top_navigation), and some will have a newer one (supporting reserved.top_navigation_(start|commit)).

Tracking clicks in our current tests is a critical component. Therefore, we are concerned that the proposed change should not affect either our tests or our partners' (SSP) tests.

How can we prepare for this? Would registering 3 automatic beacons: reserved.top_navigation, reserved.top_navigation_start, reserved.top_navigation_commit and then calling window.fence.setReportEventDataForAutomaticBeacons for 3 different event type in the ad be a solution?

Something like: 
In report win / report result: 
```
function reportWin() {
    registerAdBeacon({
        "reserved.top_navigation": "URL_for_commit",
        "reserved.top_navigation_start": "URL_for_start",
        "reserved.top_navigation_commit": "URL_for_commit"
    })
}
```

In ad: 
    
```
window.fence.setReportEventDataForAutomaticBeacons({
    eventType: 'reserved.top_navigation',
    destination: ['buyer', 'direct-seller']
});
window.fence.setReportEventDataForAutomaticBeacons({
    eventType: 'reserved.top_navigation_start',
    destination: ['buyer', 'direct-seller']
});
window.fence.setReportEventDataForAutomaticBeacons({
    eventType: 'reserved.top_navigation_commit',
    destination: ['buyer', 'direct-seller']
});

```


Or do you suggest something else?

blu25:
To clarify, `reserved.top_navigation` will continue to work in new versions of Chrome for the time being.

Our current plan is to remove `reserved.top_navigation` functionality at some point in the future once enough partners have switched over to `reserved.top_navigation_commit`. The initial version of this spec patch reflects what we want the eventual end-state to be. However, I can't commit to any timeline for actually removing `reserved.top_navigation`. In the meantime, and to avoid confusion about what is being supported on the Chrome side, I'll add it back to the spec with an advisement that it will be deprecated.

In terms of preparing for this change, what you described would be the preferred way of handling this. Once enough people are on the newer version of Chrome, you can remove the `reserved.top_navigation` part of your code.

blu25:
Yes. That's in a separate repository, and I [have a PR up](https://github.com/WICG/turtledove/pull/885) for that as well.

spec.bs
```diff
@@ -761,25 +761,40 @@ following [=struct/items=]:
...
 
+An <dfn for=fencedframetype>automatic beacon event type</dfn> is either "<dfn
+for="automatic beacon event type">`reserved.top_navigation_start`</dfn>", "<dfn
+for="automatic beacon event type">`reserved.top_navigation_commit`</dfn>", or "<dfn
+for="automatic beacon event type">`reserved.top_navigation`</dfn>".
+
+Advisement: <code>[=automatic beacon event type/reserved.top_navigation=]</code> is an earlier
+naming of <code>[=automatic beacon event type/reserved.top_navigation_commit=]</code>. While they
+both do the same thing, <code>[=automatic beacon event type/reserved.top_navigation=]</code> will be
+depreacted in the future and should not be used for new code.
```
domfarolino, line 772: Two things here.

1. Deprecated is spelled wrong
2. It is deprecated *now*, so let's say it will be **removed** in the future (i.e., we'll follow-up on that deprecation).

domfarolino (commented):
LGTM % nit

domfarolino (approved):

PR #133, opened 2023-12-04
Spec out cross-origin automatic beacon support

blu25:
[See document outlining the changes.](https://docs.google.com/document/d/11S8a_C6XeMjvTWxytjGcZ_IzN-IFBLugWR0iVSKrPDA/edit?usp=sharing)

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/133.html" title="Last updated on Jan 10, 2024, 8:58 PM UTC (09e9651)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/133/feff58e...09e9651.html" title="Last updated on Jan 10, 2024, 8:58 PM UTC (09e9651)">Diff</a>


domfarolino:
Looks like there is a conflict now. Would you be able to fix this before I take a closer look?

spec.bs
```diff
@@ -1842,6 +1845,34 @@ Each {{Window}} object has an associated <dfn for=Window>fence</dfn>, which is a
   </wpt>
 </div>
 
+<h3 id=document-extension>Extensions to the {{Document}} interface</h3>
```
domfarolino, line 1848: "Interface" is specifically used when we're modifying or extending an IDL interface, but the stuff in this section is more about internal infra / supporting concepts that are scoped to a Document. Given that, I'd either:
 1. Rename this section to `{{Document}} supporting concepts`
 2. Move its contents elsewhere we define things on the "Document" object (assuming we do that elsewhere in the spec).

I'll let you make the call.

spec.bs
```diff
@@ -2471,19 +2502,52 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
-  reporter/automatic beacon data map=] to be sent out if the <{fencedframe}>-initiated navigation
-  succeeds. The [=source snapshot params/target fenced frame config=] on the other hand, is the
-  non-[=instantiate a config|instantiated=] [=fenced frame config=] that will be loaded into a a
-  <{fencedframe}> element for navigations targeting fenced frames. These fields do not interact
-  *together* in any meaningful way.
+  It is used by the [=attempt to send an automatic beacon=] algorithm to compare [=origin=]s and
+  determine which {{FenceReportingDestination}}s to send beacons to, if the
+  <{fencedframe}>-initiated navigation succeeds. The [=source snapshot params/target fenced frame
+  config=] on the other hand, is the non-[=instantiate a config|instantiated=] [=fenced frame
+  config=] that will be loaded into a a <{fencedframe}> element for navigations targeting fenced
```
domfarolino, line 2518: double "a" here

spec.bs
```diff
@@ -2471,19 +2502,52 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+  frames. These fields do not interact *together* in any meaningful way.
```
domfarolino, line 2519: This is a nice disclaimer, thanks for adding it.

spec.bs
```diff
@@ -2471,19 +2502,52 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+</div>
+
+<div algorithm>
+  To <dfn>get the automatic beacon data mapping to use</dfn> given a {{Document}} |sourceDocument|:
+
+  1. Let |automatic beacon data map| be an empty [=map=] whose [=map/keys=] are [=fencedframetype/
+     automatic beacon event type=]s and whose [=map/values=] are null or an [=fencedframetype/
+     automatic beacon data=].
```
domfarolino, line 2527: Maybe we could just make this a new empty "automatic beacon data map", since this seems to match exactly that definition's shape. It feels weird to initialize something as a "new" type that is only defined as a member type, but honestly it's probably fine and making this less redundant is probably worth it.

domfarolino (commented):
Looking pretty good

spec.bs
blu25: I don't think we modify `{{Document}}` in the same way elsewhere in the spec, so I'll rename the section.

domfarolino (approved):

PR #134, opened 2023-12-18
Prevent reserved beacons from being sent manually.

blu25:
Reporting an event through the `window.fence.reportEvent()` call should not be allowed to work if it's passed in a "reserved." event type.

This PR also exports `automatic beacon event type`, which will be used by the TURTLEDOVE spec for a similar patch to `registerAdBeacon()`.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/134.html" title="Last updated on Dec 18, 2023, 7:24 PM UTC (b94150e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/134/79fad8a...b94150e.html" title="Last updated on Dec 18, 2023, 7:24 PM UTC (b94150e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -761,7 +761,7 @@ following [=struct/items=]:
   :: a [=boolean=], initially false
 </dl>
 
-An <dfn for=fencedframetype>automatic beacon event type</dfn> is either "<dfn
+An <dfn export for=fencedframetype>automatic beacon event type</dfn> is either "<dfn
 for="automatic beacon event type">`reserved.top_navigation_start`</dfn>", "<dfn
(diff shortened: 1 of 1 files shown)


domfarolino (approved):
Let's just make sure the Protected Audience indeed makes use of this term so we don't accidentally export it for no reason.

blu25:
> Let's just make sure the Protected Audience indeed makes use of this term so we don't accidentally export it for no reason.

I just put up https://github.com/WICG/turtledove/pull/968 for review which references this term.

PR #140, opened 2024-01-22
Allow some fenced frames to inherit permissions

blu25:
See the "Permissions Changes" section of [this document](https://docs.google.com/document/d/1RSufNtL0KKTvmS9ZajJqyhkuxFfnyIW2hnLE9Mb9ehw/edit?usp=sharing) for details.

Fenced frames with unpartitioned data access need to be able to allow certain permissions policy-backed features (mainly Shared Storage) in order to function properly. This PR accomplishes that by allowing these fenced frames to inherit permissions policies the way that iframes do, but only allows a select list of permissions policies to be enabled. More specifically, this PR:

- Introduces the concept of a "fixed permissions policy" and "flexible permissions policy" in a FencedFrameConfig.
- Removes `Derive a permissions policy directly from a fenced frame config instance` in favor of a new `Create a permissions policy for a fenced navigable` algorithm, which handles both "fixed" and "flexible" permissions policies being created in a fenced frame.
- Renames the `fenced` parameter used when checking if navigation to a fenced frame should be blocked by permissions policy to `matches all`. This is done to avoid confusion with the `Create a permissions policy for a fenced navigable` algorithm and to make it clear that the boolean specifically affects whether a permission will be inherited in the ultimately created permissions policy if the allowlist matches itself or just matches the wildcard `*`.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/140.html" title="Last updated on Dec 10, 2024, 8:09 PM UTC (6156904)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/140/14f148d...6156904.html" title="Last updated on Dec 10, 2024, 8:09 PM UTC (6156904)">Diff</a>


spec.bs
```diff
@@ -3013,47 +3040,57 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
-  Given null or an [=element=] (|container|), this algorithm returns a new [=permissions policy=].
+  Given null or an [=element=] (|container|) and an [=origin=] (|origin|), this algorithm returns a
+  new [=Permissions Policy=].
 
-  1. Let |inherited policy| be a new [=ordered map=].
+  1. [=Assert=]: if not null, container is a [=fenced navigable container=].
 
-  1. Let |effective permissions| be an empty [=list=].
-  
   1. Let |fenced frame config| be |container|'s [=Node/node document=]'s [=navigable/active browsing
```
domfarolino, line 3050: It seems like there is a pre-existing bug in this algorithm where we unconditionally grab things off of `|container|` which may be null, right?

spec.bs
```diff
@@ -2917,20 +2917,27 @@ directive wouldn't give web sites enough control over their CSP rules. Introduce
...
-config instance=]'s [=fenced frame config instance/effective enabled permissions=] is consulted
-[=Should navigation response to navigation request be blocked by Permissions Policy?|during
-navigation=]. A <{fencedframe}> navigation can only succeed if the [=Document/permissions policy=]
-for the navigation's resulting {{Document}} has an [=permissions policy/inherited policy=] such that
-the [=inherited policy for a feature|inherited policy value=] is "`Enabled`" for each feature in the
+The [=policy-controlled features=] available to {{Document}}s inside of a <{fencedframe}>, as well
+as the manner in which they are calculated, vary depending on how the [=fenced frame config=] that
+the <{fencedframe}> navigates to is constructed.
+
+A {{FencedFrameConfig}} constructed through the web platform will be considered to have a [=fenced
```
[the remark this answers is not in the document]
blu25: Clarified this to say a `[=fenced frame config instance=]` created via the `{{FencedFrameConfig}}` constructor.

spec.bs
blu25: I think allowing `|container|` to be set to null is incorrect.

> If container is a [fenced navigable container](https://github.com/WICG/fenced-frame/pull/140/files/207f86fd773cfcc629867cde249766c27c56d713#fenced-navigable-containers), then let policy be the result of running [create a permissions policy for a fenced navigable](https://github.com/WICG/fenced-frame/pull/140/files/207f86fd773cfcc629867cde249766c27c56d713#create-a-permissions-policy-for-a-fenced-navigable) given container and origin.

This line, which is the only place that `Create a permissions policy for a fenced navigable` is called, ensures that `|container|` will always be a fenced navigable container inside the algorithm. So instead I think we need to assert just that `|container|` is a fenced navigable container, and not allow `null` to be an option for the parameter.

Alternatively we can have the `|container|` parameter in the algorithm be a `fenced navigable container` rather than just an `element`, but I'm going to keep it just as an `element` to be consistent with [Create a Permissions Policy for a navigable](https://w3c.github.io/webappsec-permissions-policy/#create-for-navigable).

domfarolino:
I took another look at this after finding it my inbox. Feel free to resolve conflicts and formally re-request my review (so it appears in my dashboard) and I'll take another look at this PR.

spec.bs
```diff
@@ -3322,22 +3333,32 @@ directive wouldn't give web sites enough control over their CSP rules. Introduce
...
+frame config/effective enabled permissions=] will be considered to have a [=permissions policy
+behavior/fixed=] [=fencedframetype/permissions policy behavior=], and the inner {{Document}} of the
+<{fencedframe}> it navigates to will have the [=fenced frame config/effective enabled permissions=]
+be the exclusive list of [=policy-controlled features=] that will be enabled in the {{Document}}
+(all others will be disabled).
+
+During a <{fencedframe}> navigation to a [=fenced frame config instance=] with a [=permissions
+policy behavior/fixed=] [=fencedframetype/permissions policy behavior=], it checks the [=fenced
+frame config instance/effective enabled permissions=] in the [=browsing context/fenced frame config
+instance=] stored in the [=browsing context=], and only allows the navigation if the
```
domfarolino, line 3356: The middle of this is a bit wordy. Instead of "it checks ....", I think maybe this would be better:

```
it compares the [=fenced frame config instance/effective enabled permissions=]
of the [=fenced frame config instance=] being navigated to, against the resulting
{{Document}}'s [=Document/permissions policy=]'s
[=permissions policy/inherited policy=]. The navigation only succeeds if each
inherited feature whose [=inherited policy for a feature|inherited policy value=] is
"`Enabled`" also appears in the [=fenced frame config instance/effective enabled permissions=]
[=fenced frame config instance=]. Otherwise, the environment........
```

(but with real wrapping)

How does that sound?

spec.bs
```diff
@@ -3418,47 +3445,57 @@ algorithms to achieve the outcomes described in the above explanatory content.
 </div>
 
 <div algorithm>
-  Create a new algorithm, called <dfn>Derive a permissions policy directly from a fenced frame
-  config instance</dfn>.
+  Create a new algorithm, called <dfn>Create a permissions policy for a fenced navigable</dfn>.
 
-  Given null or an [=element=] (|container|), this algorithm returns a new [=permissions policy=].
+  Given an [=element=] (|container|) and an [=origin=] (|origin|), this algorithm returns a
```
domfarolino, line 3450: Do you know why this used to accept possibly-null elements, and why that is no longer necessary?

spec.bs
```diff
@@ -3418,47 +3445,57 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
 
+      1. If |feature| matches one of the [=fenced frame allowed permissions=], then set
+        |inherited policy|[feature] to the result of running [$Define an inherited policy for
+        feature in container at origin$] given |feature|, |container|, and |origin|.
+
+         Otherwise, set |inherited policy|[feature] to "`Disabled`".
+    
   1. Let |policy| be a new [=permissions policy=], with [=permissions policy/inherited policy=]
-     |inherited policy| and [=permissions policy/declared policy=] a new [=ordered map=].
+     |inherited policy| and [=permissions policy/declared policy=] «[], []».
```
domfarolino, line 3488: I don't think this change is valid. I'd say "initialized to two new [=ordered map=]s", since a declared policy is a struct with two ordered maps, but you've specified something that isn't quite valid. A single empty map would be `«[]»`, but you're using `«»` as if it were struct-literal syntax which doesn't exist. I think you mean something like `{«[]», «[]»}` (i.e., an "object"/struct that contains two empty maps), except `{}` syntax isn't real, so I'd just explicitly mention "initialized to two new ordered maps".

spec.bs
blu25: That looks better. I've updated that section.

spec.bs
blu25: An oversight. The one place where this algorithm is called checks that container is a fenced navigable container before passing it into this algorithm, so there's no chance that a null container will be passed into here.

spec.bs
blu25: I've replaced that with the phrase "initialized to two new ordered maps". If that's still too vague, I can explicitly create a new |declared policy| variable in a separate step and this to that variable's value.

spec.bs
```diff
@@ -775,6 +775,10 @@ following [=struct/items=]:
   :: a [=string=]
 </dl>
 
+A <dfn export for=fencedframetype>permissions policy behavior</dfn> is either "<dfn export
```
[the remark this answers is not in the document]
blu25: I'll start out not exporting any of them, and if down the line we need this accessible in other specs I can export them on a per-case basis.

spec.bs
```diff
@@ -3711,47 +3737,58 @@ algorithms to achieve the outcomes described in the above explanatory content.
 </div>
 
 <div algorithm>
-  Create a new algorithm, called <dfn>Derive a permissions policy directly from a fenced frame
-  config instance</dfn>.
+  Create a new algorithm, called <dfn>Create a permissions policy for a fenced navigable</dfn>.
 
-  Given null or an [=element=] (|container|), this algorithm returns a new [=permissions policy=].
+  Given an [=element=] (|container|) and an [=origin=] (|origin|), this algorithm returns a
```
domfarolino, line 3742: Actually, let's just make the type of this parameter a `[=fenced navigable container=]` instead of an `[=element=]`. Then we can remove the assert below.

spec.bs
```diff
@@ -3711,47 +3737,58 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
-       to "`Enabled`".
-       
-       Otherwise, set |inherited policy|[feature] to "`Disabled`".
+  1. Otherwise:
+
+    1. [=list/For each=] |feature| [=supported features|supported=]:
 
+      1. If |feature| matches one of the [=fenced frame allowed permissions=], then set
+        |inherited policy|[feature] to the result of running [$Define an inherited policy for
+        feature in container at origin$] given |feature|, |container|, and |origin|.
```
domfarolino, line 3775: We're intentionally not passing in `|fenced|` here, and that's correct, right?

spec.bs
```diff
@@ -3711,47 +3737,58 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
-    1. If |effective permissions| [=list/contains=] |feature|, then set |inherited policy|[feature]
-       to "`Enabled`".
-       
-       Otherwise, set |inherited policy|[feature] to "`Disabled`".
+  1. Otherwise:
+
+    1. [=list/For each=] |feature| [=supported features|supported=]:
 
+      1. If |feature| matches one of the [=fenced frame allowed permissions=], then set
+        |inherited policy|[feature] to the result of running [$Define an inherited policy for
```
domfarolino, line 3774: This line and the below line should have one more preceding space, to line them up with the `I` in `If` above.

spec.bs
```diff
@@ -3830,20 +3867,20 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
-     algorithm, just like when it is called on iframes, will include all of the permissions
-     specified in the <{fencedframe/allow}> attribute, even if that permission isn't specified in
-     the [=fenced frame config=]'s [=fenced frame config/effective enabled permissions=]. We create
-     it now and run tests on it since this is the appropriate time to determine if a navigation will
-     fail, and then throw it away. If the navigation succeeds, it will be recreated and
-     unconditionally installed on the {{Document}}. However, the recreation will not include any
-     additional enabled permissions that are not included in the [=fenced frame config/effective
-     enabled permissions=], effectively locking the enabled permissions to only what is specified in
-     [=fenced frame config/effective enabled permissions=].
+     and <var ignore>match all</var> set to true.
```
domfarolino, line 3870: I think what was here before your change was slightly wrong. We should've just been passing "true" since `<var ignore>fenced</var>` isn't a linkable definition. So let's just pass in true.

domfarolino (commented):
I think we're just about there.

spec.bs
blu25: Correct. `|fenced|`/`|match all|` is only needed when determining whether to block navigation. In this instance, we want "flexible" permissions fenced frames to behave the same as their iframe counterparts (for permissions we allow to be enabled), so we want `|match all|` to be set to their default "false" value.

domfarolino (approved):
Thanks!

PR #146, opened 2024-04-03
Spec disableUntrustedNetwork API surface

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/146.html" title="Last updated on Nov 1, 2024, 6:54 PM UTC (026b875)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/146/eae30ee...026b875.html" title="Last updated on Nov 1, 2024, 6:54 PM UTC (026b875)">Diff</a>


spec.bs
```diff
@@ -1699,6 +1715,44 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   </wpt>
 </div>
 
+<div algorithm>
+  The <dfn method for=Fence>disableUntrustedNetwork()</dfn> method steps are:
+
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then return.
```
gtanzer, line 1724: Really these "return"s before creating the promise shouldn't be returns; the API just shouldn't be available in those situations. Not sure how to actually specify that.

spec.bs
```diff
@@ -1143,11 +1143,14 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
 
   : <dfn>embedder shared storage context</dfn>
   :: null, or a [=string=]
-  
+
   : <dfn>is ad component</dfn>
-  :: A [=boolean=]. Defaulting to false.
+  :: A [=boolean=], defaults to false.
```
domfarolino, line 1148:
```suggestion
  :: A [=boolean=], initially false.
```

spec.bs
```diff
@@ -1143,11 +1143,14 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
...
+
+  : <dfn>can disable untrusted network</dfn>
+  :: A [=boolean=], defaults to true.
```
domfarolino, line 1151:
```suggestion
  :: A [=boolean=], initially true.
```

spec.bs
```diff
@@ -1196,8 +1199,14 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
 
   : <dfn>embedder shared storage context</dfn>
   :: null, or a [=string=]
-  
+
   : <dfn>is ad component</dfn>
+  :: A [=boolean=]
+
+  : <dfn>can disable untrusted network</dfn>
+  :: A [=boolean=]
```
domfarolino, line 1207: Do these ones not need defaults?

spec.bs
domfarolino: What do you mean it shouldn't be available? Do we neuter the API surface in the Blink implementation somehow? I would think these should just return rejected promises.

spec.bs
```diff
@@ -1699,6 +1715,44 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |fencedFrameNonce| be |instance|'s [=fenced frame config instance/partition nonce=].
+
+     1. <span class=XXX>TODO: Let |credentiallessNonce| be the page credentialless nonce.</span>
```
domfarolino, line 1741: These TODOs could probably use some single follow-up bug that references all of the remaining work to do here, so that these tasks don't get lost forever here.

spec.bs
gtanzer: The fields in `config instance` are always initialized by `instantiate a config`. I wanted to specify defaults for the fields in `config` so that if other specs don't fill them in when constructing their config they get default values. So I don't know if it's really "initial" values per se.

spec.bs
gtanzer: I guess I was thinking of window.fence not being available, but for these cases we just return `ScriptPromise<IDLUndefined>()` and log console errors.

gtanzer:
Addressed comments % a question about "default" vs "initial" values.

domfarolino:
Can you just "resolve" all comment threads that have been resolved at this point?

domfarolino:
And resolve the merge conflict that this PR now has with `master`?

gtanzer:
@domfarolino Fixed merge conflicts.

The open threads aren't resolved per se. They depend on your response here: (whether I should commit the suggestions or do nothing)
https://github.com/WICG/fenced-frame/pull/146/#discussion_r1559812615

spec.bs
domfarolino: OK. So should `has disabled untrusted network` not have `, initially false` below it then?

gtanzer: That one is initially false because you always started with network disabled (has disabled=false), and then (if you can disable network) you later disable it (has disabled->true).

domfarolino:
I don't care which word we use "default" vs "initial". I just want it to be consistent, and this PR makes it inconsistent with https://github.com/WICG/fenced-frame/pull/146/files#diff-6f5a1d8263b0b0c42e2716ba5750e3652e359532647ac934c1c70086ae3ceddaR1213. So I'd either make this PR consistent with https://github.com/WICG/fenced-frame/pull/146/files#diff-6f5a1d8263b0b0c42e2716ba5750e3652e359532647ac934c1c70086ae3ceddaR1213 or make that line consistent with this PR. Whichever word you prefer.

spec.bs
```diff
@@ -1738,6 +1754,52 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+     Issue: Spec the revocation process.
+     (<a href="https://github.com/WICG/fenced-frame/issues/151">WICG/fenced-frame#151</a>)
+
+     1. Set |instance|'s [=fenced frame config instance/has disabled untrusted network=] to true.
+
+     1. Wait on all nested fenced frames to disable network too.
+
+     Issue: Spec this waiting more formally.
+     (<a href="https://github.com/WICG/fenced-frame/issues/151">WICG/fenced-frame#151</a>)
```
domfarolino, line 1795: This is a HUGE TODO, so I'm not really comfortable landing this PR without those fundamental mechanisms spec'ed out a bit more rigorously, just FYI. "Wait on all nested fenced frames to disable network too" is pretty rough, so I'd like to have a good draft of how this is done in hand before landing this PR.

spec.bs
gtanzer: changed to "initially", w/e

spec.bs
blu25: I have created a draft PR https://github.com/WICG/fenced-frame/pull/176 to spec this more formally. That should hopefully unblock this PR.

VergeA: Hey y'all, I'm likely going to pick up the network revocation spec work from gtanzer, since he's transitioned to other work. This seems like the first major PR to have in place, and this is the last outstanding comment. Is the draft PR #176 sufficient scaffolding for disabling network in nested frames? If so, I'll do whatever else is needed to get this PR over the finish line.

domfarolino: I'd like to get https://github.com/WICG/fenced-frame/pull/176 just a tiny bit further along before merging this, if that's alright. It's in decent shape, I just want to be more confident in its direction before proceeding.

spec.bs
```diff
@@ -1738,6 +1754,52 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |fencedFrameNonce| be |instance|'s [=fenced frame config instance/partition nonce=].
+
+     1. Let |credentiallessNonce| be
+
+     Issue: the page credentialless nonce
+     (<a href="https://github.com/WICG/fenced-frame/issues/151">WICG/fenced-frame#151</a>)
```
blu25, line 1783: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176/files/7c89c4edea5d948e98efb3652f6a0b5b7c1c673d#r1748933801):

> I don't see how https://github.com/WICG/fenced-frame/issues/151 is related to credentialless. I expected to see an issue related to https://wicg.github.io/anonymous-iframe/#dom-window-credentialless integration or something.

spec.bs
```diff
@@ -1199,8 +1202,14 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
 
   : <dfn>embedder shared storage context</dfn>
   :: null, or a [=string=]
-  
+
   : <dfn>is ad component</dfn>
+  :: A [=boolean=]
+
+  : <dfn>can disable untrusted network</dfn>
```
blu25, line 1209: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176#discussion_r1748920494):

> Can you add a brief note about how this relates to the disableUntrustedNetwork() API? It seems pretty clear that it controls the behavior of that API in some way, but (a) clearly enunciating that would be great, and (b) describing what conditions make this false would be even better. (I.e., why would a fenced frame not be able to disable its own network?).

spec.bs
```diff
@@ -1738,6 +1754,52 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
+
+  1. If |instance|'s [=fenced frame config instance/can disable untrusted network=] is false, then
+     [=resolve=] |p| with {{undefined}} and return |p|.
+
+  1. If the [=relevant settings object=]'s [=environment settings object/origin=] and |instance|'s
+     [=fenced frame config instance/mapped url=]'s [=url/origin=] are not [=same origin=], then
+     [=resolve=] |p| with {{undefined}} and return |p|.
```
blu25, line 1772: Copied from [this comment](https://github.com/WICG/fenced-frame/pull/176#discussion_r1748934441):

> Let's very briefly describe in a note what this condition means. I think it means that the fenced frame has navigated away from the origin described by its original config (?), but in that case I would expect the config to be null anyways. Maybe I'm wrong about that part though.

spec.bs
blu25: I don't think this should actually be in the spec. From [the equivalent field in the impl](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_config.h;l=664;drc=57c74b90d4a6e506099c97022377d84f523b37cb):

> Currently true in all fenced frame configs, but set to false if loaded in a urn iframe

I think what we want instead is to just check if the relevant traversable navigable is a fenced navigable to see if we're in a fenced frame tree or not. The intention is to make sure that we can't call this from within a urn iframe. I'll remove this field and add the explicit non-fenced frame check in the disableUntrustedNetwork() definition.

spec.bs
blu25: I'm also expecting that to be the case. I think this check is redundant. If we want to be extra safe (and make it clear to implementers that only documents that are same-origin to the mapped URL can call this), we can assert they're same origin instead. But let me know if you think even that's unnecessary.

spec.bs
blu25: Created a new issue https://github.com/WICG/fenced-frame/issues/191 and linked that here.

spec.bs
```diff
@@ -1819,6 +1826,55 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
+
+  1. [=Assert=]: the [=relevant settings object=]'s [=environment settings object/origin=] and
+     |instance|'s [=fenced frame config instance/mapped url=]'s [=url/origin=] are [=same origin=].
+
+     Note: Only contexts that are [=same origin=] to the [=fenced frame config instance/mapped url=]
+     described in the [=fenced frame config instance=] are allowed to call this API.
```
domfarolino, line 1843: I'm happy the assert is "sensible" and passes, but I still don't quite understand it. Is it even possible for same-origin but different-from-the-original-navigation Documents to make it here? Or would the config be reset for those navigations?

spec.bs
domfarolino: Does that mean the equivalent field in the implementation can be removed, in favor of the the same check you describe? I'm asking because if the answer is no, then maybe there is some observable difference between what we'd end up spec'ing (without the member) and the implementation (which _does_ use the member, and deviates from the spec in a subtle way).

spec.bs
blu25: Actually, this should be a promise rejection with a TypeError (rather than an undefined resolution) to match what we do in impl. Since cross-origin subframes get access to `window.fence`, they are able to call this method.

It is possible if a same-origin child iframe calls `disableUntrustedNetwork()`, which is a case that we allow. It's also possible if the fenced frame performs a same-origin navigation before having the new navigated document call `disableUntrustedNetwork()`.

spec.bs
blu25: The main difference is that the impl accounts for the existence urn iframes, where it disallows disabling untrusted network across the board. Once urn iframes are removed, the check will strictly be a cross-origin check.

spec.bs
```diff
@@ -1881,6 +1888,53 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
 
+<div algorithm>
+  The <dfn method for=Fence>disableUntrustedNetwork()</dfn> method steps are:
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
```
domfarolino, line 1899: OK I am so sorry but I feel like I've asked when this can be null before, and I cannot for the life of me find the previous comments with the answer to this. This method is only accessible when instance is non-null, right? https://wicg.github.io/fenced-frame/#dom-window-fence. So is this condition needed? /cc @blu25 who might also know.

domfarolino (approved):
LGTM % small comment about null fenced frame config instance

domfarolino:
@blu25 I think this can be closed now that #169 has landed, which subsumed this, right?

blu25:
> @blu25 I think this can be closed now that #169 has landed, which subsumed this, right?

Yes, this can be closed.

PR #148, opened 2024-04-09
Spec "disable embedder-initiated nested fenced frame navigations after disabling untrusted network"

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/148.html" title="Last updated on Jan 21, 2025, 1:29 PM UTC (82f21b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/148/82a9f09...82f21b3.html" title="Last updated on Jan 21, 2025, 1:29 PM UTC (82f21b3)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -516,6 +516,16 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
 
   1. If |navigation url or urn| is failure, then return.
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null:
+     1. If |instance|'s [=fenced frame config instance/untrusted network status=] is not 
+        [=untrusted network status/enabled=], then return.
+
+     Note: Embedder-initiated navigations of fenced frame roots are not allowed after the
+     embedder's network has been disabled.
+
   1. Let |shared storage context| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/
      sharedStorageContext=].
 
--- a/spec.bs
+++ b/spec.bs
@@ -2120,6 +2120,9 @@ exempt specific URLs from network revocation.
 
 Issue: This will require a RFC to add a test-only function to the WPT web driver.
 (<a href="https://github.com/WICG/fenced-frame/issues/192">WICG/fenced-frame#192</a>)
+Once that web driver changes is made, existing Chromium-internal web platform tests for
+{{Fence/disableUntrustedNetwork()}} need to be upstreamed and linked here.
+(<a href="https://github.com/WICG/fenced-frame/issues/207">WICG/fenced-frame#207</a>)
 
 <div algorithm>
   To <dfn>revoke network for a partition nonce</dfn> using a [=fenced frame config
--- a/spec.bs
+++ b/spec.bs
@@ -513,8 +513,8 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
   1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
      [=browsing context/fenced frame config instance=].
 
-  1. If |instance| is not null, then if |instance|'s [=fenced frame config instance/untrusted
-     network status=] is not [=untrusted network status/enabled=], then return.
+  1. If |instance| is not null, and its [=fenced frame config instance/untrusted network status=]
+     is not [=untrusted network status/enabled=], then return.
 
   1. Let |navigation url or urn| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] if
      the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] is not null, and the given
--- a/spec.bs
+++ b/spec.bs
@@ -510,21 +510,17 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null, then if |instance|'s [=fenced frame config instance/untrusted
+     network status=] is not [=untrusted network status/enabled=], then return.
+
   1. Let |navigation url or urn| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] if
      the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] is not null, and the given
      {{FencedFrameConfig}}'s [=fencedframeconfig/urn=] otherwise.
 
   1. If |navigation url or urn| is failure, then return.
-
-  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
-     [=browsing context/fenced frame config instance=].
-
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -464,6 +464,16 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null:
+     1. If |instance|'s [=fenced frame config instance/has disabled untrusted network=] is true,
+        then return early.
```
domfarolino, line 472:
```suggestion
        then return.
```

spec.bs
```diff
@@ -464,6 +464,16 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null:
```
domfarolino, line 470: new line after this

spec.bs
```diff
@@ -464,6 +464,16 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
...
+     1. If |instance|'s [=fenced frame config instance/has disabled untrusted network=] is true,
+        then return early.
+
+     Note: Embedder-initiated navigations of fenced frame roots are not allowed after the
+     embedder's network has been disabled.
```
domfarolino, line 475: This is stating the obvious: https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts. It's better to describe _why_ this is important.

gtanzer: I'll just delete it then. The reason I left the note was to clarify that this was about the embedder's network being disabled (i.e. it's a nested fenced frame), but maybe this is obvious.

gtanzer:
@domfarolino Addressed comments but still waiting for other PR to unblock

domfarolino:
Just pinging @VergeA to make sure he's aware of this PR since I think he'll be taking it over once #146 lands.

VergeA:
Alright, this should be up to date now, although this change will be affected by the transition to "untrusted network status" as written in #176.

spec.bs
```diff
@@ -500,6 +500,14 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null:
+
+     1. If |instance|'s [=fenced frame config instance/has disabled untrusted network=] is true,
```
domfarolino, line 508: No need for this to be a whole substep. Just ", then ...."

domfarolino (commented):
I just want to double check something — you are intending for this PR to have no effect in a primary frame's Document, right? The only place this would have an effect in is inside a fenced frame which is trying to set the `config` attribute on a nested fenced frame. If that top-level fenced frame has a config instance with disabled network, then we would fail to navigate the nested fenced frame as expected. But if we're running the `config` setter from a primary main frame's script, the config instance will never be null and therefore this PR won't have an effect in that case. That's the intention, right?

Also, if there are any tests you can link to from the PR, please do so. That would help a lot.

VergeA:
Yes, your understanding is correct. See the corresponding Chromium CL [here](https://chromium-review.googlesource.com/c/chromium/src/+/5201401) which implements that behavior explicitly.

domfarolino:
> Is there a way you recommend linking internal tests that doesn't clash with the existing WPT linking format? Or should they be left out if they aren't upstreamed?

The right way would be to file whatever bugs are necessary against web driver and WPT's test harness infrastructure so that some sort of engine-agnostic API can be implemented to expose the right behavior. I recall being able to file RFCs about this somewhere, maybe you could try and track this down. In the meantime, I would file an issue against our repository describing the task of doing exactly that, and linking to it from our spec where this is implemented, since it's a pretty big issue — it means no other browser that implements fenced frames can assert the correct behavior here.

VergeA:
@blu25 has already filed an issue against our repository to get a new Webdriver API implemented: https://github.com/WICG/fenced-frame/issues/192.

I'll go ahead and add this issue to the spec.

domfarolino:
Sure, but starting that process with an actual RFC would be probably more useful to get the ball rolling on it.

VergeA:
> Sure, but starting that process with an actual RFC would be probably more useful to get the ball rolling on it.

To clarify, is starting that process a blocker for this PR? This RFC work came up again in a Chromium [code review](https://chromium-review.googlesource.com/c/chromium/src/+/6163876/comment/5a7de8f3_b3b077d0/) recently, so I'm circling back here as well. We're planning to have the Webdriver/RFC work done by the end of this year.

domfarolino:
No as long as we have an issue filed (and maybe a TODO in this PR linking to that issue) to finally add the tests to the spec once they are upstreamed into `external/wpt`.

VergeA:
I created a sub-issue of #192 to link the tests in the spec: #207.

I've also linked to it in the spec in my latest commit.

VergeA:
Build appears to be broken, will re-request review one I fix it.

VergeA:
This feature branch was added directly to a repo instead of the fork, and was actually pretty behind. I rebased locally, force-pushed, and used the new `untrusted network status` concept to re-implement the logic here.

spec.bs
```diff
@@ -510,6 +510,12 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is not null, then if |instance|'s [=fenced frame config instance/untrusted
```
domfarolino, line 516:
```suggestion
  1. If |instance| is not null, and it's [=fenced frame config instance/untrusted
```

(Will likely need rewrapping)

domfarolino (commented):
So this generally looks good, and it's simple. But I have a question about https://github.com/WICG/fenced-frame/pull/148#pullrequestreview-2466082757 and your response in https://github.com/WICG/fenced-frame/pull/148#issuecomment-2511908986. Why is it OK to allow primary-frame-initiated navigations to take place on top-level fenced frame elements whose network is disabled?

Or would those navigations actually fail somewhere further down the pipeline, because their network has been revoked? And if that's the case (i.e., navigations would organically fail somewhere in the loading pipeline because their network was revoked), then why do we need this PR?

VergeA:
When the top-level fenced frame navigates to a new document, then the network revocation state will reset and it will regain network access. This would be an issue if any child frame could retain network access. However, because the top-level fenced frame navigated, all of its child frames will be unloaded, and the top-level fenced frame will become a "clean slate" with a new document. Additionally, the navigation URL is provided by the embedder in the `FencedFrameConfig`, so any cross-site data that is accessed inside of a fenced frame with network revoked cannot be used to decorate that URL.

spec.bs

domfarolino (approved):

PR #149, opened 2024-04-09
Spec partition nonce functionality

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/149.html" title="Last updated on Mar 10, 2025, 6:22 PM UTC (e117ec2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/149/fa4ec2c...e117ec2.html" title="Last updated on Mar 10, 2025, 6:22 PM UTC (e117ec2)">Diff</a>


spec.bs
```diff
@@ -3291,6 +3304,53 @@ at the expense of some utility.
...
+frames.
+
+Add the following algorithm:
+
+<div algorithm>
+  To <dfn for="browsing context">compute the effective partition nonce</dfn> given a
+  boolean |isCredentialless| and [=fenced frame config instance/partition nonce=] or null
+  |newFencedFrameNonce|:
+
+  1. If |credentialless| is true, return |this|'s [=top-level browsing context=]'s
```
shivanigithub, line 3324: credentialless => isCredentialless

gtanzer: I can't really change this, since it corresponds to an exposed IDL attribute.
https://wicg.github.io/anonymous-iframe/#spec-window-attribute

gtanzer: I mean it's a local variable, but it would be a bit weird to call it something else when it corresponds exactly to this preexisting field, and the same name is used throughout the Iframe credentialless spec

domfarolino: Where is this local variable defined? Sorry I don't think I'm getting it.

spec.bs
```diff
@@ -3291,6 +3304,53 @@ at the expense of some utility.
...
+
+Note: The [[!IFRAME-CREDENTIALLESS]] specification defines a new object, the
+[=page credentialless nonce=]. At a high level, the
+[=fenced frame config instance/partition nonce=] serves the same purpose as the
+[=page credentialless nonce=] (partitioning storage and network), but scoped to each fenced
+frame rather than to each credentialless iframe. The
+[=fenced frame config instance/partition nonce=] is also used to revoke network access in fenced
+frames.
+
+Add the following algorithm:
```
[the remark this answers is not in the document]
gtanzer: Added "Perform the following monkeypatches to the [[!IFRAME-CREDENTIALLESS]] spec."

spec.bs
gtanzer: Oh, I misread this. I didn't mean to name the variable `isCredentialless` above in the declaration. The iframe credentialless spec refers to this kind of variable as `credentialless` throughout, so it would be inconsistent naming to use `isCredentialless` here.

spec.bs
```diff
@@ -3291,6 +3304,53 @@ at the expense of some utility.
...
+
+  1. Let |instance| be |this|'s [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, return null.
+
+  1. Return |instance|'s [=fenced frame config instance/partition nonce=].
+</div>
+
+<div algorithm="process a navigate fetch">
+  Modify the step added to [=process a navigate fetch=] to read:
```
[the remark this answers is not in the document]
gtanzer: Fixed

spec.bs
```diff
@@ -3291,6 +3304,53 @@ at the expense of some utility.
...
+
+  13. Let |partitionNonce| be the result of [=computing the effective partition nonce=] on
+      |browsingContext| given |credentialless| and |sourceSnapshotParams|'s
+      [=source snapshot params/target fenced frame config=].
+</div>
+
+<div algorithm="initialize the document object">
+  Modify the step added to [=initialize the document object=] to read:
```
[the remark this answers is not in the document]
gtanzer: Fixed

gtanzer:
@domfarolino I think all the comments are addressed now

VergeA:
Hey y'all, this seems important to have for completeness of the network revocation spec, so I'm going to take it over from gtanzer.

Looks like all outstanding comments have been addressed, but wanted to request a re-review for any additional feedback. If there's additional changes necessary, I'll address them before we merge.

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+  |newFencedFrameNonce|:
+
+  1. If |credentialless| is true, return |this|'s [=top-level browsing context=]'s
+     [=page credentialless nonce=].
+
+  1. If |newFencedFrameNonce| is not null, return |newFencedFrameNonce|.
+
+  1. Let |instance| be |this|'s [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, return null.
```
domfarolino, line 3337:
```suggestion
  1. If |instance| is null, then return null.
```

Per convention.

domfarolino:
I think most of this could use rewrapping. I think it rewraps too aggressively, avoiding all spaces in linked-to terms.

spec.bs
VergeA: Yes, I believe the "credentialless" variable needs to be declared. Reasoning:

- The "process a navigate fetch" algorithm below already has a "credentialless" variable 
- The "initialize a document object" algorithm below also already has a "credentialless" variable

Without it, we don't know whether the partition nonce should come from the credentialless iframe or the fenced frame.

spec.bs
```diff
@@ -254,6 +270,9 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
 spec: turtledove; urlPrefix: https://wicg.github.io/turtledove/
   type: dfn
     text: construct a pending fenced frame config; url: construct-a-pending-fenced-frame-config
+spec: iframe-credentialless; urlPrefix: https://wicg.github.io/anonymous-iframe/
+  type: dfn
+    text: credentialless; url: navigation-params-credentialless
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+[=page credentialless nonce=] (partitioning storage and network), but scoped to each fenced
+frame rather than to each credentialless iframe. The
+[=fenced frame config instance/partition nonce=] is also used to revoke network access in fenced
+frames. Perform the following monkeypatches to the [[!IFRAME-CREDENTIALLESS]] spec.
+
+Add the following algorithm:
+
+<div algorithm>
+  To <dfn for="browsing context">compute the effective partition nonce</dfn> given a
+  boolean |credentialless| and [=fenced frame config instance/partition nonce=] or null
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
   </wpt>
 </div>
 
+<h3 id=credentialless-monkeypatch>Iframe credentialless</h3>
+
+The [[!IFRAME-CREDENTIALLESS]] specification defines a new object, the
+[=page credentialless nonce=]. At a high level, the
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+  1. If |credentialless| is true, return |this|'s [=top-level browsing context=]'s
+     [=page credentialless nonce=].
+
+  1. If |newFencedFrameNonce| is not null, return |newFencedFrameNonce|.
+
+  1. Let |instance| be |this|'s [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, return null.
+
+  1. Return |instance|'s [=fenced frame config instance/partition nonce=].
```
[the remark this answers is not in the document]
VergeA: Done, except we should return the nonce if |instance| is _not_ null (since otherwise, there's nothing to return)

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+frames. Perform the following monkeypatches to the [[!IFRAME-CREDENTIALLESS]] spec.
+
+Add the following algorithm:
+
+<div algorithm>
+  To <dfn for="browsing context">compute the effective partition nonce</dfn> given a
+  boolean |credentialless| and [=fenced frame config instance/partition nonce=] or null
+  |newFencedFrameNonce|:
+
+  1. If |credentialless| is true, return |this|'s [=top-level browsing context=]'s
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+      <var ignore>sourceSnapshotParams</var>'s
+      [=source snapshot params/target fenced frame config=].
+</div>
+
+<div algorithm="initialize the document object">
+  Modify the step added to [=initialize the document object=] to read:
+
+  6.9. Let <var ignore>partitionNonce</var> be the result of
+       [=browsing context/computing the effective partition nonce=] on
+       <var ignore>browsingContext</var> given |navigationParams|'s [=credentialless=] and null if
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -3291,6 +3310,56 @@ at the expense of some utility.
...
+
+  1. Let |instance| be |this|'s [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, return null.
+
+  1. Return |instance|'s [=fenced frame config instance/partition nonce=].
+</div>
+
+<div algorithm="process a navigate fetch">
+  Modify the step added to [=process a navigate fetch=] to read:
```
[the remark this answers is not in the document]

spec.bs
VergeA: This line was removed when consolidating with the line below it, per your previous comment.

VergeA:
Addressed all the outstanding comments, but still need to go back through and rewrap everything.

VergeA:
Alright, this should be all rewrapped now.

spec.bs
```diff
@@ -3990,6 +4007,55 @@ at the expense of some utility.
   </wpt>
 </div>
 
+<h3 id=credentialless-monkeypatch>Iframe credentialless</h3>
+
+The [[!IFRAME-CREDENTIALLESS]] specification defines a new object, the [=page credentialless 
+nonce=]. At a high level, the [=fenced frame config instance/partition nonce=] serves the same
+purpose as the [=page credentialless nonce=] (partitioning storage and network), but scoped to each
+fenced frame rather than to each credentialless iframe. The [=fenced frame config instance/
```
domfarolino, line 4015: Is page credentialless nonce really scoped to each credentialless iframe? The definition you link to seems to be tied to a top-level browsing context, which spans many iframes.

spec.bs
```diff
@@ -3990,6 +4007,55 @@ at the expense of some utility.
...
+partition nonce=] is also used to revoke network access in fenced frames. Perform the following
```
domfarolino, line 4016:
```suggestion
partition nonce=] is also used to revoke network access in fenced frames. Apply the following
```

spec.bs
```diff
@@ -3990,6 +4007,55 @@ at the expense of some utility.
...
+
+Add the following algorithm:
+
+<div algorithm>
+  To <dfn for="browsing context">compute the effective partition nonce</dfn> given a boolean 
+  |credentialless| and [=fenced frame config instance/partition nonce=]-or-null
+  |newFencedFrameNonce|:
+
+  1. If |credentialless| is true, return the associated [=browsing context=]'s [=top-level
+     browsing context=]'s [=page credentialless nonce=].
```
domfarolino, line 4027: Note that the usage of "top-level browsing context" here will be fenced — is that intentional? That's because this specification does not override HTML's definition of top-level browsing context, which just iteratively reaches upward via the "parent" property, never the "unfenced parent" property.

If you want the unfenced version, you'll need to grab a [navigable's top-level traversable](https://wicg.github.io/fenced-frame/#navigable-top-level-traversable) and get its browsing context and grab its page credentialless nonce.

I just wanted to raise this in case it was unexpected. WDYT?

It seems like this algorithm prefers:
 1. A "fenced" page redentialless nonce, when we're in a credentialless iframe
 2. A "new" fenced frame nonce, when it exists
 3. The "old" navigated-away-from browsing context's fenced frame partition nonce, when we're in a fenced frame whose browsing context is apparently being navigated away from. (Is that right?)

Actually (3) above is a little confusing or surprising to me. Below on `L4031` we're always grabbing the current browsing context's fenced frame config instance's nonce, even when said browsing context is I guess being navigated away from? Am I missing something?

spec.bs
```diff
@@ -3990,6 +4007,55 @@ at the expense of some utility.
...
+
+  1. Return |instance|'s [=fenced frame config instance/partition nonce=] if |instance| is not
+     null, and return null otherwise.
+</div>
+
+<div algorithm="process a navigate fetch">
+  Modify the step added to <a href=https://wicg.github.io/anonymous-iframe/#spec-navigation-partition-nonce>
+  process a navigate fetch</a> to read:
+
+  13. Let <var ignore>partitionNonce</var> be the result of [=browsing context/computing the
```
domfarolino, line 4042: Just to triple check here — the partitionNonce returned from the "computing" algorithm can be null. Can all of the consumers of this (presumably in the iframe credentialless spec) consume null in this case? Just want to make sure we're not in a position where we're passing null around to things that only accept/expect a populated nonce.

VergeA: The credentialless iframe spec explicitly defines a partition nonce as "[an identifier or null](https://wicg.github.io/anonymous-iframe/#spec-environment-partition-nonce)" in the environment.

In the [algorithms](https://wicg.github.io/anonymous-iframe/#spec-navigation-partition-nonce) we're changing here from the credentialless iframe spec, the partition nonce is already allowed to be null if the page credentailless nonce is missing.

spec.bs
VergeA: Done.

spec.bs
VergeA: Yeah I agree this is just worded wrong; every credentialless iframe "has" a nonce, which I think was the original intention here, but the value of that nonce is scoped to the top-level document. I updated the wording here to indicate as such.

spec.bs
VergeA: Ok I think I have figured (3)

For the "process a navigate fetch" algorithm, we always want to use the "old" partition nonce, so we can determine if the navigation is originating from a fenced frame, and if that navigation should be revoked due to disabled network access. 

For "initializing the document" object, if a new fenced frame config instance was set as the result of an embedder-initiated navigation, we want to use the nonce from that one. Otherwise, the navigation was content-initiated or a subframe navigation, so we should use the nonce available from the current browsing context instead.

spec.bs
```diff
@@ -4315,6 +4332,61 @@ at the expense of some utility.
...
+  access the correct partition nonce via the [=fenced frame config instance=]'s
+  [=fenced frame config instance/partition nonce=].
+</div>
+
+<div algorithm="initialize the document object">
+  Modify the step added to <a href=https://wicg.github.io/anonymous-iframe/#spec-window-partition-nonce>
+  initialize the document object</a> to read:
+
+  6.9. Let <var ignore>partitionNonce</var> be the result of [=browsing context/computing the
+       effective partition nonce=] on <var ignore>browsingContext</var> given |navigationParams|'s
```
blu25, line 4384: To confirm, this is what eventually ends up being put into the new environment object when it's created, right? The changes made to the fetch patches (where null is unconditionally passed in for the fenced frame nonce when computing the effective partition nonce) will have no effect on the environment that's ultimately created for the document?

blu25: Ack, then I agree that this should be fine.

spec.bs
```diff
@@ -4315,6 +4332,63 @@ at the expense of some utility.
...
 
+<h3 id=credentialless-monkeypatch>Iframe credentialless</h3>
+
+*This first introductory paragraph is non-normative.*
+
+The [[!IFRAME-CREDENTIALLESS]] specification defines a new object, the [=page credentialless
+nonce=]. At a high level, the [=fenced frame config instance/partition nonce=] serves the same
+purpose as the [=page credentialless nonce=] (partitioning storage and network). Howevever, it is
+scoped to each fenced frame, whereas the [=page credentialless nonce=] is scoped to the top-level
+document and shared by all descendent credentialless iframes. The [=fenced frame config instance/
```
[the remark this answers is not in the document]
VergeA: I fleshed out this paragraph a bit:

1. I clarified that the credentialless nonce is scoped to roughly the traversable navigable's Window.

2. I clarified that the child's nonce always takes precedence over the parent's when credentialless iframes and fenced frames compose.

spec.bs
```diff
@@ -4315,6 +4332,63 @@ at the expense of some utility.
...
+  1. Let |instance| be the associated [=browsing context=]'s [=browsing context/fenced frame
+     config instance=].
+
+  1. Return |instance|'s [=fenced frame config instance/partition nonce=] if |instance| is not
+     null, and return null otherwise.
+</div>
+
+<div algorithm="process a navigate fetch">
+  Modify the step added to <a href=https://wicg.github.io/anonymous-iframe/#spec-navigation-partition-nonce>
+  process a navigate fetch</a> to read:
```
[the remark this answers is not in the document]
VergeA: I believe my latest patch should be doing this correctly now, as it patches "create navigation params by fetching," which replaced "process a navigate fetch." PTAL!

spec.bs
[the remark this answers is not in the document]
domfarolino: Looks great, thanks!

spec.bs
domfarolino: > But the environment for that request isn't inserted into the document itself. That happens during "initializing the document object" later.

I might be totally forgetting how this works, or am missing something in the current spec, but I believe:

1. [Here is where "create navigation params by fetching" creates the reserved environment](https://html.spec.whatwg.org/multipage/browsing-the-web.html#populating-a-session-history-entry:concept-request-reserved-client-6)
2. [Here is where where carry it over from the request to the navigation params](https://html.spec.whatwg.org/multipage/browsing-the-web.html#populating-a-session-history-entry:navigation-params-reserved-environment-4)
3. [Here is where the shared document creation infra](https://html.spec.whatwg.org/multipage/document-lifecycle.html#shared-document-creation-infrastructure:navigation-params-reserved-environment-2) assigns the carried-over reserved environment over to the new ESO that the Document loads into

Given that, I don't think we throw out the reserved client/environment like you suspected we did. Does the change the calculus here at all? Maybe we can get away without monkeypatching "create navigation params by fetching" altogether?

VergeA: Two things:

1. You're right, the full environment gets carried over into the navigation params and used in the new document. I tried looking through all the linked references to "request's reserved client" in the spec to determine where it was being used, but I must have missed this one.

2. There is still a new FencedFrameConfig that is also assigned in the navigation params, and the environment of the document we create should be using the nonce from that one. As a result, I am not sure anything changes that much here. If the environment is used to compute the network partition key while fetching, then we still want to use the "old" nonce to start. After that fetch completes and the document inherits it, that original environment is just one input into the "[set up a window environment settings object](https://wicg.github.io/anonymous-iframe/#spec-window-partition-nonce)," and the new partition nonce is another.

domfarolino: Thanks, this makes perfect sense to me.

domfarolino (approved):
Thanks a lot!

PR #150, opened 2024-04-10
Udpate spec for processing attribution eligible response

linnan-github:
`Process an attribution eligible response` interface was changed in  https://github.com/WICG/attribution-reporting-api/pull/1238.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/fenced-frame/pull/150.html" title="Last updated on Aug 14, 2024, 2:52 PM UTC (bef63a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/150/c799ffe...linnan-github:bef63a5.html" title="Last updated on Aug 14, 2024, 2:52 PM UTC (bef63a5)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -894,9 +894,8 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
 
         1. Set |processResponse| to these steps given a [=response=] |response|:
            1. Let |fenced| be true.
-           1. Let |triggerVerificationMetadata| be null.
            1. Run [=process an attribution eligible response=] with |event|'s
-              [=destination enum event/attributionReportingContextOrigin=], |attributionReportingEligibility|, |fenced|, |triggerVerificationMetadata|, and |response|.
+              [=destination enum event/attributionReportingContextOrigin=], |attributionReportingEligibility|, |fenced|, and |response|.
 
         1. Set |useParallelQueue| to true.
 
--- a/spec.bs
+++ b/spec.bs
@@ -252,6 +252,8 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
       text: event-source; url: eligibility-event-source
       text: navigation-source; url: eligibility-navigation-source
       text: unset; url: eligibility-unset
+    for: request
+      text: trigger verification metadata; url: request-trigger-verification-metadata
 spec: turtledove; urlPrefix: https://wicg.github.io/turtledove/
   type: dfn
     text: construct a pending fenced frame config; url: construct-a-pending-fenced-frame-config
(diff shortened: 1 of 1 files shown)


linnan-github:
@domfarolino Could you PTAL, thanks.

spec.bs
```diff
@@ -888,9 +890,9 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
         1. Otherwise, set |attributionReportingEligibility| to "<code>[=eligibility/event-source=]</code>".
 
         1. Set |processResponse| to these steps given a [=response=] |response|:
-
+           1. Let |fenced| be true.
            1. Run [=process an attribution eligible response=] with |event|'s
-              [=destination enum event/attributionReportingContextOrigin=], |attributionReportingEligibility|, and |response|.
+              [=destination enum event/attributionReportingContextOrigin=], |attributionReportingEligibility|, |fenced|, |request|'s [=request/trigger verification metadata=], and |response|.
```
domfarolino, line 895: Can we just pass in true? Looks like we don't need the extra `|fenced|` variable?

domfarolino, line 895: Where is the `|request|` variable here? Is it in scope? Do we have access to it?

linnan-github: yeah we can, but just thought it's clearer what this parameter is. Wdyt?

linnan-github: I'm not sure what's the best way to spec this. `|request|` is defined a few lines below and this is the function to process the response of the request.

domfarolino: I see, so it is basically self-documenting code like `/*parameter_foo=*/true` would be. Hmm, OK I'm fine with it.

domfarolino, line 895: The "process" algorithm accepts only a `trigger verification metadata`, but `[=request/trigger verification metadata=]` could be null. Can you handle this appropriately either by catching this and dipping out early, or asserting that it is never null on these kinds of requests perhaps? (It's not clear to me which is right).

domfarolino: At first I considered recommending you to define request higher up (but keep it empty), so that we could at least reference it here. The implicit behavior here would be that by the time _processBody_ finally runs, the `|request|` it accesses is not the old stale empty one that is in scope, but is instead the version of `|request|` that has gone through the Fetch infra and had its `trigger verification metadata` populated etc.

But that doesn't feel very rigorous to me, so I guess I would rather see the attribution reporting spec monkeypatch Fetch in way that transfers over the `trigger verification metadata` bit from the request to the response, if you really need to access it in the response during _processBody_. That's probably the cleanest & most rigorous way. It kind of looks like that's how we do it in the implementation but I lost track of tracing all of the associated callbacks back from this point: https://source.chromium.org/chromium/chromium/src/+/main:services/network/attribution/attribution_request_helper.cc;l=315;drc=63635fff4bce239256a42d5f1a28fa3d0934bf35.

linnan-github: Thanks Dom. That sounds reasonable. cc @agarant.

In the meantime, since fenced frame beacons are only eligible for sources and therefore trigger verification metadata is irrelevant. Updated to hardcode it to null instead. Thanks.

linnan-github: https://github.com/WICG/attribution-reporting-api/pull/1271 to address this in ARA spec, thanks.

agarant: This is indeed how the implementation works, I can look into making this change.

domfarolino: Thanks. I'd be most comfortable landing this after that ARA spec change is fixed, or possibly by putting a prominent TODO around the hardcoding that you added (something [like this](https://wicg.github.io/fenced-frame/#issue-0ffad658)). I prefer the former though, and keeping this thread up to date on the ARA spec progress there.

linnan-github: Hi @domfarolino , we've removed trigger verification from ARA spec. Could you please review this PR again, thanks!

linnan-github:
@domfarolino Could you PTAL, thanks!

domfarolino (approved):

linnan-github:
@domfarolino I don't have write access, could you please merge the change, thanks!

PR #152, opened 2024-04-12
Spec cross-origin reportEvent() support

blu25:
Reporting beacons can now be sent with `window.fence.reportEvent()` from documents that are cross origin to a fenced frame config's mapped URL. To do this, there must be opt-in from both the document created with the FencedFrameConfig as well as the cross-origin document that wants to send the beacon. The document created with the FencedFrameConfig opts in with a new "Allow-Cross-Origin-Event-Reporting=true" response header. The cross-origin document opts in by calling `reportEvent()` with the `crossOriginExposed=true` parameter.

This PR updates the spec to match that behavior. More specifically:
- Updates the FenceEvent IDL to match the [current .idl file](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/fence_event.idl) on the codebase.
- Adds a way to store a new "Allow-Cross-Origin-Event-Reporting" opt-in header in the fenced frame config. This header is set in the document whose browsing context houses the fenced frame config. It makes most sense to store this in the fenced frame config directly, since it will be consulting that when calling reportEvent() anyway, and cross-origin iframes still have access to the fenced frame config instance object.
- Modifies `reportEvent()` to support being called from a document that is cross-origin to the fenced frame config's mapped URL.
  - The private aggregation path has no changes, as it still doesn't have cross-origin support.
  - The destination enum/URL path adds a carve-out that allows it to continue if it's cross-origin but also has both header opt-in for the document that owns the fenced frame config + `crossOriginExposed=true` opt-in from the document that's sending the beacon.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/152.html" title="Last updated on Apr 30, 2024, 3:34 PM UTC (c899272)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/152/b1c7abe...c899272.html" title="Last updated on Apr 30, 2024, 3:34 PM UTC (c899272)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1145,10 +1145,10 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
   :: null, or a [=string=]
   
   : <dfn>is ad component</dfn>
-  :: A [=boolean=]. Defaulting to false.
+  :: A [=boolean=], initially false.
 
   : <dfn>cross-origin reporting allowed</dfn>
-  :: A [=boolean=]. Defaulting to false.
+  :: A [=boolean=], initially false.
 </dl>
   
   Note: When true, this [=fenced frame config=] reprsents an ad component. An ad component can be
@@ -1204,7 +1204,7 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
   :: A [=boolean=], initially false.
 
   : <dfn>cross-origin reporting allowed</dfn>
-  :: A [=boolean=]. Defaulting to false.
+  :: A [=boolean=], initially false.
 </dl>
 
 <div algorithm>
@@ -1499,10 +1499,11 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   };
 
   dictionary FenceEvent {
-    // This dictionary has two mutually exclusive modes:
-
-    // When reporting to a preregistered destination (specified by enum), the
-    // following properties are used:
+    // This dictionary has two mutually exclusive modes that aren't represented as
+    // distinct IDL types due to distinguishability issues:
+    //
+    // When reporting to a preregistered destination (specified by enum), the following
+    // properties are used:
     DOMString eventType;
     DOMString eventData;
     sequence&lt;FenceReportingDestination&gt; destination;
@@ -1517,12 +1518,10 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
     // When setting event data to be used later in an automatic beacon, the
     // following properties are used:
-    // Determines if the beacon data will be used for only the next automatic
-    // beacon event, or if it will be reused for all subsequent automatic beacons.
     boolean once = false;
 
-    // When reporting to a custom destination URL (with substitution of macros
-    // defined by the buyer), the following property is used:
+    // When reporting to a custom destination URL (with substitution of macros defined by
+    // the Protected Audience buyer), the following property is used:
     USVString destinationURL;
   };
 
@@ -2693,11 +2692,6 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
       1. Set |config|'s [=fenced frame config/embedder shared storage context=] to
          |sharedStorageContext|.
 
-      1. If |response| is non-null, set |config|'s [=fenced frame config/cross-origin reporting
-         allowed=] to the result of running [=header list/get a structured field value=] on
-         |response|'s [=response/header list=] given
-         "<code>Allow-Cross-Origin-Event-Reporting</code>" and "`item`".
-
       1. Set <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/target fenced frame
          config=] to |config|.
 
@@ -2784,6 +2778,12 @@ content in the <{fencedframe}> or its embedder, exactly one of two things will h
        Note: This is because embedder-initiated navigations (indicated by |navigationParams|'s
        [=navigation params/fenced frame config instance=] being non-null) always cause a
        [[#bcg-swap]].
+
+    1. Set |navigationParams|'s [=navigation params/fenced frame config instance=]'s [=fenced frame
+       config/cross-origin reporting allowed=] to the result of running [=header list/get a
+       structured field value=] on |navigationParams|'s [=navigation params/response=]'s
+       [=response/header list=] given "<code>Allow-Cross-Origin-Event-Reporting</code>" and
+       "`item`".
 
     1. Set |browsingContext|'s [=browsing context/fenced frame config instance=] to
        |navigationParams|'s [=navigation params/fenced frame config instance=].
--- a/spec.bs
+++ b/spec.bs
@@ -1146,6 +1146,9 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
   
   : <dfn>is ad component</dfn>
   :: A [=boolean=]. Defaulting to false.
+
+  : <dfn>cross-origin reporting allowed</dfn>
+  :: A [=boolean=]. Defaulting to false.
 </dl>
   
   Note: When true, this [=fenced frame config=] reprsents an ad component. An ad component can be
@@ -1199,6 +1202,9 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
   
   : <dfn>is ad component</dfn>
   :: A [=boolean=], initially false.
+
+  : <dfn>cross-origin reporting allowed</dfn>
+  :: A [=boolean=]. Defaulting to false.
 </dl>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


domfarolino:
> `window.fence.reportEvent()` will now support being sent from documents

Methods cannot be sent. Do you mean exposed? Or am I missing something?

domfarolino:
> The same-origin document opts in with a new "Allow-Cross-Origin-Event-Reporting=true" response header. The cross-origin document opts in by calling reportEvent() with the crossOriginExposed=true parameter.

Can you link to the rationale for why these are different?

spec.bs
```diff
@@ -1146,6 +1146,9 @@ A <dfn export>fenced frame config</dfn> is a [=struct=] with the following [=str
   
   : <dfn>is ad component</dfn>
   :: A [=boolean=]. Defaulting to false.
+
+  : <dfn>cross-origin reporting allowed</dfn>
+  :: A [=boolean=]. Defaulting to false.
```
domfarolino, line 1151:
```suggestion
  :: A [=boolean=], initially false.
```

Just like the `[=fenced frame config instance/is ad component=]` dfn, which is what the `[=fenced frame config/is ad component=] definition should say too.

spec.bs
```diff
@@ -1199,6 +1202,9 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
   
   : <dfn>is ad component</dfn>
   :: A [=boolean=], initially false.
+
+  : <dfn>cross-origin reporting allowed</dfn>
+  :: A [=boolean=]. Defaulting to false.
```
domfarolino, line 1207: ditto

spec.bs
```diff
@@ -1490,22 +1499,30 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   };
 
   dictionary FenceEvent {
-    // This dictionary has two mutually exclusive modes that aren't represented as
-    // distinct IDL types due to distinguishability issues:
-    //
-    // When reporting to a preregistered destination (specified by enum), the following
-    // properties are used:
+    // This dictionary has two mutually exclusive modes:
+
```
domfarolino, line 1503: Why remove the line about distinguishability? These kinds of changes can come in another PR I think.

spec.bs
```diff
@@ -1490,22 +1499,30 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+    // URL of the fenced frame config that loaded this frame tree.
+    // Note that automatic beacon data can only be set from documents that are
+    // same-origin to the fenced frame config's mapped URL, so this effectively
+    // opts in the data to being used in a cross-origin subframe.
+    boolean crossOriginExposed = false;
+
     // When setting event data to be used later in an automatic beacon, the
     // following properties are used:
+    // Determines if the beacon data will be used for only the next automatic
+    // beacon event, or if it will be reused for all subsequent automatic beacons.
```
domfarolino, line 1521: I would leave changes like this out of this PR and include them in another.

spec.bs
```diff
@@ -1490,22 +1499,30 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
     boolean once = false;
-    boolean crossOriginExposed = false;
 
-    // When reporting to a custom destination URL (with substitution of macros defined by
-    // the Protected Audience buyer), the following property is used:
+    // When reporting to a custom destination URL (with substitution of macros
+    // defined by the buyer), the following property is used:
```
domfarolino, line 1525: same?

spec.bs
```diff
@@ -1529,19 +1546,31 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
-     [=fenced frame config instance/mapped url=]'s [=url/origin=] are not [=same origin=], then
-     return.
-
   1. If |instance|'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
 
-  1. If |event| is a {{DOMString}}, run [=report a private aggregation event=] using |instance|'s
-     [=fenced frame config instance/fenced frame reporter=] with |event|.
+  1. If |event| is a {{DOMString}}:
+
+     1. If the [=relevant settings object=]'s [=environment settings object/origin=] and
```
domfarolino, line 1553: Please see the above usage of RSO and match it. You cannot pull it own of nowhere.

spec.bs
```diff
@@ -1529,19 +1546,31 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+     1. Run [=report a private aggregation event=] using |instance|'s [=fenced frame config
+        instance/fenced frame reporter=] with |event|.
 
   1. If |event| is a {{FenceEvent}}:
 
      1. If |event|'s {{FenceEvent/eventType}} [=string/starts with=] "`reserved.`", then return.
 
+     1. If all of the following conditions are true:
+
+        * the [=relevant settings object=]'s [=environment settings object/origin=] and
```
domfarolino, line 1566: same

spec.bs
```diff
@@ -1529,19 +1546,31 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
-
   1. If |instance|'s [=fenced frame config instance/fenced frame reporter=] is null, then return.
 
-  1. If |event| is a {{DOMString}}, run [=report a private aggregation event=] using |instance|'s
-     [=fenced frame config instance/fenced frame reporter=] with |event|.
+  1. If |event| is a {{DOMString}}:
+
+     1. If the [=relevant settings object=]'s [=environment settings object/origin=] and
+        |instance|'s [=fenced frame config instance/mapped url=]'s [=url/origin=] are not [=same
+        origin=], then return.
```
domfarolino, line 1555: Just to be clear, we don't need to consider the other conditions that we do on L1566, right? About the `crossOriginExposed` bit being true and so on, to allow this report to go through even in the cross-origin case?

spec.bs
```diff
@@ -2653,6 +2693,11 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
       1. Set |config|'s [=fenced frame config/embedder shared storage context=] to
          |sharedStorageContext|.
 
+      1. If |response| is non-null, set |config|'s [=fenced frame config/cross-origin reporting
```
domfarolino, line 2696: This is before we actually have the network response though. This `|response|` is only non-null in the x-mixed-replace and object/embed case, right? https://html.spec.whatwg.org/C#navigation-response.

I think you want this logic elsewhere.

spec.bs
blu25: I didn't modify this as part of the cross-origin reportEvent() CL so I'll leave this out and handle it in the IDL follow up.

spec.bs
blu25: Correct. In this path, `event` is a `DOMString`, not a `FenceEvent`, so there's no `crossOriginExposed` bit that can be checked.

blu25:
> > `window.fence.reportEvent()` will now support being sent from documents
> 
> Methods cannot be sent. Do you mean exposed? Or am I missing something?

I mean exposed. The wording should be "Beacons can be sent with `window.fence.reportEvent()` from documents that are cross-origin..."

> > The same-origin document opts in with a new "Allow-Cross-Origin-Event-Reporting=true" response header. The cross-origin document opts in by calling reportEvent() with the crossOriginExposed=true parameter.
> 
> ~Can you link to the rationale for why these are different?~
> 
> **Edit**: I think I misunderstood. Is "the same-origin document" just the fenced frame document? That is, the document created by the FencedFrameConfig, i.e., the top-level document in a `<fencedframe>` element?

Yes. Same-origin iframes embedded in fenced frames have no control over the opt-in process. I'll update the wording to match that.

spec.bs
blu25: I've moved this to the [create and initialize a Document object](https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object) patches, since that's where the automatic beacon opt-in piece already lives.

spec.bs
domfarolino: Oh yeah, of course. But do we also not need to concern ourself with `|instance|'s [=fenced frame config instance/cross-origin reporting allowed=]`?

blu25: No. `DOMString` events don't support cross-origin beacons.

The only way for this feature to be safely supported is if we can get opt-in from both the document loaded from the fenced frame config as well as the document sending the `reportEvent()` beacon. The former is through the response header whose value is noted in `|instance|'s [=fenced frame config instance/cross-origin reporting allowed=]`. The latter is through the `crossOriginExposed` bit.

Since a `DOMString` event is just that, a `DOMString`, it has no `crossOriginExposed` bit, i.e. no way to opt in for cross-origin beacons. Because of that, there's no way we can safely support this feature for DOMString events. So it doesn't matter what kind of opt-in the document loaded with the fenced frame config had; it will just be ignored.

The only place we need to check `|instance|'s [=fenced frame config instance/cross-origin reporting allowed=]?` this is in the "If event is a [FenceEvent](https://github.com/WICG/fenced-frame/pull/152#dictdef-fenceevent):" section.

domfarolino: Got it, thanks.

domfarolino (approved):

PR #153, opened 2024-04-15
Add IntersectionObserver metrics bit

gtanzer:


gtanzer:
Or should this go in https://github.com/WICG/fenced-frame/blob/master/explainer/README.md#privacy-considerations instead?

explainer/integration_with_web_platform.md
```diff
@@ -11,7 +11,7 @@ The origin of the fenced frame will be the regular origin that it got navigated
...
+Viewability events using APIs like [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) can be a communication channel from the embedding context to the fenced frame. Viewability events are important information in the ads workflow for conversion reports and payments and would be made available in Protected Audience and other ads use cases via separate mechanisms. Initially, intersection observer will be fully supported in fenced frames and it will be phased out only when the alternative mechanism is launched. In the meantime, we plan to add metrics to understand honest use and detect abuse, to help guide the development of these eventual mechanisms.
```
shivanigithub, line 14: "Viewability events using APIs like [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) can be used as a communication channel from the embedding context to the fenced frame. However, to be used as a covert channel, it does require movement of the frame containing the ad on the screen and therefore is hostile to user experience.
Viewability events are important information in the ads workflow for conversion reports and billing and therefore, intersection observer will be fully supported in fenced frames and it will be phased out only if/when an alternative mechanism is launched. In the meantime, we plan to add metrics to understand honest use and detect abuse, to help guide the need/development of these eventual mechanisms or mitigations."

shivanigithub: Let's also remove the contents in the "intersection observer" section in https://github.com/WICG/fenced-frame/blob/master/explainer/README.md and link this section in that point.

gtanzer: Done for both

gtanzer:
@shivanigithub addressed comments

shivanigithub (approved):
lgtm, thanks!

PR #154, opened 2024-04-15
Add bit about top-level navigation metrics

gtanzer:


explainer/README.md
```diff
@@ -179,7 +179,7 @@ The fenced frame’s main goal is to improve privacy by disallowing communicatio
...
+*   **Navigation url:** Since fenced frames are allowed to open popups or navigate the top-level page in some use cases, gated on user activation, the navigation url can carry bits of information out of the fenced frame tree. If the embedder and the destination are same-origin, the information in the url and embedder's info can be joined locally on navigation. This might need mitigations going forward (currently being brainstormed); we plan to add metrics to understand how often this happens in practice. Additionally, this is vulnerable to the network side channel as mentioned above if the embedding site and destination site are colluding, even when not same-origin.
```
shivanigithub, line 182: "Additionally, this is vulnerable to the network side channel as mentioned above if the embedding site and destination site are colluding, even when not same-origin. Although it is less concerning than non-navigation network requests, since navigation is only allowed on user interaction like a click."

shivanigithub (approved):
lgtm, thanks!

gtanzer:
@shivanigithub addressed comments

PR #156, opened 2024-04-26
Spec for window.fence.notifyEvent and onfencedtreeclick

VergeA:
This adds event handlers for `onfencedtreeclick`, and describes the algorithm for `window.fence.notifyEvent()`, which fires `fencedtreeclick` events from the fenced frame's document to the `HTMLFencedFrameElement` in the embedder.

***
<a href="https://pr-preview.s3.amazonaws.com/VergeA/fenced-frame/pull/156.html" title="Last updated on Oct 21, 2024, 4:40 PM UTC (160aadd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/156/73d37f8...VergeA:160aadd.html" title="Last updated on Oct 21, 2024, 4:40 PM UTC (160aadd)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1644,7 +1644,7 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
       undefined reportEvent(optional ReportEventType event = {});
       undefined setReportEventDataForAutomaticBeacons(optional FenceEvent event = {});
       sequence&lt;FencedFrameConfig&gt; getNestedConfigs();
-      undefined notifyEvent(optional Event event = {});
+      undefined notifyEvent(Event event);
   };
 </pre>
 
@@ -1841,8 +1841,6 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
 
       * |navigable| is not a [=fenced navigable container/fenced navigable=];
-
-      * |event| is null;
   
       * |event|'s {{Event/isTrusted}} is false;
   
@@ -1857,13 +1855,13 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   1. Let |parentNavigable| be |navigable|'s [=navigable/unfenced parent=].
 
   1. [=Queue a global task=] on the [=DOM manipulation task source=] given |parentNavigable|'s     
-     [=navigable/active window=] to fire a "<code>[=fencedtreeclick=]</code>" event:
+     [=navigable/active window=] to run these steps:
   
      1. Perform the [=activation notification=] steps.
 
      1. [=Fire an event=] named "<code>[=fencedtreeclick=]</code>" at |navigable|'s 
         [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}} attributes to `true`. When running the 
-        <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all invocations of this method.
+        <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, set the <var ignore=''>time</var> to an [=implementation-defined=] value that is consistent across all invocations of this method.
 
   <wpt>
     /fenced-frame/notify-event-iframe.https.html
--- a/spec.bs
+++ b/spec.bs
@@ -117,6 +117,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: create a new browsing context and document; url: creating-a-new-browsing-context
       text: create a new browsing context group and document; url: creating-a-new-browsing-context-group
       text: document base url; url: document-base-url
+      text: fully active; url: fully-active
       text: initialize the navigable; url: initialize-the-navigable
       text: node navigable; url: node-navigable
       text: system visibility state; url: system-visibility-state
@@ -164,6 +165,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: interaction.html
       text: activation notification; url: activation-notification
       text: consume user activation; url: consume-user-activation
+      text: transient activation; url: transient-activation 
       text: activation; url: activation
       text: click focusable; url: click-focusable
       text: focusable area; url: focusable-area
@@ -188,6 +190,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       for: environment
         text: target browsing context; url: concept-environment-target-browsing-context
       text: navigation and traversal task source
+      text: event handlers on elements, document objects, and window objects; url: event-handlers-on-elements,-document-objects,-and-window-objects
+      text: idl definitions; url: idl-definitions
     urlPrefix: document-sequences.html
       for: browsing context
         text: active document; url: active-document
@@ -1633,6 +1637,7 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
       undefined reportEvent(optional ReportEventType event = {});
       undefined setReportEventDataForAutomaticBeacons(optional FenceEvent event = {});
       sequence&lt;FencedFrameConfig&gt; getNestedConfigs();
+      undefined notifyEvent(optional Event event = {});
   };
 </pre>
 
@@ -1836,6 +1841,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   <wpt>
     /fenced-frame/get-nested-configs.https.html
     /fenced-frame/config-cross-origin-apis.https.html
+  </wpt>
+</div>
+
+<div algorithm>
+  The <dfn method for=Fence>notifyEvent(|event|)</dfn> method steps are:
+
+  1. [=Assert=]: [=this=]'s {{Document}} is [=fully active=].
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=navigable=].
+  
+  1. If |navigable| is not a [=fenced navigable container/fenced navigable=], then
+     throw a {{SecurityError}} {{DOMException}}.
+
+  1. If |event| is null, then throw a {{SecurityError}} {{DOMException}}.
+  
+  1. If |event|'s {{Event/isTrusted}} is false, then throw a {{SecurityError}} {{DOMException}}.
+  
+  1. If |event|'s [=Event/dispatch flag=] is unset, then throw a {{SecurityError}} {{DOMException}}.
+
+  1. If |event|'s  {{Event/type}} is not "fencedtreeclick", throw a {{SecurityError}} 
+     {{DOMException}}.
+  
+  1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
+
+  1. [=Consume user activation=] for [=this=]'s [=relevant global object=].
+
+  1. [=Fire an event=] named "fencedtreeclick" at |navigable|'s 
+     [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}}
+     attributes to `true`. When running the 
+     <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, 
+     set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all 
+     invocations of this method.
+
+  <wpt>
+    /fenced-frame/notify-event-iframe.https.html
+    /fenced-frame/notify-event-invalid.https.html
+    /fenced-frame/notify-event-nested-fenced-frames.https.html
+    /fenced-frame/notify-event-success.https.html
+    /fenced-frame/notify-event-transient-user-activation.https.html
   </wpt>
 </div>
 
@@ -2954,6 +2998,22 @@ The <a href=https://html.spec.whatwg.org/#page-visibility>Page visibility</a> se
 modified such that the first step of the algorithm that runs when a user-agent changes a
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
+
+<h3 id=events>Events</h3>
+
+The {{Fence/notifyEvent}} method of the {{Fence}} interface introduces a new event type, `fencedtreeclick`.
+
+The [=event handlers on elements, Document objects, and Window objects=] section of [[HTML]] is modified to 
+include a new row. The [=event handler=] name is `onfencedtreeclick`, and the [=event handler event type=] is
+`fencedtreeclick`.
+
+The {{GlobalEventHandlers}} interface is modified as follows:
+
+<pre class=idl>
+  partial interface mixin GlobalEventHandlers {
+    attribute EventHandler onfencedtreeclick;
+  };
+</pre>
 
 <h2 id="interaction-with-other-specs">Interactions with other specifications</h2>
 
--- a/spec.bs
+++ b/spec.bs
@@ -1851,17 +1851,17 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
   1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=navigable=].
   
-  1. If |navigable| is not a [=fenced navigable container/fenced navigable=], then
-     throw a {{SecurityError}} {{DOMException}}.
+  1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
 
-  1. If |event| is null, then throw a {{SecurityError}} {{DOMException}}.
-  
-  1. If |event|'s {{Event/isTrusted}} is false, then throw a {{SecurityError}} {{DOMException}}.
-  
-  1. If |event|'s [=Event/dispatch flag=] is unset, then throw a {{SecurityError}} {{DOMException}}.
+      1. |navigable| is not a [=fenced navigable container/fenced navigable=].
 
-  1. If |event|'s  {{Event/type}} is not "fencedtreeclick", throw a {{SecurityError}} 
-     {{DOMException}}.
+      1. |event| is null.
+  
+      1. |event|'s {{Event/isTrusted}} is false.
+  
+      1. |event|'s [=Event/dispatch flag=] is unset.
+
+      1. |event|'s  {{Event/type}} is not "fencedtreeclick".
   
   1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
 
--- a/spec.bs
+++ b/spec.bs
@@ -1848,18 +1848,20 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   
       * |event|'s [=Event/dispatch flag=] is unset;
 
-      * |event|'s {{Event/type}} is not <code>click</code>
+      * |event|'s {{Event/type}} is not "<code>click</code>"
   
   1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
 
   1. [=Consume user activation=] for [=this=]'s [=relevant global object=].
 
-  1. [=Fire an event=] named <code>[=fencedtreeclick=]</code> at |navigable|'s 
-     [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}}
-     attributes to `true`. When running the 
-     <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, 
-     set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all 
-     invocations of this method.
+  1. Let |parentNavigable| be |navigable|'s [=navigable/unfenced parent=].
+
+  1. [=Queue a global task=] on the [=DOM manipulation task source=] given |parentNavigable|'s     
+     [=navigable/active window=] to fire a "<code>[=fencedtreeclick=]</code>" event:
+  
+     1. [=Fire an event=] named "<code>[=fencedtreeclick=]</code>" at |navigable|'s 
+        [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}} attributes to `true`. When running the 
+        <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all invocations of this method.
 
   <wpt>
     /fenced-frame/notify-event-iframe.https.html
@@ -3004,38 +3006,9 @@ navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
 <h3 id=events>Events</h3>
 
-<h4 id=fencedtreeclick-header><code>fencedtreeclick</code> event type</h4>
-The <a href="https://html.spec.whatwg.org/C#events-2">Events</a> table in the
-<a href="https://html.spec.whatwg.org/multipage/indices.html">Index</a> section of [[HTML]] is
-modified to include a new row.
-
-<table>
-  <thead>
-    <tr>
-      <th>Event</th> 
-      <th>Interface</th> 
-      <th>Interesting targets</th> 
-      <th>Description</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <td><dfn><code>fencedtreeclick</code></dfn></td> 
-      <td>{{Event}}</td> 
-      <td>Elements</td>
-      <td>
-        Fired at a <{fencedframe}> element asynchronously after its
-        [=fenced navigable container/fenced navigable=]'s {{Document}} calls {{Fence/notifyEvent()}}
-      </td>
-    </tr>
-  </tbody>
-</table>
-
-<h4 id=onfencedtreeclick-header><code>onfencedtreeclick</code> event handler</h4>
+<h4 id=onfencedtreeclick-header><code>[=onfencedtreeclick=]</code> event handler</h4>
 The table in the <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects">
-event handlers on elements, Document objects, and Window objects</a> section of [[HTML]] is modified 
-to include a new row. The [=event handler=] name is <code>onfencedtreeclick</code>, and the 
-[=event handler event type=] is <code>fencedtreeclick</code>.
+event handlers on elements, Document objects, and Window objects</a> section of [[HTML]] is modified to include a new row.
 
 <table>
   <thead>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -164,6 +165,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: interaction.html
       text: activation notification; url: activation-notification
       text: consume user activation; url: consume-user-activation
+      text: transient activation; url: transient-activation
```
domfarolino, line 168: This is exported by HTML so shouldn't need to include this here. The same seems to be true for fully active.

I believe if [it appears in webdex](https://dontcallmedom.github.io/webdex/t.html#Transient%20activation%40%40html%25%25dfn), it is exported and doesn't need to be added to this section manually. (Also for transient activation I checked the HTML source).

spec.bs
```diff
@@ -1699,6 +1704,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   </wpt>
 </div>
 
+<div algorithm>
+  The <dfn method for=Fence>notifyEvent(|event|)</dfn> method steps are:
+
+  1. [=Assert=]: [=this=]'s {{Document}} is [=fully active=].
```
domfarolino, line 1710: Why is this assert true? Can't you call this API on a detached document?

domfarolino, line 1710: I think this should use `[=Document/fully active]`. See https://dontcallmedom.github.io/webdex/f.html#fully%20active%40%40Document%40dfn.

spec.bs
```diff
@@ -1699,6 +1704,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=navigable=].
+  
+  1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
+
+      1. |navigable| is not a [=fenced navigable container/fenced navigable=].
```
domfarolino, line 1716: This should be an unordered list of conditions with semi-colons after each line except for the last. See https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document:~:text=Otherwise%2C%20if%20all%20of%20the%20following%20are%20true for example.

spec.bs
```diff
@@ -1699,6 +1704,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+      1. |event| is null.
+  
+      1. |event|'s {{Event/isTrusted}} is false.
+  
+      1. |event|'s [=Event/dispatch flag=] is unset.
+
+      1. |event|'s  {{Event/type}} is not "fencedtreeclick".
```
domfarolino, line 1724:
```suggestion
      1. |event|'s {{Event/type}} is not "fencedtreeclick".
```

spec.bs
```diff
@@ -2792,6 +2836,22 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h3 id=events>Events</h3>
+
+The {{Fence/notifyEvent}} method of the {{Fence}} interface introduces a new event type, `fencedtreeclick`.
+
+The [=event handlers on elements, Document objects, and Window objects=] section of [[HTML]] is modified to
```
domfarolino, line 2843: This should not be a definition it should just be a manual anchor link to that section.

spec.bs
```diff
@@ -2792,6 +2836,22 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h3 id=events>Events</h3>
+
+The {{Fence/notifyEvent}} method of the {{Fence}} interface introduces a new event type, `fencedtreeclick`.
+
+The [=event handlers on elements, Document objects, and Window objects=] section of [[HTML]] is modified to 
+include a new row. The [=event handler=] name is `onfencedtreeclick`, and the [=event handler event type=] is
```
domfarolino, line 2844: Sometimes you use backticks and other times you use double quotes when wrapping `fencedtreeclick`. I think both are wrong though. There should be an actual definition for this wrapped in code, like https://html.spec.whatwg.org/multipage/media.html#event-media-change, and each reference of it should be a code-wrapped reference to that definition.

For examples in our own spec, see https://wicg.github.io/fenced-frame/#automatic-beacon-event-type-reservedtop_navigation and https://wicg.github.io/fenced-frame/#ref-for-automatic-beacon-event-type-reservedtop_navigation%E2%91%A1.

spec.bs
VergeA: Got it, removed them both. Thanks for the webdex link! Was wondering where all the autodef magic was coming from.

spec.bs

spec.bs

spec.bs

spec.bs

spec.bs
```diff
@@ -2792,6 +2834,23 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h3 id=events>Events</h3>
+
+The {{Fence/notifyEvent}} method of the {{Fence}} interface introduces a new event type, <code><dfn>fencedtreeclick</dfn></code>.
```
domfarolino, line 2839: This method does not introduce this new event type. This spec does; but any code create events of this type. I would dissociate these two.

spec.bs
```diff
@@ -2792,6 +2834,23 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+
+The <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects">
```
domfarolino, line 2841:
```suggestion
The table in the <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects">
```

Note that wrapping might be off after this suggestion?

spec.bs
```diff
@@ -2792,6 +2834,23 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+event handlers on elements, Document objects, and Window objects</a> section of [[HTML]] is modified to 
+include a new row. The [=event handler=] name is <code>onfencedtreeclick</code>, and the [=event handler event type=] is
```
domfarolino, line 2843:
```suggestion
include a new row. The [=event handler=] name is <dfn><code>onfencedtreeclick</code></dfn>, and the [=event handler event type=] is
```

This is where the definition should be. See that in HTML, the definitions are all in the table.

domfarolino, line 2843: It'd be best to replicate the style of table that you see in https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects. Just have a single row pointing to the stuff. you want.

### Event handler definition
Note that the "event handler" column should include a definition for `onfencedtreeclick`, and the "event handler event type" should be a _link_ (not dfn) to the new `fencedtreeclick` event type.

### Event type definition
Then the `fencedtreeclick` event type should be defined in a new section of this spec that monkeypatches the table you find in https://html.spec.whatwg.org/C#events-2, which is responsible for defining all of the event types (or linking to them when another spec defines them; but since we're monkeypatching HTML here, we want to add the dfn to that table).

### Misc

You'll also need to monkeypatch this list (https://html.spec.whatwg.org/multipage/dom.html#global-attributes%3Aevent-handler-content-attributes) to include a link to the `onfencedtreeclick` event handler, which is defined in the table I mention in "Event handler definition" section above.

spec.bs
domfarolino: Yeah, see https://github.com/WICG/fenced-frame/pull/156#discussion_r1586865226.

spec.bs
domfarolino: Not being "fully active" is a superset of being detached. Detached documents are all "not fully active". Chromium handles this in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/fence.cc;l=425-430;drc=599cc06027321ac29b8b518dfb8f4de5f7449254.

You can get and interact with a detached frame or document like so:
```js
const iframe = document.createElement('iframe');
document.body.append(iframe);
const innerFence = iframe.contentWindow.fence;
iframe.remove();
innerFence.notifyEvent(); // Chromium will gracefully early-return. The spec should too.
```

spec.bs
VergeA: I believe I've got all of this addressed, PTAL!

VergeA: Done, but I've added the definition in the table itself, see https://github.com/WICG/fenced-frame/pull/156#discussion_r1586865226

spec.bs
VergeA: I removed this line.

spec.bs
VergeA: I changed the assert here to an early-return.

spec.bs
```diff
@@ -1699,6 +1711,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  
+      * |event|'s [=Event/dispatch flag=] is unset;
+
+      * |event|'s {{Event/type}} is not <code>click</code>
+  
+  1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
+
+  1. [=Consume user activation=] for [=this=]'s [=relevant global object=].
+
+  1. [=Fire an event=] named <code>[=fencedtreeclick=]</code> at |navigable|'s
```
domfarolino, line 1737: You can't fire this synchronously from here. I assume the implementation doesn't either, right (I'm pretty sure we asyncly go through the browser process and mojo and so on)? It should be asynchronous probably posting a task to the unfenced parent's Window's event loop on the DOM manipulation task source to fire the event inside _that_ outer event loop. Does that make sense?

VergeA: Yeah the impl does go through Mojo. I wasn't sure how specific we had to be about how the event gets to its destination.

I am not familiar with the specifics of task sources and event loops yet, but I understand that ultimately the parent should be the one responsible for firing the event. In the impl, the IPC is received by the <fencedframe>, which essentially fires the event at itself. Does firing an event on an element inherently use the associated Window's event loop?

spec.bs
```diff
@@ -2792,6 +2843,68 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><dfn export><code>fencedtreeclick</code></dfn></td> 
+      <td>{{Event}}</td> 
+      <td>Elements</td>
+      <td>
+        Fired at a <{fencedframe}> element when its contained content calls {{Fence/notifyEvent}}
```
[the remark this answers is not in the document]
VergeA: Done

spec.bs
```diff
@@ -2792,6 +2843,68 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+    <tr>
+      <th>Event</th> 
+      <th>Interface</th> 
+      <th>Interesting targets</th> 
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><dfn export><code>fencedtreeclick</code></dfn></td>
```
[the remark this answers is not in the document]
VergeA: Got rid of the export, here and below.

spec.bs
```diff
@@ -1699,6 +1711,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
+
+  1. [=Consume user activation=] for [=this=]'s [=relevant global object=].
+
+  1. [=Fire an event=] named <code>[=fencedtreeclick=]</code> at |navigable|'s 
+     [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}}
+     attributes to `true`. When running the 
+     <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, 
+     set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all 
+     invocations of this method.
```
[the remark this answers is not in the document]
VergeA: Yeah, the idea was to not provide specific timestamps from notifyEvent() to the embedder. Probably not strictly necessary given that the embedder can grab their own close-enough timestamp, but follows the principle of passing as little data to the embedder as possible.

spec.bs
```diff
@@ -1699,6 +1711,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   </wpt>
 </div>
 
+<div algorithm>
+  The <dfn method for=Fence>notifyEvent(|event|)</dfn> method steps are:
+
+  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then return.
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=navigable=].
```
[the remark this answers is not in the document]
VergeA: Done. Tried [=window-navigable=] and {{Window/navigable}} before realizing the [=Window/navigable=] custom definition was at the top of this file.

spec.bs
domfarolino: No, firing an event at an element uses whatever event loop is on the stack. It's no different from calling a function really, and you can't call a JS function across documents that live in different processes.

That's why you have to queue a task _from_ the child onto the parent's event loop, and the task will fire an event from entirely within the parent's context. Check out the usages of queue a task to see how to use it, or feel free to ask me specifics.

spec.bs
```diff
@@ -2792,6 +2843,69 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+      <td>Elements</td>
+      <td>
+        Fired at a <{fencedframe}> element asynchronously after its
+        [=fenced navigable container/fenced navigable=]'s {{Document}} calls {{Fence/notifyEvent()}}
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+<h4 id=onfencedtreeclick-header><code>onfencedtreeclick</code> event handler</h4>
```
domfarolino, line 2875: Let's link to the dfn of the `onfencedtreeclick` handler here.

spec.bs
```diff
@@ -2792,6 +2843,69 @@ modified such that the first step of the algorithm that runs when a user-agent c
 [=traversable navigable=]'s [=system visibility state=] calls the [=Document/inclusive descendant
 navigables=] algorithm with [=inclusive-dn-unfenced|unfenced=] set to true.
 
+<h3 id=events>Events</h3>
+
+<h4 id=fencedtreeclick-header><code>fencedtreeclick</code> event type</h4>
```
domfarolino, line 2848: Let's link to the dfn of `fencedtreeclick` here.

spec.bs
domfarolino, line 2875: If I'm reading it correctly, in HTML this table comes before the one you have above, so can you reorder these? (Unless I'm wrong about the ordering in HTML, and this was intentional.)

spec.bs
```diff
@@ -2792,6 +2843,69 @@ modified such that the first step of the algorithm that runs when a user-agent c
...
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+<h4 id=onfencedtreeclick-header><code>onfencedtreeclick</code> event handler</h4>
+The table in the <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects">
+event handlers on elements, Document objects, and Window objects</a> section of [[HTML]] is modified 
+to include a new row. The [=event handler=] name is <code>onfencedtreeclick</code>, and the 
+[=event handler event type=] is <code>fencedtreeclick</code>.
```
domfarolino, line 2879: This last sentence does n't seem all that useful, I'd remove it.

VergeA:
FYI I had to force-push my feature branch after rebasing, which is why there are 5 new commits. There have been no new changes, just a rebase. Sorry for letting this sit, hoping to get back to it soon!

spec.bs
VergeA: Done.

spec.bs
VergeA: You are indeed correct, reordered them.

spec.bs
VergeA: Done

spec.bs
VergeA: Done

spec.bs
```diff
@@ -1699,6 +1711,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+<div algorithm>
+  The <dfn method for=Fence>notifyEvent(|event|)</dfn> method steps are:
+
+  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then return.
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=Window/navigable=].
+  
+  1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
+
+      * |navigable| is not a [=fenced navigable container/fenced navigable=];
```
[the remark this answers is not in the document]
VergeA: We are only exposing this interface in fenced frame root documents, and IIUC URN iframes expose window.fence as well, so we shouldn't allow notifyEvent there.

spec.bs
```diff
@@ -1699,6 +1711,45 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+      * |event| is null;
+  
+      * |event|'s {{Event/isTrusted}} is false;
+  
+      * |event|'s [=Event/dispatch flag=] is unset;
+
+      * |event|'s {{Event/type}} is not <code>click</code>
```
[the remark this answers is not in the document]
VergeA: Done

spec.bs
VergeA: I think I've figured out how to queue the event firing task in the unfenced parent's event loop, PTAL!

spec.bs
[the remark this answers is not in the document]
VergeA: Based on conversation in the TPAC breakout, I still think this censored timestamp field is good to keep (even if the parent knows all the information, it's still subject to the time taken between when the fenced frame sends notifyEvent and when the browser actually pops the event listener off of the event loop).

spec.bs
```diff
@@ -1819,6 +1831,49 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |parentNavigable| be |navigable|'s [=navigable/unfenced parent=].
+
+  1. [=Queue a global task=] on the [=DOM manipulation task source=] given |parentNavigable|'s     
+     [=navigable/active window=] to fire a "<code>[=fencedtreeclick=]</code>" event:
+  
+     1. Perform the [=activation notification=] steps.
+
+     1. [=Fire an event=] named "<code>[=fencedtreeclick=]</code>" at |navigable|'s 
+        [=fenced navigable container=]. Initialize the event's {{Event/bubbles}} and {{Event/cancelable}} attributes to `true`. When running the 
+        <a href="https://dom.spec.whatwg.org/#inner-event-creation-steps">inner event creation steps</a>, set the <var ignore=''>time</var> to a user-agent-defined value that is consistent across all invocations of this method.
```
domfarolino, line 1866: use `[=implementation-defined=]` instead.

spec.bs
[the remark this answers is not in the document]
domfarolino: So what should the value _actually_ be? Should we just make it always null or something? What does it mean for it to be consistent across all invocations of this method? Certainly a proper "time stamp" can't behave like that, right? (Hence why I ask if we should make it "null")

spec.bs
```diff
@@ -1633,6 +1644,7 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
       undefined reportEvent(optional ReportEventType event = {});
       undefined setReportEventDataForAutomaticBeacons(optional FenceEvent event = {});
       sequence&lt;FencedFrameConfig&gt; getNestedConfigs();
+      undefined notifyEvent(optional Event event = {});
```
domfarolino, line 1647: This shouldn't be optional or have a default value, right? The implementation doesn't: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/fence.idl;l=15;drc=d06e0de0964242fff0a093360e9065401d9e0e00

spec.bs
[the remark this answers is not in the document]
domfarolino: OK so this will throw a security error for both URN iframes, as well as iframes nested within fenced frames, right? And that's the intention?

spec.bs
```diff
@@ -1819,6 +1831,49 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then return.
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=Window/navigable=].
+  
+  1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
+
+      * |navigable| is not a [=fenced navigable container/fenced navigable=];
+
+      * |event| is null;
```
domfarolino, line 1845: I don't think `event` is nullable in the current PR (where `event` is optional), and it definitely is not if we adopt the changes I suggest above about making the parameter optional.

spec.bs
```diff
@@ -1819,6 +1831,49 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+      * |event|'s {{Event/type}} is not "<code>click</code>"
+  
+  1. If [=this=]'s [=relevant global object=] does not have [=transient activation=], then return.
+
+  1. [=Consume user activation=] for [=this=]'s [=relevant global object=].
+
+  1. Let |parentNavigable| be |navigable|'s [=navigable/unfenced parent=].
+
+  1. [=Queue a global task=] on the [=DOM manipulation task source=] given |parentNavigable|'s     
+     [=navigable/active window=] to fire a "<code>[=fencedtreeclick=]</code>" event:
```
domfarolino, line 1860: I think this should be "to run these steps" instead of "to fire an event", since we manually fire the event _in the steps_ below, right?

domfarolino (commented):
Looks pretty good, I think we're just about there!

spec.bs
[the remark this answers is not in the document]
domfarolino: > We are only exposing this interface in fenced frame root documents

Actually, this isn't true right? @blu25 [mentioned](https://github.com/WICG/fenced-frame/pull/146#discussion_r1788261165) that cross-origin iframes inside a fenced frame get access to `window.fence`. Is that true, and if so does it change the calculus here at all?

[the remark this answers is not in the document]
blu25: "This interface" is referring to `fence.notifyEvent()` specifically.

> cross-origin iframes inside a fenced frame get access to window.fence. Is that true, and if so does it change the calculus here at all?

[Step 4 of the create browsing context patch](https://wicg.github.io/fenced-frame/#creating-browsing-contexts-patch) is the piece that has subframes inherit their creator's fenced frame config instance, which is then detected by the [fence getter](https://wicg.github.io/fenced-frame/#dom-window-fence) and ultimately lets subframes call `window.fence`. I don't think that fact changes the calculus of this algorithm, since we only want fenced frame roots to be able to call this method, which is what's currently happening.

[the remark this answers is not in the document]
VergeA: Yes, +1 to everything Liam said (sorry, I now realize my wording was a bit confusing).

spec.bs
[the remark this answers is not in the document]
VergeA: When writing the Chromium implementation, I spent a while thinking about this. Chromium has a TODO to remove [remove](https://source.chromium.org/chromium/chromium/src/+/main:base/time/time.h;drc=4b44bd63028a5aa82976dc4d7ed51cb78cbc6891;l=415) the concept of "null" time. As a result, I used the [Unix epoch](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/events/event.h;drc=a125f7c5687ed22bcd476ce713160096a3d0b037;l=112) instead, since that's a well-known fixed value. I figured given that this timestamp is intentionally a placeholder, its exact value isn't that important.

If we'd like to pick a specific value instead of leaving it up to the user agent, we could just use 0, or the Unix epoch, given that it's defined in the [same spec](https://w3c.github.io/hr-time/#sec-domhighrestimestamp) as DomHighResTimeStamp.

By "consistent across all invocations of this method" I meant: any time a "fencedtreeclick" event is fired, regardless of what the actual current time is, all event objects will have the same timestamp. 

Thinking about it more, I'm leaning toward just using the Unix epoch in this spec directly, to make things less ambiguous.

spec.bs
VergeA: Done.

spec.bs
VergeA: I deleted this line, since I made the change you specified above about no optional/default value being provided.

spec.bs
VergeA: Fixed.

spec.bs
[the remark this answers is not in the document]
domfarolino: OK, so the interface _and_ method are exposed to script in all of these fenced and nested contexts, and we have to catch the nested ones specifically and throw security errors in the case that the method is called in there. Makes sense.

spec.bs
```diff
@@ -1819,6 +1831,47 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
 </div>
 
+<div algorithm>
+  The <dfn method for=Fence>notifyEvent(|event|)</dfn> method steps are:
+
+  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then return.
+
+  1. Let |navigable| be [=this=]'s [=relevant global object=]'s [=Window/navigable=].
+  
+  1. If any of the following conditions are met, throw a {{SecurityError}} {{DOMException}}:
```
domfarolino, line 1841:
```suggestion
  1. If any of the following conditions are met, then throw a {{SecurityError}} {{DOMException}}:
```

spec.bs
```diff
@@ -191,6 +192,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       for: environment
         text: target browsing context; url: concept-environment-target-browsing-context
       text: navigation and traversal task source
+      text: event handlers on elements, document objects, and window objects; url: event-handlers-on-elements,-document-objects,-and-window-objects
```
domfarolino, line 195: This can be removed now

spec.bs
```diff
@@ -191,6 +192,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+      text: idl definitions; url: idl-definitions
```
domfarolino, line 196: Is this needed?

VergeA: Removed.

spec.bs
VergeA: Done.

spec.bs
VergeA: Done.

domfarolino (approved):

Issue #158, opened 2024-05-08
Allow-Cross-Origin-Event-Reporting not defined?


miketaylr:
Looking at https://github.com/WICG/fenced-frame/pull/152 - I see `Allow-Cross-Origin-Event-Reporting` referenced, but not defined. I can see that it's parsed as a structured field Item, but doesn't say if it's allowed values are string literals `true`/`false` or boolean`?0`/`?1`.

https://httpwg.org/specs/rfc8941.html#rfc.section.2 is a useful guide to reference.

domfarolino:
@blu25

PR #159, opened 2024-05-08
Add definitions for event reporting opt in headers

blu25:
This issue was brought up in https://github.com/WICG/fenced-frame/issues/158.

This PR adds official definitions for the `Allow-Fenced-Frame-Automatic-Beacons` and the `Allow-Cross-Origin-Event-Reporting` response headers that are used for cross-origin opt in for their respective features.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/159.html" title="Last updated on May 21, 2024, 12:04 PM UTC (bff5d37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/159/dbb22ca...bff5d37.html" title="Last updated on May 21, 2024, 12:04 PM UTC (bff5d37)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2498,19 +2498,22 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 <h4 id=allow-automatic-beacons>The
 \`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
 
-A {{Document}} that is not [=same origin=] with its [=Document/browsing context=]'s [=fenced frame
-config instance=]'s [=fenced frame config instance/mapped url=] can opt into sending an automatic
-beacon when it navigates by using the <dfn
-http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header. This
-header is a [=structured header=] whose value must be either the [=string=] "true" or "false".
+Serving a document [=resource=] that gets loaded into a {{Document}} that is not [=same origin=]
+with its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
+instance/mapped url=] with the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header can
+automatically opt in the {{Document}} to having automatic beacon events trigger when it performs
+navigations. This header is a [=structured header=] whose value must be either the [=string=] "true"
+or "false".
 
 <h4 id=allow-cross-origin-reporting>The
 \`<a http-header><code>Allow-Cross-Origin-Event-Reporting</code></a>\` HTTP response header</h4>
 
-A {{Document}} loaded from a [=fenced frame config instance=] can opt into its cross-origin [=child
-navigables=]' {{Document}}s sending {{Fence/reportEvent()}} beacons by using the <dfn
-http-header><code>Allow-Cross-Origin-Event-Reporting</code></dfn> HTTP response header. This header
-is a [=structured header=] whose value must be either the [=string=] "true" or "false".
+Serving a document [=resource=] that gets loaded into a <{fencedframe}> by a [=fenced frame config
+instance=] with the <dfn http-header><code>Allow-Cross-Origin-Event-Reporting</code></dfn> HTTP
+response header can automatically opt in the {{Document}}'s cross-origin [=child navigables=]
+to be able to send {{Fence/reportEvent()}} beacons. This header is a [=structured header=] whose
+value must be either the [=string=] "true" or "false".
 
 <h4 id=coop-coep>COOP, COEP, and cross-origin isolation</h4>
 
--- a/spec.bs
+++ b/spec.bs
@@ -2498,22 +2498,22 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 <h4 id=allow-automatic-beacons>The
 \`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
 
-Serving a document [=resource=] that gets loaded into a {{Document}} that is not [=same origin=]
-with its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
+Serving a document resource that gets loaded into a {{Document}} that is not [=same origin=] with
+its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
 instance/mapped url=] with the <dfn
-http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header can
-automatically opt in the {{Document}} to having automatic beacon events trigger when it performs
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header
+automatically opts in the {{Document}} to having automatic beacon events trigger when it performs
 navigations. This header is a [=structured header=] whose value must be either the [=string=] "true"
 or "false".
 
 <h4 id=allow-cross-origin-reporting>The
 \`<a http-header><code>Allow-Cross-Origin-Event-Reporting</code></a>\` HTTP response header</h4>
 
-Serving a document [=resource=] that gets loaded into a <{fencedframe}> by a [=fenced frame config
+Serving a document resource that gets loaded into a <{fencedframe}> by a [=fenced frame config
 instance=] with the <dfn http-header><code>Allow-Cross-Origin-Event-Reporting</code></dfn> HTTP
-response header can automatically opt in the {{Document}}'s cross-origin [=child navigables=]
-to be able to send {{Fence/reportEvent()}} beacons. This header is a [=structured header=] whose
-value must be either the [=string=] "true" or "false".
+response header automatically opts in the {{Document}}'s cross-origin [=child navigables=] to be
+able to send {{Fence/reportEvent()}} beacons. This header is a [=structured header=] whose value
+must be either the [=string=] "true" or "false".
 
 <h4 id=coop-coep>COOP, COEP, and cross-origin isolation</h4>
 
--- a/spec.bs
+++ b/spec.bs
@@ -2495,6 +2495,24 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+A {{Document}} that is not [=same origin=] with its [=fenced frame config instance=]'s [=fenced
```
domfarolino, line 2501: Documents do not have fenced frame config instances, but I think browsing contexts do.

spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
...
+frame config instance/mapped url=] can opt into sending an automatic beacon when it navigates by
+using the <dfn http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response
+header. This header is a [=structured header=] whose value must be either the [=string=] "true" or
+"false".
+
+<h4 id=allow-cross-origin-reporting>The
+\`<a http-header><code>Allow-Cross-Origin-Event-Reporting</code></a>\` HTTP response header</h4>
+
+A {{Document}} loaded with a [=fenced frame config instance=] can opt into cross-origin
```
domfarolino, line 2510:
```suggestion
A {{Document}} loaded from a [=fenced frame config instance=] can opt into cross-origin
```

spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
...
+{{Document}}s in [=child navigables=] sending {{Fence/reportEvent()}} beacons by using the <dfn
```
domfarolino, line 2511: in whose child navigables? Its own child navigables right?

spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+A {{Document}} that is not [=same origin=] with its [=fenced frame config instance=]'s [=fenced
+frame config instance/mapped url=] can opt into sending an automatic beacon when it navigates by
```
domfarolino, line 2502: Remind me how a fenced frame config with a mapped URL can load a document whose origin is different from that mapped URL? Redirects?

spec.bs
blu25: Yes. I'll reword this to be more clear.

spec.bs
blu25: If the document navigates itself to a new URL, the fenced frame config instance [persists across the cross-document navigation](https://wicg.github.io/fenced-frame/#creating-browsing-contexts-patch:~:text=Inner%2Dcontent%2Dinitiated%20navigations%3A%20The%20fenced%20frame%20config%20instance%20referenced%20in%20the%20immediately%2Dpreceding%20point%20persists%20through%20this%20kind%20of%20cross%2DDocument%20navigation.).

Cross-origin child iframes will also [get the fenced frame config instance of its fenced traversable navigable](https://wicg.github.io/fenced-frame/#:~:text=(2)%20inherited%20during%20initial%20Document%20creation%20for%20Documents%20in%20child%20navigables%20whose%20traversable%20navigable%20is%20a%20fenced%20navigable%2C%20as%20handled%20in%20the%20%C2%A7%E2%80%AF3.3%20Modifications%20to%20creating%20browsing%20contexts%20section.), which will result in the origins being different.

domfarolino: OK thanks

spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
...
+beacon when it navigates by using the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header. This
+header is a [=structured header=] whose value must be either the [=string=] "true" or "false".
+
+<h4 id=allow-cross-origin-reporting>The
+\`<a http-header><code>Allow-Cross-Origin-Event-Reporting</code></a>\` HTTP response header</h4>
+
+A {{Document}} loaded from a [=fenced frame config instance=] can opt into its cross-origin [=child
+navigables=]' {{Document}}s sending {{Fence/reportEvent()}} beacons by using the <dfn
+http-header><code>Allow-Cross-Origin-Event-Reporting</code></dfn> HTTP response header. This header
```
domfarolino, line 2512: Documents can't use the header, but the server can set them. Or do you mean the document can literally use the header somehow with a web platform API? I would probably say something like: serving the document resource that gets loaded into a `<fencedframe>` by a fenced frame config instance with a FOO header can automatically opt in the document and its cross-origin child navigables to be able to send reportEvent() beacons etc. I think this is more clear since it traces the chronology / flow of how it actually works.

spec.bs
```diff
@@ -2495,6 +2495,23 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
...
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+A {{Document}} that is not [=same origin=] with its [=Document/browsing context=]'s [=fenced frame
+config instance=]'s [=fenced frame config instance/mapped url=] can opt into sending an automatic
+beacon when it navigates by using the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header. This
+header is a [=structured header=] whose value must be either the [=string=] "true" or "false".
```
domfarolino, line 2505: I would apply the below comment to this paragraph as well. Start out by talking about serving the document with this header, and then explain what it unlocks for the document that is loaded.

cbiesinger:
I'm curious, why is this a structured header with a string "true" or "false" instead of a boolean structured header?

blu25:
> I'm curious, why is this a structured header with a string "true" or "false" instead of a boolean structured header?

The implementation of this expects string literals (true/false) rather than boolean values (?0/?1). I wasn't aware of the [precedent set in RFC8941](https://www.rfc-editor.org/rfc/rfc8941#name-booleans) at the time, and I had copied the syntax of an existing flag that was also not using the boolean convention when I had implemented this feature. Since the code is already in place, and since changing this to a boolean would delay getting this feature out by at least a month, I think it's better to keep the implementation as is and make sure the spec matches the implementation.

spec.bs
```diff
@@ -2495,6 +2495,26 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+Serving a document [=resource=] that gets loaded into a {{Document}} that is not [=same origin=]
+with its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
+instance/mapped url=] with the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header can
```
domfarolino, line 2504: "can"? or does? https://infra.spec.whatwg.org/#conformance. Same with the other header section below.

blu25: Removed "can" from the 2 sections.

spec.bs
```diff
@@ -2495,6 +2495,26 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+Serving a document [=resource=] that gets loaded into a {{Document}} that is not [=same origin=]
```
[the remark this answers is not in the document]
blu25: Removed the links. I didn't realize the definition was specifically for localization.

miketaylr:
> would delay getting this feature out by at least a month

As far as long-term web platform health is concerned - 1 month is basically nothing. Any other compelling reasons to not do this work?

shivanigithub:
> 
> As far as long-term web platform health is concerned - 1 month is basically nothing. Any other compelling reasons to not do this work?

The team is actually proceeding with the change to boolean: https://chromium-review.googlesource.com/c/chromium/src/+/5542760

spec.bs
```diff
@@ -2495,6 +2496,26 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
```
miketaylr, line 2500: (Maybe a fenced frame style thing, but not sure why there's a backtick surrounding `<a><code></a></code>`. I see it's already in the spec elsewhere though - so free to ignore me I guess)

spec.bs
```diff
@@ -2495,6 +2496,26 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
...
+
+Serving a document resource that gets loaded into a {{Document}} that is not [=same origin=] with
+its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
+instance/mapped url=] with the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header
+automatically opts in the {{Document}} to having automatic beacon events trigger when it performs
```
miketaylr, line 2506: editorial nit, take it or leave it: consider removing "automatically" here and below - it seems redundant.

miketaylr (approved):
LGTM, with some take-em-or-leave-em nits.

spec.bs
blu25: I see backticks for headers defined and referenced in other specs, so I think that's standardized.

[example from the Fetch spec](https://fetch.spec.whatwg.org/#http-access-control-allow-origin)
[example from PA spec](https://wicg.github.io/turtledove/#ad-auction-signals-header)
[example from HTML spec](https://html.spec.whatwg.org/multipage/browsers.html#origin-keyed-agent-clusters)

miketaylr: I think it's just a personal preference thing - but like I said, feel free to ignore. :)

Issue #160, opened 2024-05-14
Spec does not describe navigator.deprecatedURNToURL


brusshamilton:
navigator.deprecatedURNToURL is not defined in the Fenced Frame spec (or the Protected Audiences spec), but it is shipping on Chrome. It should probably be defined somewhere near where [deprecatedReplaceInURN](https://wicg.github.io/fenced-frame/#dom-navigator-deprecatedreplaceinurn) is defined.

PR #161, opened 2024-05-21
Spec deprecatedURNtoURL

JensenPaul:
Addresses #160

***
<a href="https://pr-preview.s3.amazonaws.com/JensenPaul/fenced-frame/pull/161.html" title="Last updated on Jun 6, 2024, 1:33 PM UTC (6495120)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/161/4966a9e...JensenPaul:6495120.html" title="Last updated on Jun 6, 2024, 1:33 PM UTC (6495120)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -218,6 +218,7 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     text: structured header; url: #section-1
     for: structured header
       text: token; url: name-tokens
+      text: boolean; url: boolean
 spec: permissions-policy; urlPrefix: https://w3c.github.io/webappsec-permissions-policy
   type: dfn
     text: ASCII-serialized policy directive; url: serialized-policy-directive
@@ -2562,6 +2563,25 @@ container/fenced navigable=] will fail, as outlined in [[#navigation-patch]].
 
 </div>
 
+<h4 id=allow-automatic-beacons>The
+\`<a http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></a>\` HTTP response header</h4>
+
+Serving a document resource that gets loaded into a {{Document}} that is not [=same origin=] with
+its [=Document/browsing context=]'s [=fenced frame config instance=]'s [=fenced frame config
+instance/mapped url=] with the <dfn
+http-header><code>Allow-Fenced-Frame-Automatic-Beacons</code></dfn> HTTP response header opts in the
+{{Document}} to having automatic beacon events trigger when it performs navigations. This header is
+a [=structured header=] whose value must be a [=structured header/boolean=].
+
+<h4 id=allow-cross-origin-reporting>The
+\`<a http-header><code>Allow-Cross-Origin-Event-Reporting</code></a>\` HTTP response header</h4>
+
+Serving a document resource that gets loaded into a <{fencedframe}> by a [=fenced frame config
+instance=] with the <dfn http-header><code>Allow-Cross-Origin-Event-Reporting</code></dfn> HTTP
+response header opts in the {{Document}}'s cross-origin [=child navigables=] to be able to send
+{{Fence/reportEvent()}} beacons. This header is a [=structured header=] whose value must be a
+[=structured header/boolean=].
+
 <h4 id=coop-coep>COOP, COEP, and cross-origin isolation</h4>
 
 Outside of this specification, the \`<a http-header><code>Cross-Origin-Opener-Policy</code></a>\`
@@ -2849,8 +2869,7 @@ content in the <{fencedframe}> or its embedder, exactly one of two things will h
     1. Set |navigationParams|'s [=navigation params/fenced frame config instance=]'s [=fenced frame
        config/cross-origin reporting allowed=] to the result of running [=header list/get a
        structured field value=] on |navigationParams|'s [=navigation params/response=]'s
-       [=response/header list=] given "<code>Allow-Cross-Origin-Event-Reporting</code>" and
-       "`item`".
+       [=response/header list=] given [:Allow-Cross-Origin-Event-Reporting:] and "`item`".
 
     1. Set |browsingContext|'s [=browsing context/fenced frame config instance=] to
        |navigationParams|'s [=navigation params/fenced frame config instance=].
@@ -2859,11 +2878,11 @@ content in the <{fencedframe}> or its embedder, exactly one of two things will h
 
   10. Let |automaticBeaconsAllowed| be the result of running [=header list/get a structured field
       value=] on |navigationParams|'s [=navigation params/response=]'s [=response/header list=]
-      given "<code>Allow-Fenced-Frame-Automatic-Beacons</code>" and "`item`".
+      given [:Allow-Fenced-Frame-Automatic-Beacons:] and "`item`".
   
   Further rewrite step 10 (now step 12) to return a new {{Document}} with an additional parameter:
   : [=Document/automatic beacons allowed=]
-  :: |automaticBeaconsAllowed|
+  :: |automaticBeaconsAllowed|.
 </div>
 
 <span class=XXX>TODO: Call the [=fenced frame config instance/on navigate callback=] at the
--- a/spec.bs
+++ b/spec.bs
@@ -1531,7 +1531,7 @@ partial interface Navigator {
 
   1. Run the following steps [=in parallel=]:
 
-     1. Let |mapping| be |global|'s [=associated Document=]'s [=node navigable=]'s
+     1. Let |mapping| be |global|'s [=Window/navigable=]'s
         [=navigable/traversable navigable=]'s [=traversable navigable/fenced frame config mapping=].
 
      1. If |mapping|'s [=fenced frame config mapping/finalized config mapping=][|urn|], does not
--- a/spec.bs
+++ b/spec.bs
@@ -1733,7 +1733,7 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
     /fenced-frame/fence-report-event-cross-origin-urn-iframe-no-embedder-opt-in.https.html 
     /fenced-frame/fence-report-event-cross-origin-urn-iframe-no-subframe-opt-in.https.html 
     /fenced-frame/fence-report-event-cross-origin-urn-iframe.https.html 
-    /fenced-frame/fence-report-event-cross-origin.https.html 
+    /fenced-frame/fence-report-event-cross-origin.sub.https.html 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1437,12 +1437,20 @@ the API {{Window/navigator}}.{{Navigator/deprecatedReplaceInURN()}}, which allow
 macros into the [=fenced frame config/mapped url=] corresponding to a given [=urn uuid=] or
 {{FencedFrameConfig}}.
 
+Note: To help with ease of adoption,
```
qingxinwu, line 1440: Does this one have a different plan from deprecatedReplaceInURN()? The note above says until 2026, and here it says third party cookie deprecation. If they have the same plan, we should probably update both, or possibly a single note paragraph to cover both the two APIs?

JensenPaul: They're different timelines, hence the different wording and different links.

qingxinwu (approved):

qingxinwu:
LGTM

domfarolino:
I saw this in my inbox but I noticed I was not added as a reviewer. Should I take a look? Feel free to add me as a reviewer and I will take a look after I'm back from vacation.

spec.bs
```diff
@@ -1505,6 +1513,46 @@ partial interface Navigator {
...
+  1. Let |p| be [=a new promise=].
+
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |mapping| be |global|'s [=associated Document=]'s [=node navigable=]'s
+        [=navigable/traversable navigable=]'s [=traversable navigable/fenced frame config mapping=].
+
+     1. Let |config| be the result of [=fenced frame config mapping/finding a config=] in |mapping|
```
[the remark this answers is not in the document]
JensenPaul: Does that mean callers have to wait for a Protected Audience auction to complete before calling `deprecatedURNtoURL()`?
It does look like we require the mapping to be finalized: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/interest_group/ad_auction_service_impl.cc;l=487;drc=2b8b8c15ea96feadd34e04ca947e5f05069e3b24
Should I just change this to instead say this?
```
    1. If |mapping|'s [=fenced frame config mapping/finalized config mapping=][|urn|], does not
       [=map/exist=], [=queue a global task=] on the [=DOM manipulation task source=]
       given |global|, to [=resolve=] |p| with {{undefined}}, and abort these steps.
    1. Let |config| be |mapping|'s [=fenced frame config mapping/finalized config mapping=][|urn|].
```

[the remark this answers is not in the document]
gtanzer: It's not an issue for PA because the pending urn is never returned to script, so it can't be passed to deprecatedURNtoTURL; instead the promise waits for the auction result to be resolved. In selectURL it immediately resolves to a pending urn, so you could pass it as input (and it won't work).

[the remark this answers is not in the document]
JensenPaul: So what's the right thing to put in the spec?  it sounds like it only works with finalized URNs, in which case I think my suggestion in my last comment is accurate?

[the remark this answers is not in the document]
gtanzer: Yeah, afaict what you put in the comment matches the implementation.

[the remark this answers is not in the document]
JensenPaul: Okay, I switched to my suggested steps.

spec.bs
```diff
@@ -1505,6 +1513,46 @@ partial interface Navigator {
...
+        [=map/exist=], [=queue a global task=] on the [=DOM manipulation task source=]
+        given |global|, to [=resolve=] |p| with {{undefined}}, and abort these steps.
+
+     1. Let |config| be |mapping|'s [=fenced frame config mapping/finalized config mapping=][|urn|].
+
+     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |global|, to
+        [=resolve=] |p| with |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
+
+     1. If |send_reports| is true, run steps in |config|'s
+        [=fenced frame config/on navigate callback=].
```
domfarolino, line 1547: Why run these just on lookup? I assume that's what the imple does? It just feels a little strange.

domfarolino (commented):
Pretty much lgtm; nothing major here.

spec.bs
JensenPaul: `deprecatedURNtoURL()` was added eons ago and the motivation was stated in https://github.com/WICG/turtledove/blob/main/Proposed_First_FLEDGE_OT_Details.md#advertisement-rendering as: "the `sendReports` parameter...is included to [promote accurate measurement](https://github.com/WICG/turtledove/blob/main/Proposed_First_FLEDGE_OT_Details.md#promoting-accurate-measurement-of-fledge-effectiveness.) and comparison of FLEDGE API utility by allowing auction winning URLs to be treated and used as they are in today’s ad auctions."
Basically we wanted a way to test Protected Audience (especially reporting scripts and APIs) when the URL coming from `deprecatedURNtoURL()` was rendered in a some other way that's not visible to the browser (so the browser doesn't know when to trigger Protected Audience reporting without the hint from `sendReports`).

spec.bs
```diff
@@ -1505,6 +1513,46 @@ partial interface Navigator {
...
+
+  1. If |urn| is <span class=XXX>TODO invalid</span>, [=exception/throw=] a {{TypeError}}.
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |mapping| be |global|'s [=associated Document=]'s [=node navigable=]'s
```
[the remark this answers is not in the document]
JensenPaul: I think I fixed this.  Please take a look.

spec.bs
```diff
@@ -1505,6 +1513,46 @@ partial interface Navigator {
   </wpt>
 </div>
 
+<div algorithm>
+  The <dfn method for=Navigator>deprecatedURNtoURL(|urnOrConfig|, |send_reports|)</dfn>
+  method steps are:
```
domfarolino, line 1518: All of this can be wrapped like so:

```
  The <dfn method for=Navigator>deprecatedURNtoURL(|urnOrConfig|, |send_reports|)</dfn> method
  steps are:
```

and be compliant with 100 lines.

spec.bs
domfarolino: Any chance there's a WPT for this?

spec.bs
```diff
@@ -1505,6 +1513,46 @@ partial interface Navigator {
...
+
+  1. Let |urn| be null.
+
+  1. If |urnOrConfig| is a {{USVString}}, set |urn| to |urnOrConfig|.
+
+  1. Otherwise, set |urn| to |urnOrConfig|'s [=fencedframeconfig/urn=].
+
+  1. If |urn| is <span class=XXX>TODO invalid</span>, then [=exception/throw=] a {{TypeError}}.
```
domfarolino, line 1526: https://github.com/WICG/fenced-frame/pull/161#discussion_r1626061123 got marked outdated, so I guess I'll re-file this comment as a stand-in for it.

domfarolino (commented):
Btw what is the motivation to include this API in this spec, instead of the PA spec? Impl-wise, it looks like it was included closer to the PA stuff: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ad_auction/navigator_auction.idl;l=41;drc=ac83a5a2d3c04763d86ce16d92f3904cc9566d3a;bpv=1;bpt=0

spec.bs
JensenPaul: I proposed some new words here, WDYT?

spec.bs
JensenPaul: It looks like we don't have WPTs for when `sendReports` is `true`, but I should note that `deprecatedURNtoURL()` is [going away at 3PCD](https://developers.google.com/privacy-sandbox/relevance/protected-audience-api/feature-status#:~:text=Note%3A%20The%20temporary%20navigator.deprecatedURNToURL()%20will%20be%20removed%20by%20third%2Dparty%20cookie%20deprecation.).

JensenPaul:
> Btw what is the motivation to include this API in this spec, instead of the PA spec? Impl-wise, it looks like it was included closer to the PA stuff: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ad_auction/navigator_auction.idl;l=41;drc=ac83a5a2d3c04763d86ce16d92f3904cc9566d3a;bpv=1;bpt=0

I feel like stuff that only relates to URNs, like `deprecatedURNtoURL()`, should be in the FF spec, like `deprecatedReplaceInURN()` already is.  The use of URNs (and then FFConfig) seems tied to seeding FFs, and could be used by specs other than PA.

domfarolino:
Ah darn it looks there are conflicts now that the other PR merged. Would you be able to sort them out please?

JensenPaul:
> Ah darn it looks there are conflicts now that the other PR merged. Would you be able to sort them out please?

I fixed the merge conflict.  Now it looks like the build is failing on an unrelated line:
`LINE 1724: Couldn't find WPT test 'fenced-frame/fence-report-event-cross-origin.https.html' - did you misspell something?`
Looks related to https://chromium-review.googlesource.com/c/chromium/src/+/5544804
So I updated the WPT file name.

spec.bs
```diff
@@ -1507,7 +1516,47 @@ partial interface Navigator {
...
+  The <dfn method for=Navigator>deprecatedURNtoURL(|urnOrConfig|, |send_reports|)</dfn> method
+  steps are:
+
+  1. Let |urn| be null.
+
+  1. If |urnOrConfig| is a {{USVString}}, set |urn| to |urnOrConfig|.
+
+  1. Otherwise, set |urn| to |urnOrConfig|'s [=fencedframeconfig/urn=].
 
+  1. If |urn| is not a valid [=urn uuid=] (i.e. won't pass the ABNF in Section 3 of
```
domfarolino, line 1528:
```suggestion
  1. If |urn| is not a valid [=urn uuid=] (i.e., won't pass the ABNF in Section 3 of
```

nit: https://whatwg.org/style-guide#grammar

spec.bs
domfarolino: I think that sounds right? Although it would be inelegant to have to write that `i.e.,` each time we reference the concept of an invalid URN. But I suppose this is an incremental improvement so that's fine.

domfarolino (approved):
(Meant to "approve" with nits, whoops)

NameNoQuality (approved):

PR #162, opened 2024-05-21
Fix TODO for calling "on navigate callback"

JensenPaul:
***
<a href="https://pr-preview.s3.amazonaws.com/JensenPaul/fenced-frame/pull/162.html" title="Last updated on May 31, 2024, 6:51 PM UTC (53bac09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/162/e67a4f9...JensenPaul:53bac09.html" title="Last updated on May 31, 2024, 6:51 PM UTC (53bac09)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1825,9 +1825,6 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
      cases where a [=navigate|navigation=] to a [=top-level traversable=] does not originate from a
      <{fencedframe}>.
 
-  1. If |eventType| is <code>[=automatic beacon event type/reserved.top_navigation_start=]</code>,
-     run steps in |config|'s [=fenced frame config/on navigate callback=].
-
   1. Let |beacon data| be |sourceSnapshotParams|'s [=source snapshot params/automatic beacon data
      map=][|eventType|].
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2837,9 +2840,6 @@ content in the <{fencedframe}> or its embedder, exactly one of two things will h
   :: |automaticBeaconsAllowed|.
 </div>
 
-<span class=XXX>TODO: Call the [=fenced frame config instance/on navigate callback=] at the
```
gtanzer, line 2840: The `on navigate callback` should be called when the urn/config is loaded into a fenced frame, not when there is a top-level navigation.

JensenPaul: Fixed.  I misinterpreted Fenced_Frames_Ads_Reporting.md to mean that `reserved.top_navigation_start` is sent on initial ad navigation.

spec.bs
```diff
@@ -2739,6 +2739,8 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
       1. Set |url| to |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
 
+      1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
```
domfarolino, line 2742:
```suggestion
      1. Run the steps in |config|'s [=fenced frame config/on navigate callback=].
```

domfarolino (approved):
This looks good to me. The only usage of this appears to be https://wicg.github.io/turtledove/#ref-for-fenced-frame-config-on-navigate-callback%E2%91%A0, and https://wicg.github.io/turtledove/#asynchronously-finish-reporting indeed appears to expect to run in parallel, which is good. Just putting this down in writing for future archaeology if necessary.

domfarolino:
Thank you!

PR #163, opened 2024-06-03
Spec adAuctionComponents

HabibiYou:
Addresses https://github.com/WICG/turtledove/issues/1174

***
<a href="https://pr-preview.s3.amazonaws.com/HabibiYou/fenced-frame/pull/163.html" title="Last updated on Jun 4, 2024, 2:56 PM UTC (162281a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/163/e67a4f9...HabibiYou:162281a.html" title="Last updated on Jun 4, 2024, 2:56 PM UTC (162281a)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2804,15 +2804,10 @@ finalized.
   : <dfn for="navigation params">fenced frame config instance</dfn>
   :: A [=fenced frame config instance=] or null, initially null.
 
-  : <dfn for="navigation params">ad auction components</dfn>
-  :: A [=list=] of [=fencedframeconfig/urn=]s or null, initially null.
-
-     Note: These members are only set for embedder-initiated <{fencedframe}> navigations,
-     and if a new {{Document}} is created as a result of such a navigation, these members are transferred
-     to the new [=fenced navigable container/fenced navigable=]'s [=navigable/active browsing context=]. 
-     In that case they become the new [=fenced navigable container/fenced navigable=]'s 
-     [=navigable/active browsing context=]'s [=browsing context/fenced frame config instance=] and
-     [=browsing context/ad auction components=], respectively.
+     Note: This is only set for embedder-initiated <{fencedframe}> navigations, and if a new
+     {{Document}} is created as a result of such a navigation, this member is transferred to the new
+     [=fenced navigable container/fenced navigable=]'s [=navigable/active browsing context=]'s
+     [=browsing context/fenced frame config instance=] member.
 </div>
 
 <div algorithm=create-navigation-params-config-instance>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1505,6 +1506,39 @@ partial interface Navigator {
...
+    origin=], then then throw a {{DOMException}}.
+
+  1. Let |maxAdComponents| be 40.
+
+  1. If |numAdComponents| is greater than |maxAdComponents|, then set |numAdComponents| to 
+    |maxAdComponents|.
+
+  1. Let |adComponentsURNs| be an empty [=list=] of [=fencedframeconfig/urn=]s.
+
+  1. [=map/For each=] |urn| → _ of |instance|'s [=fenced frame config instance/nested
```
domfarolino, line 1529: We don't really use `_` in specs I don't think? Just make this an ignored variable I think.

spec.bs
```diff
@@ -1505,6 +1506,39 @@ partial interface Navigator {
...
+
+  1. If |instance| is null, then throw a {{DOMException}}.
+
+  1. If [=this=]'s [=relevant settings object=]'s [=environment settings object/origin=] and
+    |instance|'s [=fenced frame config instance/mapped url=]'s [=url/origin=] are not [=same
+    origin=], then then throw a {{DOMException}}.
+
+  1. Let |maxAdComponents| be 40.
+
+  1. If |numAdComponents| is greater than |maxAdComponents|, then set |numAdComponents| to
```
domfarolino, line 1524: I would just use the `&gt;` sign here directly. Will probably need to adjust the wrapping after.

spec.bs
```diff
@@ -1505,6 +1506,39 @@ partial interface Navigator {
...
+  1. Let |adComponentsURNs| be an empty [=list=] of [=fencedframeconfig/urn=]s.
+
+  1. [=map/For each=] |urn| → _ of |instance|'s [=fenced frame config instance/nested
+     configs=]:
+     
+     1. If |numAdComponents| equals 0, then [=iteration/break=].
+
+     1. [=list/Append=] |urn| to |adComponentsURNs|.
+
+     1. Decrement |numAdComponents| by 1.
```
domfarolino, line 1536:
```suggestion
     1. Set |numAdComponents| to |numAdComponents| &minus; 1.
```

spec.bs
HabibiYou: Sounds good to me.

domfarolino (commented):
Same question as https://github.com/WICG/fenced-frame/pull/161#pullrequestreview-2097119348

JensenPaul:
> Same question as [#161 (review)](https://github.com/WICG/fenced-frame/pull/161#pullrequestreview-2097119348)

I feel like `adAuctionComponents()` belongs here as it references only terms defined in the FF spec.  Were it in the PA spec, it would have to cross-reference everything from the FF spec, and would hence be a bit non sequitur and lack relevant context for readers.

domfarolino (approved):

PR #164, opened 2024-06-03
Spec "Referer" and "Origin" headers in reporting beacons.

blu25:
This PR specs the following behavior:

- Setting the "Origin" header to the origin of the frame that invoked `reportEvent()` for destination URL events.
- Setting the "Origin" header to the worklet's origin (which we are calling `reporting url declarer origin`) for destination enum and automatic beacon events.
- Setting the "Referer" header to the ad frame root's origin (which we are calling `initiator ancestor root origin`) for automatic beacons triggered by component ad frames.
- Setting the "Referer" header to the origin of the frame that triggered the beacon for all beacons triggered not by component ads.

**See [this document](https://docs.google.com/document/d/1_DtHmvb9_0VNNnQS4nLLhdIJcxWFyGxjlFgVQIDSS38/edit?usp=sharing) for notes/reasonings on why things were done the way they were.**

There will need to be a separate effort in the [Protected Audience spec](https://wicg.github.io/turtledove/) to set some of the values that these changes are expecting to be set. More specifically:

- The new `reporting url declarer origin` variable will need to be set during the [asynchronously finish reporting](https://wicg.github.io/turtledove/#asynchronously-finish-reporting) when it invokes [finalize a reporting destination](https://wicg.github.io/fenced-frame/#finalize-a-reporting-destination).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/164.html" title="Last updated on Oct 1, 2024, 9:52 PM UTC (cb7f4e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/164/c799ffe...cb7f4e5.html" title="Last updated on Oct 1, 2024, 9:52 PM UTC (cb7f4e5)">Diff</a>


domfarolino:
Is this a draft or is this ready for my review?

blu25:
> Is this a draft or is this ready for my review?

It's back to draft after changes were made during the impl's code review. I'll re-request when it's ready for you to take a look.

spec.bs
```diff
@@ -1855,7 +1844,8 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
      <{fencedframe}>.
 
   1. Let |request initiator| be |sourceOrigin| if |config|'s [=fenced frame config instance/is ad
-     component=] is false, |config|'s [=fenced frame config instance/ad root origin=] otherwise.
+     component=] is false, |sourceSnapshotParams|'s [=source snapshot params/initiator ancestor root
```
domfarolino, line 1847: add "and" before `|sourceSnapshotParams|` (I realize this problem predates this PR, but since we're touching this line anyways here...)

spec.bs
```diff
@@ -2280,6 +2268,27 @@ To mend the intended distinction between [=top-level traversables=] and [=fenced
 Note: With these new definitions, a [=top-level traversable=] is essentially "unfenced" as described
 in the [[#nested-traversables-intro]].
 
+The algorithm as it exists today is a "fenced" algorithm, which will be re-introduced as a "fenced"
```
domfarolino, line 2271: "The algorithm" --> please reference the algorithm explicitly.

spec.bs
```diff
@@ -2280,6 +2268,27 @@ To mend the intended distinction between [=top-level traversables=] and [=fenced
...
+variant as follows:
+
+<div algorithm>
+  A <dfn>fenced top-level traversable</dfn> is a [=traversable navigable=] with a null
+  [=navigable/parent=].
+</div>
+
+<div algorithm>
+  To get the <dfn noexport for="navigable">fenced top-level traversable</dfn> of a [=navigable=]
```
domfarolino, line 2280: Is `noexport` necessary here? It sounds like dfns are not exported by default: https://speced.github.io/bikeshed/#dfn-export

spec.bs
domfarolino, line 2271: Wait, why are we doing this anyways? None of this seems to be mentioned in the PR's OP.

Before this PR, existing algorithms in HTML that were non-prefixed, were "fenced" by default, and we introduced unfenced versions in our spec. But now I guess we're re-introducing (only one or two of them) as explicitly "fenced", and then I guess we also have our other "unfenced" version that our spec has introduced previously?

Mayybeee this is important, but I am not initially sold.

spec.bs
```diff
@@ -2280,6 +2268,27 @@ To mend the intended distinction between [=top-level traversables=] and [=fenced
...
 
+The algorithm as it exists today is a "fenced" algorithm, which will be re-introduced as a "fenced"
+variant as follows:
+
+<div algorithm>
+  A <dfn>fenced top-level traversable</dfn> is a [=traversable navigable=] with a null
+  [=navigable/parent=].
+</div>
+
+<div algorithm>
```
domfarolino, line 2279: Also if we're going to do this I think we should just say we're changing the HTML algorithm's name. No need to copy its text exactly, right?

spec.bs
```diff
@@ -2280,6 +2268,27 @@ To mend the intended distinction between [=top-level traversables=] and [=fenced
 Note: With these new definitions, a [=top-level traversable=] is essentially "unfenced" as described
 in the [[#nested-traversables-intro]].
 
+The algorithm as it exists today is a "fenced" algorithm, which will be re-introduced as a "fenced"
+variant as follows:
+
+<div algorithm>
+  A <dfn>fenced top-level traversable</dfn> is a [=traversable navigable=] with a null
```
domfarolino, line 2275: Like the below, I'm not sure this is necessary.

spec.bs
```diff
@@ -2603,6 +2644,9 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   : <dfn for="source snapshot params">initiator fenced frame config instance</dfn>
   :: a [=fenced frame config instance=] or null, initially null.
 
+  : <dfn for="source snapshot params">initiator ancestor root origin</dfn>
+  :: an [=origin=] or null, initially null.
```
domfarolino, line 2648: Let's add a Note about what this actually is in practice.

spec.bs
```diff
@@ -947,11 +959,11 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
        beacons](https://wicg.github.io/attribution-reporting-api/#ref-for-request-service-workers-mode).
 
     : [=request/origin=]
-    :: <span class=XXX>Get the origin from somewhere, like the [implementation
-       does](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/fenced_frame/fenced_frame_reporter.cc;l=384;drc=183dd920ea5d6aed92e007c40e226f1f840c0567)</span>.
+    :: |request initiator| if |event| is a [=fencedframetype/destination URL event=],
+       |destination info|'s [=reporting destination info/reporting url declarer origin=] otherwise.
```
domfarolino, line 963: "and" before `|destination info|'s` (might need to re-wrap).

domfarolino, line 963: What about when this origin is null?

spec.bs
```diff
@@ -947,11 +959,11 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
        beacons](https://wicg.github.io/attribution-reporting-api/#ref-for-request-service-workers-mode).
 
     : [=request/origin=]
-    :: <span class=XXX>Get the origin from somewhere, like the [implementation
-       does](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/fenced_frame/fenced_frame_reporter.cc;l=384;drc=183dd920ea5d6aed92e007c40e226f1f840c0567)</span>.
+    :: |request initiator| if |event| is a [=fencedframetype/destination URL event=],
```
domfarolino, line 962: It's not obvious to me why it's important (or even safe) to change the request's origin based on this criteria. Can you add a note explaining why?

spec.bs
```diff
@@ -727,12 +727,18 @@ A <dfn export for=fencedframetype>pending event</dfn> is a [=struct=] with the f
...
+  : <dfn>request initiator</dfn>
+  :: an [=origin=]
 </dl>
 
 A <dfn export for=fencedframetype>reporting destination info</dfn> is a [=struct=] with the
 following [=struct/items=]:
 
 <dl export dfn-for="reporting destination info">
+  : <dfn>reporting url declarer origin</dfn>
+  :: null, or an [=origin=]
```
domfarolino, line 740: When is this null? Maybe I missed it but doesn't seem to be set to null anywhere.

spec.bs
```diff
@@ -947,11 +959,11 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
...
 
     : [=request/origin=]
-    :: <span class=XXX>Get the origin from somewhere, like the [implementation
-       does](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/fenced_frame/fenced_frame_reporter.cc;l=384;drc=183dd920ea5d6aed92e007c40e226f1f840c0567)</span>.
+    :: |request initiator| if |event| is a [=fencedframetype/destination URL event=],
+       |destination info|'s [=reporting destination info/reporting url declarer origin=] otherwise.
 
     : [=request/referrer=]
-    :: `"no-referrer"`
+    :: |request initiator|
```
domfarolino, line 966: This is also a big change. Do we really perform zero referrer policy redaction?

spec.bs
blu25: On [the implementation side](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_reporter.h;l=371-378;drc=27da6a274be3aaa246d3397d3bd8595f581e2820) of things, the optional value is serving the same purpose that the [pending events or reporting destination info in the fenced frame reporting map](https://wicg.github.io/fenced-frame/#fencedframetype-fenced-frame-reporting-map) is doing. Namely, marking whether the metadata has been finalized yet. There's no need to have it be null in the spec, so I'll make it just an origin.

spec.bs
blu25: I'll add a note to this.

[See this part of the design doc for a more detailed explanation](https://docs.google.com/document/d/1gKRZ9g_X5HCZbW__GinViwcC8iWEai_kBXsiJCXrt80/edit#bookmark=id.y3ru3k6q9pvv).

spec.bs
blu25: There's a note further down that explains this field along with `initiator fenced frame config instance` and `target fenced frame config`. I've moved that block up to be closer to these fields.

spec.bs
blu25: Resolving since `reporting url declarer origin` is now just an origin, and can't be set to null.

spec.bs
blu25: The intention for this is to have an algorithm to get the "main frame" as well as get the "outermost main frame" (as it's called in the Chromium impl). The initial top-level traversable patch turns it into an "outermost main frame", but there's nothing left over to traverse up a frame tree until you reach the fenced frame root.

This is needed so that ad components sending automatic beacons can use the origin of the ad frame root. There are a couple alternatives to what I have here:

1. Turn this into a "get the ancestor fenced navigable" algorithm, and move it to the [nested traversables](https://wicg.github.io/fenced-frame/#nested-traversables-inner) section. Remove the `fenced top-level traversable` definition altogether.
2. Even though the implementation looks directly at the frame's origin rather than at the fenced frame config when determining the origin of the ad frame root, I think we could just put the ad root's origin in the fenced frame config for ad components. While a cross-origin navigation to the ad root would make that information out of sync, the cross-origin navigation will result in the new page losing access to ad components + the ad components being unloaded, so realistically there will be no behavioral differences between the spec and implementation.

Let me know which one you'd prefer.

spec.bs
blu25: If by redaction you're referring to the "referrer policy", we do set it to ["origin"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-origin) in the impl, which should be reflected in the spec (I'll add that in). Or is this referring to something else?

domfarolino: Redaction as in allow a referrer policy to limit the contents of whatever "referrer" is eventually present on the request. "no-referrer" means the https://w3c.github.io/webappsec-referrer-policy/ performs no redaction

domfarolino:
Any idea why the latest preview does not seem to be updated? In the preview ["declarer origin" can be null](https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/164.html#reporting-destination-info-reporting-url-declarer-origin) but I don't think that matches the spec text, right?

spec.bs
blu25: I had forgotten to update that. We redact it to [just show the origin](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_reporter.cc;l=699;drc=5b6ee2e0c7eca5b5c5c26f59349f4c1daeb7de1e), rather than having zero redaction. It should be reflected in the spec now (and in the preview once I figure out why it's not updating).

spec.bs
blu25: I think this was referring to the `noexport` comment? I'm going to resolve this but please re-open if this is referring to something else.

spec.bs
blu25: I'm thinking about this some more, and I don't think option 1 is feasible, since it'll become really awkward where it's being used. A couple more options:

3. Have "top-level traversable" take an optional "fenced" parameter that defaults to "false" (so this continues to be unfenced by default). But that still will be awkward to write out.
4. Write a "get the initiator ancestor root origin" algorithm and set `[=source snapshot params/initiator ancestor root origin=]` to the result of running that function. The algorithm will do the traversal manually. I think this will also be easier to understand, and we won't have to worry about making a distinction between fenced/unfenced top-level traversables; we can just keep all top-level traversables unfenced.

I'm going to go ahead and do option 4 since I think it'll be the cleanest way to handle this.

blu25:
> Any idea why the latest preview does not seem to be updated? In the preview ["declarer origin" can be null](https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/164.html#reporting-destination-info-reporting-url-declarer-origin) but I don't think that matches the spec text, right?

I'm not sure. It seems to be working again since I uploaded the latest patch, so you should be good to look at the preview/diff.

spec.bs
```diff
@@ -2660,6 +2696,32 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+  1. While |navigable| is not null and |navigable| is not a [=fenced navigable container/fenced
+     navigable=]:
+  
+    1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+  1. If |navigable| is null, return null.
+
+  1. Otherwise:
+
+    1. Set |navigable| to |navigable|'s [=navigable/unfenced parent=].
```
domfarolino, line 2713: Assert that we have a fenced navigable before this line, just for clarity.

spec.bs
```diff
@@ -2660,6 +2696,32 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+
+    1. While |navigable| is not null and |navigable| is not a [=fenced navigable container/fenced
+      navigable=]:
+
+      1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+    1. If |navigable| is null, return null.
+
+       Otherwise, return |navigable|'s associated {{Document}}'s [=Document/origin=].
```
domfarolino, line 2722: New step number here, for consistency with the last "Otherwise"

domfarolino, line 2722: Link between navigable -> Document is unclear, since what you linked to `{{Document}}` is not a member of navigable.

spec.bs
```diff
@@ -2660,6 +2696,32 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   [=automatic beacon data/once=] set to true.
 </div>
 
+<div algorithm=get-initiator-ancestor-root-origin>
```
domfarolino, line 2699: Let's add a note or an example to the bottom of this algorithm explaining what it actually does / what level of the tree it targets. From my reading, if `|sourceDocument|` is:

```
Main frame <-- origin A
  <iframe> <-- origin B
    <fencedframe> <-- origin C
      <iframe> <-- origin D
       <iframe> <-- origin E
          |sourceDocument|
```

... then this algorithm will return origin A, right?

And if `|sourceDocument|` is:

```
<fencedframe> <-- origin A
  <iframe> <-- origin B
    <fencedframe> <-- origin C
      {{sourceDocument}}
```

... it will return A in this case too, right?

So basically it returns the nearest fenced frame's unfenced parent's fenced parent's origin? I think?

----

Separate question: is there a guarantee that this will be called on a Document that is somewhere inside of a fenced frame tree?

blu25: > So basically it returns the nearest fenced frame's unfenced parent's fenced parent's origin? I think?

Yes. The main use case for this is to get the origin of the fenced frame root whose `FencedFrameConfig` created the ad component that `|sourceDocument|` lives in. I've added a note to the algorithm.

> Separate question: is there a guarantee that this will be called on a Document that is somewhere inside of a fenced frame tree?

The returned value will only be used inside of a fenced frame tree, but since this is being set in the source snapshot params, it can be called on a Document outside of a fenced frame tree. The `If |navigable| is null, return null.` early return step will account for that.

spec.bs
```diff
@@ -2660,6 +2696,38 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+
+        1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+     1. If |navigable| is null, return null.
+
+     1. Otherwise, return |navigable|'s [=navigable/active document=]'s [=Document/origin=].
+
+  Note: This algorithm is used to get the [=origin=] of the [=fenced navigable container/fenced
+  navigable=] whose descendant [=fenced navigable container/fenced navigable=]'s descendant is
+  |sourceDocument|, if one exists.
```
domfarolino, line 2728: I think this note is kind of confusing, sorry. Instead of targeting from top->bottom (i.e., "descendant's descendant's [...]) can you describe what this is supposed to return given a fenced frame? I.e., center the language around bottom->top. So given a fenced frame, will the return origin always come from a fenced frame ancestor? Or can it come from a main frame? When are the "If navigable is null" conditions true and how does that effect the ancestor chain. Maybe even provide an example or two like I did in the comment I left about this algorithm. It's a tricky one so I want to make sure we're as expressive as possible so we don't have to think through this again.

blu25: Added a non-normative explanatory section with examples.

spec.bs
```diff
@@ -2741,6 +2777,59 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   [=automatic beacon data/once=] set to true.
 </div>
 
+<div algorithm=get-initiator-ancestor-root-origin>
+  To <dfn>get the initiator ancestor root origin</dfn> of a given {{Document}} |sourceDocument|:
+
+  1. Let |navigable| be |sourceDocument|'s [=node navigable=].
+
+  1. While |navigable| is not null and |navigable| is not a [=fenced navigable container/fenced
```
domfarolino, line 2785: When would `|navigable|` be null? First, if we call this on a detached document I guess, but navigations cannot be initiated there so I don't think that would cause it. Second... if we called this algorithm on a document that has nothing to do with a fenced frame tree, then I guess we'd just reach the top-level traversable's parent (null) and that's when we'd get kicked out of this loop with a null navigable, right?

And that's only possible because we invoke this algorithm indiscriminately for all navigations? Do I have that right?

spec.bs
```diff
@@ -2741,6 +2777,59 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+     navigable=]:
+  
+     1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+  1. If |navigable| is null, return null.
+
+  1. Otherwise:
```
domfarolino, line 2792: I don't think this is needed? We can just de-dent all of the below steps and put them right below without an "Otherwise", since the prior steps returns. Right?

spec.bs
```diff
@@ -2741,6 +2777,59 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+     1. While |navigable| is not null and |navigable| is not a [=fenced navigable container/fenced
+        navigable=]:
+
+        1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+     1. If |navigable| is null, return null.
+
+     1. Otherwise, return |navigable|'s [=navigable/active document=]'s [=Document/origin=].
+
+  *The rest of this section is non-normative*.
```
domfarolino, line 2807: I'd get rid of this line, and just wrap the entire thing below in an example div, and put the HTML-ish-stuff below in a pre-html tag. See https://wicg.github.io/fenced-frame/#fenced-frame-src-example for an example. I think that would be more clear.

spec.bs
```diff
@@ -2741,6 +2777,59 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   [=automatic beacon data/once=] set to true.
 </div>
 
+<div algorithm=get-initiator-ancestor-root-origin>
+  To <dfn>get the initiator ancestor root origin</dfn> of a given {{Document}} |sourceDocument|:
```
domfarolino, line 2781: So this algorithm looks fine, but can we rename it something more specific? Maybe something super verbose like `get the ·initiator's fenced root's fenced root's origin`? I know that's wordy but it is crystal clear. WDYT? That or `get the initiator's fenced grand-ancestor's origin`.

spec.bs
```diff
@@ -2741,6 +2777,59 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+
+        1. Set |navigable| to |navigable|'s [=navigable/parent=].
+
+     1. If |navigable| is null, return null.
+
+     1. Otherwise, return |navigable|'s [=navigable/active document=]'s [=Document/origin=].
+
+  *The rest of this section is non-normative*.
+
+  Given a {{Document}} embedded inside of a <{fencedframe}>, this algorithm gets the origin
```
domfarolino, line 2809: Allow me to recommend a re-word:

```
Given a Document embedded inside of a <fencedframe>,
this algorithm gets the origin of the document's nearest
fenced frame ancestor's nearest fenced frame ancestor.
If no such "fenced frame grand-ancestor" exists, the algorithm
returns null. Therefore, this algorithm only returns an origin
when given a Document that is, at a minimum, in a fenced frame
which is itself in a fenced frame. For example, for the given frame
tree structure:
```

blu25: Reworded

spec.bs
blu25: Changed to `get the initiator's fenced grand-ancestor's origin`.

spec.bs
blu25: Wrapped the section in an example div.

spec.bs
blu25: We can do away with this. I'm going to keep the other "otherwise" on the last step for readability (since it doesn't result in any indentations). Let me know if you disagree with that decision and I can remove that as well.

spec.bs
blu25: > if we called this algorithm on a document that has nothing to do with a fenced frame tree, then I guess we'd just reach the top-level traversable's parent (null) and that's when we'd get kicked out of this loop with a null navigable

> that's only possible because we invoke this algorithm indiscriminately for all navigations

Yes, that's why this is in place. This will also happen for fenced frames that don't have a fenced frame ancestor, since it will cross the fenced boundary, then traverse up the non-fenced tree until it reaches the root.

blu25:
Let's hold off on merging this until the latest discussion in the [design doc](https://docs.google.com/document/d/1gKRZ9g_X5HCZbW__GinViwcC8iWEai_kBXsiJCXrt80/edit?usp=sharing) is addressed.

spec.bs
```diff
@@ -2799,34 +2821,32 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
     fenced frame. For example, for the given frame tree structure:
 
     ```
-    Main frame (origin A)
-      <fencedframe> (origin B)
-        <iframe> (origin C)
-          <fencedframe> (origin D)
-            <iframe> (origin E)
-              <fencedframe> (origin F)
+    Main frame (A)
```
domfarolino, line 2834: Keeping "origin A" here (instead of just "A") seems fine? Not sure why it was removed.

spec.bs
```diff
@@ -974,8 +1000,9 @@ A <dfn for=fencedframetype>destination event</dfn> is either a
 
 <div algorithm>
   In order to <dfn>report an event</dfn> using a [=fencedframetype/fenced frame reporter=]
-  |reporter| with a {{FenceReportingDestination}} |destination|, and a
-  [=fencedframetype/destination event=] |event|, run these steps:
+  |reporter| with a {{FenceReportingDestination}} |destination|, an [=origin=] |request initiator|,
+  a [=referrer policy=] |initiator referrer policy| and a [=fencedframetype/destination event=]
```
domfarolino, line 1004:
```suggestion
  a [=referrer policy=] |initiator referrer policy|, and a [=fencedframetype/destination event=]
```

spec.bs
blu25: The algorithm isn't returning the origin anymore. I guess it's fine either way. I'll change it back to origin since the rest of the comments still makes it clear this is returning a navigable now.

domfarolino (approved):

PR #169, opened 2024-07-10
Spec network revocation mechanism to disable fetches

gtanzer:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/169.html" title="Last updated on Nov 1, 2024, 7:06 PM UTC (f5b6b6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/169/23aed62...f5b6b6c.html" title="Last updated on Nov 1, 2024, 7:06 PM UTC (f5b6b6c)">Diff</a>


spec.bs
```diff
@@ -1699,6 +1715,83 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |fencedFrameNonce| be |instance|'s [=fenced frame config instance/partition nonce=].
+
+     1. Let |credentiallessNonce| be
+
+     Issue: the page credentialless nonce
+     (<a href="https://github.com/WICG/fenced-frame/issues/151">WICG/fenced-frame#151</a>)
```
domfarolino, line 1744: It is not clear to me how this TODO relates to credentialless.

domfarolino (commented):
Should this PR be marked as a draft for now? Or is it ready for review? It does not look ready, but I'm just double checking.

gtanzer:
@domfarolino Is there a way to explicitly mark it draft? I didn't request a review. I'll just add WIP to the title.

gtanzer:
@domfarolino I think the content is "done" now, % resolving some links. But it would be good if you could take a look at it first so I don't resolve links that will be deleted/moved around later.

The part I'm least sure about is where to define the extra objects on the user agent. In the fetch spec, it defines the connection pool on the "user agent" which isn't a formal object, so I don't know if I should monkeypatch these definitions near there or just keep them in here.

spec.bs
```diff
@@ -1699,6 +1715,99 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        [=resolve=] |p| with {{undefined}}.
+
+  1. Return |p|.
+</div>
+
+A user agent has an associated <dfn>network revocation nonce set</dfn>, which is a [=set=] of
+[=partition nonces=], and a <dfn>network revocation exemption map</dfn>, which is a [=map=] whose
+[=map/keys=] are [=partition nonces=] and [=map/values=] are [=sets=] of [=URLs=].
+
+Note: The network revocation exemption map is used only for web platform tests; in normal usage
```
blu25, line 1767: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176/files/7c89c4edea5d948e98efb3652f6a0b5b7c1c673d#r1748939872):

> Link to the map.

spec.bs
```diff
@@ -1699,6 +1715,99 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+it is always empty.
```
blu25, line 1768: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176/files/7c89c4edea5d948e98efb3652f6a0b5b7c1c673d#r1748939779):

> This is weird and sounds maybe not good to include then. Please clarify in this thread for me. How can UAs populate this map? Is there a WPT/web driver RFC somewhere? Or is it expected to just be through command line. If so, let's clarify that in the spec.

spec.bs
```diff
@@ -1699,6 +1715,99 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+<div algorithm>
+  In order to <dfn>revoke network for a partition nonce</dfn> using a
```
blu25, line 1771: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176/files/7c89c4edea5d948e98efb3652f6a0b5b7c1c673d#r1748939939):

```suggestion
  To <dfn>revoke network for a partition nonce</dfn> using a
```

> (might need to re-wrap after this suggestion, since the line might be pretty short).

spec.bs
```diff
@@ -1699,6 +1715,99 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  [=fenced frame config instance/partition nonce=] |nonce|, run these steps:
+
+  1. [=set/Append=] |nonce| to the user agent's [=network revocation nonce set=].
+
+  1. <span class=XXX>TODO: Cancel requests in progress.</span>
```
blu25, line 1776: copied from [this comment](https://github.com/WICG/fenced-frame/pull/176/files/7c89c4edea5d948e98efb3652f6a0b5b7c1c673d#r1748943317):

> Can we use https://fetch.spec.whatwg.org/#concept-fetch-group-terminate here?
> 
> For future reference, if infra like "fetch group" didn't exist, this would be too big of a bug to just have in an anonymous TODO buried in this algorithm. We'd want to file Fetch bugs and see if there are other want-to-be consumers of this desired infra, and discuss the option of adding it, or at the very least, file a bug against our own repo to do all of that.

spec.bs
```diff
@@ -1699,6 +1715,99 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+</div>
+
+<div algorithm>
+  To determine whether fetching a [=request=] |request| <dfn>should be blocked due to a revoked
```
blu25, line 1780: copied from [this comment]():

```suggestion
  To determine whether fetching a [=request=] |request| <dfn>must be blocked due to a revoked
```

> Let's use "must" here? See https://github.com/whatwg/fetch/issues/1714.

spec.bs
blu25: From https://github.com/WICG/fenced-frame/pull/169, copied comment there as well.

blu25: Updated to link to new issue.

spec.bs
blu25: I think we can use fetch groups, but I'm a little concerned by the part that checks for "[keepalive](https://fetch.spec.whatwg.org/#request-keepalive-flag) is false". I guess if a request is done, there shouldn't be any risk of further information going out, right? So I'll replace this with a call to terminate the fetch group.

spec.bs
blu25: Right now our internal WPT has a `window.internals.exemptUrlFromNetworkRevocation` helper method, but there is nothing external facing in place yet. From my understanding, the end goal is to add something to the WPT web driver to allow exemptions. I'll call that out and add an issue to track that effort.

domfarolino (commented):
Wow it looks like I forgot to send this review from a while ago, sorry! Sending now.

spec.bs
```diff
@@ -1819,6 +1826,105 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
 
+<div algorithm>
+  The <dfn method for=Fence>disableUntrustedNetwork()</dfn> method steps are:
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
```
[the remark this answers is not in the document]
blu25: Hmm...I actually don't think this will ever be hit. The [fence getter steps](https://wicg.github.io/fenced-frame/#dom-window-fence) will always return a null fence if the browsing context's ff config instance is null, so it would never even be able to access this algorithm. So this could either be an assert, or just no check since we're guaranteed to have a fenced frame config instance by this point. This also applies to the other fence algorithms (since they all have a similar check in place), so those can probably be removed in a follow up.

For the cases you listed:

- The URN iframe case is hit in step 5.
- Iframes nested in fenced frames are allowed to disable network access, so long as they're same origin.
- The cross-origin subframe case is hit in step 4.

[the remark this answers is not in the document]
domfarolino: > so those can probably be removed in a follow up.

Great, could you file a bug and make sure the work that would need to be done to fix that bug matches the implementation too?

domfarolino (approved):

Issue #168, opened 2024-07-08
[Spec] Spec that a fenced frame waits for nested fenced frames to disable network.


blu25:
A fenced frame should not be considered to have its network revoked until all its descendants have their network access revoked as well.

VergeA:
@blu25 Can we close this now?

blu25:
Yes, thanks for reminding me.

Issue #173, opened 2024-08-08
[Spec] Spec the behavior of fenced frame in-progress embedder-initiated navigation is canceled when both child and initiator fenced frame have network disabled.


xiaochen-z:
See https://chromium-review.googlesource.com/c/chromium/src/+/5385761.

xiaochen-z:
Blocked on #227

PR #176, opened 2024-08-14
Spec: Wait for network revocation in nested fenced frames before disabling network.

blu25:
This PR introduces a new algorithm: `Recalculate the untrusted network status of all frames`. This is called whenever a fenced frame marks its network as disabled, and checks to see if any ancestor fenced frames are now allowed to have their network access be fully revoked and gain access to unpartitioned data.

This PR modifies `disableUntrustedNetwork()` to not resolve the promise, and instead puts the promise into the fenced frame config instance to be resolved once the frame tree is considered to have its network fully revoked.

This builds off of the work in https://github.com/WICG/fenced-frame/pull/146, and this should only be merged after https://github.com/WICG/fenced-frame/pull/146 is merged.

See: issue https://github.com/WICG/fenced-frame/issues/168

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/176.html" title="Last updated on Jan 7, 2025, 9:46 PM UTC (668de28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/176/14f148d...668de28.html" title="Last updated on Jan 7, 2025, 9:46 PM UTC (668de28)">Diff</a>


spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  
+  1. Let |i| be |navigables|' [=list/size=] - 1.
+
+  1. Let |navigablesWithNetworkChildren| be an empty [=set=].
+
+  1. While |i| > 0:
+
+     1. Let |currentNavigable| be |navigables|[|i|].
+
+     1. Set |i| to |i| - 1.
```
domfarolino, line 1938: Actually, all of the index stuff is a little tricky. How about we just treat the list as a queue and use `dequeue` on it, under the while condition that the queue is not (link) empty.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+  1. Let |navigables| be |topLevelTraversable|'s [=navigable/active document=]'s
+     [=Document/inclusive descendant navigables=] with [=inclusive-dn-unfenced|unfenced=] set to
+     true.
+  
+  1. Let |i| be |navigables|' [=list/size=] - 1.
+
+  1. Let |navigablesWithNetworkChildren| be an empty [=set=].
+
+  1. While |i| > 0:
```
domfarolino, line 1934: Link to while

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        1. [=list/Empty=] |config|'s [=fenced frame config instance/on network disabled promises=].
+
+     1. Otherwise:
+
+        1. Let |ancestorFencedRoot| be |currentNavigable|'s [=navigable/unfenced parent=].
+
+        1. While |ancestorFencedRoot| is not null and |ancestorFencedRoot| is not a [=fenced
+           navigable container/fenced navigable=]:
+
+           1. Set |ancestorFencedRoot| to |ancestorFencedRoot|'s [=navigable/parent=].
```
domfarolino, line 1977: Do we have an algorithm for this somewhere? I'd be surprised if we don't already have one to get the root (itself a fenced frame) of a fenced frame tree.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        
+        1. If |ancestorFencedRoot| is not null, [=set/append=] |ancestorFencedRoot| to
```
domfarolino, line 1979: When would it be null? If we've made it thus far, then we have confirmed that `|currentNavigable|` is inside a fenced frame tree, and that tree must have a root, right?

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+           |navigablesWithNetworkChildren|.
+</div>
+
+A user agent has an associated <dfn>network revocation nonce set</dfn>, which is a [=set=] of
+[=partition nonces=], and a <dfn>network revocation exemption map</dfn>, which is a [=map=] whose
+[=map/keys=] are [=partition nonces=] and [=map/values=] are [=sets=] of [=URLs=].
+
+Note: The network revocation exemption map is used only for web platform tests; in normal usage
+it is always empty.
```
domfarolino, line 1988: This is weird and sounds maybe not good to include then. Please clarify in this thread for me. How can UAs populate this map? Is there a WPT/web driver RFC somewhere? Or is it expected to just be through command line. If so, let's clarify that in the spec.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        
+        1. If |ancestorFencedRoot| is not null, [=set/append=] |ancestorFencedRoot| to
+           |navigablesWithNetworkChildren|.
+</div>
+
+A user agent has an associated <dfn>network revocation nonce set</dfn>, which is a [=set=] of
+[=partition nonces=], and a <dfn>network revocation exemption map</dfn>, which is a [=map=] whose
+[=map/keys=] are [=partition nonces=] and [=map/values=] are [=sets=] of [=URLs=].
+
+Note: The network revocation exemption map is used only for web platform tests; in normal usage
```
domfarolino, line 1987: Link to the map.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+it is always empty.
+
+<div algorithm>
+  In order to <dfn>revoke network for a partition nonce</dfn> using a
```
domfarolino, line 1991:
```suggestion
  To <dfn>revoke network for a partition nonce</dfn> using a
```

(might need to re-wrap after this suggestion, since the line might be pretty short).

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  [=fenced frame config instance/partition nonce=] |nonce|, run these steps:
+
+  1. [=set/Append=] |nonce| to the user agent's [=network revocation nonce set=].
+
+  1. <span class=XXX>TODO: Cancel requests in progress.</span>
```
domfarolino, line 1996: Can we use https://fetch.spec.whatwg.org/#concept-fetch-group-terminate here?

For future reference, if infra like "fetch group" didn't exist, this would be too big of a bug to just have in an anonymous TODO buried in this algorithm. We'd want to file Fetch bugs and see if there are other want-to-be consumers of this desired infra, and discuss the option of adding it, or _at the very least_, file a bug against our own repo to do all of that.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+</div>
+
+<div algorithm>
+  To determine whether fetching a [=request=] <var ignore>request</var> <dfn>should request be blocked due to a revoked
```
domfarolino, line 2000:
```suggestion
  To determine whether fetching a [=request=] <var ignore>request</var> <dfn>must request be blocked due to a revoked
```

Let's use "must" here? See https://github.com/whatwg/fetch/issues/1714.

domfarolino (commented):
Haven't made it through _everything_ quite yet but getting there. This is a start for now.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
   </wpt>
 </div>
 
+<div algorithm>
+  The <dfn method for=Fence>disableUntrustedNetwork()</dfn> method steps are:
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
```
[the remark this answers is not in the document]
blu25: There are 5 places in the spec that have this issue, so I'm going to fix the other instances as well. The impl throws a [`SecurityError`](https://webidl.spec.whatwg.org/#securityerror), so I'm going to do the same here.

spec.bs
blu25: This is something that is from https://github.com/WICG/fenced-frame/pull/169. I'll add this comment to the line in that PR.

spec.bs
blu25: This is from https://github.com/WICG/fenced-frame/pull/169. I'll copy this comment over.

spec.bs
blu25: This is from https://github.com/WICG/fenced-frame/pull/169. I'll copy this comment over.

spec.bs
blu25: This is from https://github.com/WICG/fenced-frame/pull/169. I'll copy this comment over.

spec.bs
blu25: This is from https://github.com/WICG/fenced-frame/pull/169. I'll copy this comment over.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |fencedFrameNonce| be |instance|'s [=fenced frame config instance/partition nonce=].
+
+     1. Let |credentiallessNonce| be
+
+     Issue: the page credentialless nonce
+     (<a href="https://github.com/WICG/fenced-frame/issues/151">WICG/fenced-frame#151</a>)
```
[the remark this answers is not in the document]
blu25: This is from https://github.com/WICG/fenced-frame/pull/146. I'll add the comment there as well.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        promises=].
+
+     1. [=Recalculate the untrusted network status of all frames=] given |global|'s
+        [=Window/browsing context=]'s [=browsing context/top-level traversable=].
+
+  1. Return |p|.
+</div>
+
+<div algorithm>
+  To <dfn>Recalculate the untrusted network status of all frames</dfn> given a [=top-level
```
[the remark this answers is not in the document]
blu25: Reworded to `Recalculate the untrusted network status of all fenced frame descendants` to make it clear it's for all descendants of the provided top-level frame, and it's only applying to fenced frames/fenced navigables.

spec.bs
blu25: I found 4 other places that also need a link, so I've updated them as well.

spec.bs
blu25: Replaced the index stuff with queue operations.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  traversable=] |topLevelTraversable|, run these steps:
+
+  1. Let |navigables| be |topLevelTraversable|'s [=navigable/active document=]'s
+     [=Document/inclusive descendant navigables=] with [=inclusive-dn-unfenced|unfenced=] set to
+     true.
+  
+  1. Let |i| be |navigables|' [=list/size=] - 1.
```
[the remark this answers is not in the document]
blu25: Resolving since we're using queue operations instead.

spec.bs
blu25: I've replaced this with getting the document's `ancestor navigables` and checking that the list isn't empty and that the first item is a `fenced navigable`.

spec.bs
blu25: That's a good point. If we end up in a state where a fenced frame's ancestor root is null, then something has gone horribly wrong. I've removed the check.

spec.bs
```diff
@@ -1204,9 +1220,19 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
...
   : <dfn>embedder shared storage context</dfn>
   :: null, or a [=string=]
-  
+
   : <dfn>is ad component</dfn>
-  :: A [=boolean=], initially false.
+  :: A [=boolean=]
+
+  : <dfn>can disable untrusted network</dfn>
+  :: A [=boolean=]
```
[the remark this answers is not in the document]
blu25: Copied to https://github.com/WICG/fenced-frame/pull/146.

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+     [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
+
+  1. If |instance|'s [=fenced frame config instance/can disable untrusted network=] is false, then
+     [=resolve=] |p| with {{undefined}} and return |p|.
+
+  1. If the [=relevant settings object=]'s [=environment settings object/origin=] and |instance|'s
+     [=fenced frame config instance/mapped url=]'s [=url/origin=] are not [=same origin=], then
+     [=resolve=] |p| with {{undefined}} and return |p|.
```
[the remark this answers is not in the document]
blu25: Copied to https://github.com/WICG/fenced-frame/pull/146

spec.bs
```diff
@@ -1291,12 +1317,21 @@ A <dfn export>fenced frame config instance</dfn> is a [=struct=] with the follow
...
     :: |config|'s [=fenced frame config/cross-origin reporting allowed=]
+
+    : [=fenced frame config instance/can disable untrusted network=]
+    :: |config|'s [=fenced frame config/can disable untrusted network=]
+
+    : [=fenced frame config instance/untrusted network status=]
+    :: [=untrusted network status/enabled=]
+
+    : [=fenced frame config instance/on network disabled promises=]
+    :: an empty [=list=]
```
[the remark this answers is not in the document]
blu25: Added a note.

spec.bs
domfarolino: I still see the check.

spec.bs
domfarolino: First, I don't see the update you describe. Can't we just get the traversable navigable of `|currentNavigable|`, which I think is guaranteed to be a fenced frame container's fenced navigable, since we know we're in a fenced frame tree at this point?

spec.bs
```diff
@@ -1839,6 +1875,176 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+           1. [=Queue a global task=] on the [=DOM manipulation task source=] given |global|, to
+              [=resolve=] |promise| with {{undefined}}.
+
+        1. [=list/Empty=] |config|'s [=fenced frame config instance/on network disabled promises=].
+
+     1. Otherwise:
+
+        1. Let |ancestorFencedRoot| be |currentNavigable|'s [=navigable/unfenced parent=].
+
+        1. While |ancestorFencedRoot| is not null and |ancestorFencedRoot| is not a [=fenced
```
domfarolino, line 1974: Like https://github.com/WICG/fenced-frame/pull/176/files#r1748939549, I think we should get rid of the null check.

blu25: The changes should be up now.

spec.bs
blu25: The updates should be uploaded now.

A fenced frame's navigable would be a traversable navigable as well right? The intention of this is to get the next fenced frame root up the frame tree. If we ask for the fenced navigable's [traversable navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#nav-traversable), the while loop would never iterate and it would just return the same navigable because the condition would hold immediately.

We can probably do the navigable's unfenced parent's traversable navigable though.

spec.bs
blu25: That patch should be up now.

spec.bs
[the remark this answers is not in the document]
blu25: I don't think this should actually be in the spec. From [the equivalent field in the impl](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_config.h;l=664;drc=57c74b90d4a6e506099c97022377d84f523b37cb):

> Currently true in all fenced frame configs, but set to false if loaded in a urn iframe

I think what we want instead is to just check if the relevant traversable navigable is a fenced navigable to see if we're in a fenced frame tree or not. The intention is to make sure that we can't call this from within a urn iframe. I'll remove this field and add the explicit non-fenced frame check in the `disableUntrustedNetwork()` definition.

spec.bs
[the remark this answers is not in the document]
blu25: [See comment](https://github.com/WICG/fenced-frame/pull/146#discussion_r1757343807)

spec.bs
blu25: See: https://github.com/WICG/fenced-frame/pull/169#discussion_r1757584319

spec.bs
blu25: See: https://github.com/WICG/fenced-frame/pull/169#discussion_r1757547494

spec.bs
```diff
@@ -1881,6 +1928,178 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+  1. Let |p| be [=a new promise=].
+
+  1. Let |context| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=].
+
+  1. If |context| is null, then throw a {{SecurityError}} {{DOMException}}.
+
+  1. Let |instance| be |context|'s [=browsing context/fenced frame config instance=].
+
+  1. If |instance| is null, then [=resolve=] |p| with {{undefined}} and return |p|.
```
domfarolino, line 1942: Given https://github.com/WICG/fenced-frame/pull/169#discussion_r1788685341 is this also an irrelevant line that we can turn into an Assert?

spec.bs
```diff
@@ -1881,6 +1928,178 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  1. Let |global| be [=this=]'s [=relevant global object=].
+
+  1. Run the following steps [=in parallel=]:
+
+     1. Let |fencedFrameNonce| be |instance|'s [=fenced frame config instance/partition nonce=].
+
+     1. Let |credentiallessNonce| be
+
+     Issue: the page credentialless nonce
+     (<a href="https://github.com/WICG/fenced-frame/issues/191">WICG/fenced-frame#191</a>)
```
domfarolino, line 1961: Before we land this, I'd like https://github.com/WICG/fenced-frame/issues/191#issuecomment-2365349593 to be addressed so I can evaluate how much of a hole in the current spec this is. A rough design, or at least design thoughts (["concepts of a plan"](https://www.reddit.com/r/PoliticalHumor/comments/1fdzo9m/mr_trump_do_you_have_a_plan/#lightbox), if you will) will suffice.

spec.bs
```diff
@@ -1881,6 +1928,178 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+     1. Let |currentNavigable| be the result of [=queue/dequeue|dequeuing=] from |navigables|.
+
+     1. If |currentNavigable| is not a [=fenced navigable container/fenced navigable=], then
+        [=iteration/continue=].
+
+     1. Let |config| be |currentNavigable|'s [=navigable/active browsing context=]'s [=browsing
+        context/fenced frame config instance=].
+
+     1. If |config| is null, then [=iteration/continue=].
```
domfarolino, line 2001: Can you add documentation about when this might be null? Actually, it's not obvious to me under which conditions this would be hit.

spec.bs
```diff
@@ -1881,6 +1928,178 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+     1. If |config|'s [=fenced frame config instance/untrusted network status=] is [=untrusted
+        network status/disabled for this and descendant trees=], then [=iteration/continue=].
+
+     1. Let |networkCutoffReady| be true if |navigablesWithNetworkChildren| does not [=set/contain=]
```
domfarolino, line 2006: When would `|navigablesWithNetworkChildren|` already contain `|currentNavigable|`? That would require iterating over the same `|currentNavigable|` twice in this loop. Is that even possible?

spec.bs
```diff
@@ -1881,6 +1928,178 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+
+        1. Set |config|'s [=fenced frame config instance/untrusted network status=] to [=untrusted
+           network status/disabled for this and descendant trees=].
+
+        1. Let |global| be |currentNavigable|'s [=navigable/active window=].
+
+        1. For each |promise| in |config|'s [=fenced frame config instance/on network disabled
+           promises=]:
+
+           1. [=Queue a global task=] on the [=DOM manipulation task source=] given |global|, to
```
domfarolino, line 2020: I'm a little worried about this. `|global|` comes from the current nav's active window, but is it guaranteed that `|promise|` was constructed _within_ that global? For example, what if inside a fenced frame, a cross-origin descendant iframe calls `disableUntrustedNetwork()`. First, is that allowed? Second, if it's allowed, then I think we'd end up scheduling the resolution of that Promise (that was created in the cross-origin descendant global) on the global of its nearest fenced frame root (since at this point `|currentNavigable|` is guaranteed to be a fenced navigable's global).

Does this make sense / am I right about this being possible?

spec.bs
blu25: At this point, we're guaranteed to be dealing with a fenced navigable, which means that |config| is guaranteed to exist. So this can probably be removed.

spec.bs
```diff
@@ -1846,8 +1890,11 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 <div algorithm>
   The <dfn method for=Fence>getNestedConfigs()</dfn> method steps are:
 
-  1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
-     [=browsing context/fenced frame config instance=].
+  1. Let |context| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=].
+
+  1. If |context| is null, then throw a {{SecurityError}} {{DOMException}}.
+
+  1. Let |instance| be |context|'s [=browsing context/fenced frame config instance=].
```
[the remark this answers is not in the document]
blu25: Sure. I can get a separate PR up for that.

spec.bs
blu25: `|navigablesWithNetworkChildren|` is set every time a frame is determined to not be ready for network cutoff. However, it is the frame's `|ancestorFencedRoot|` which ends up being put into `|navigablesWithNetworkChildren|`. So in the while loop, this will be the first time this frame is visited. But if it has a child fenced frame whose network is not disabled yet, the frame will have already been added to this list.

I'm realizing I used the wrong removal method. I currently am dequeuing from the list. I should be popping instead, since my intention is to have this loop go in reverse DFS order to ensure that the deepest nested frames are hit first. I'll fix that.

spec.bs
blu25: Added concepts of a plan to the issue. I think we can just grab the nonce itself as part of this PR rather than having a TODO. I'll need to move this issue comment to another part of the algorithm that will need a patch in a follow up.

spec.bs
blu25: > For example, what if inside a fenced frame, a cross-origin descendant iframe calls disableUntrustedNetwork(). First, is that allowed?

No. `disableUntrustedNetwork()` throws a `TypeError` if the origin that calls the method is cross-origin to the mapped URL of the ff config instance. The promises can only come from either the fenced frame root or a same-origin descendant.

domfarolino: Well, same-origin descendant iframes have different globals, but we resolve them here on the fenced frame's global, which is bad, right?

spec.bs
```diff
@@ -480,8 +483,13 @@ attribute.
...
 
 <div algorithm=destroy>
-  When a <{fencedframe}> element is [=removed from a document=], the user agent <p class=XXX>TODO:
-  destroy the nested traversable</p>.
+  When a <{fencedframe}> element is [=removed from a document=], the user agent must run the
+  following steps:
+  
+  1. <p class=XXX>TODO: destroy the nested traversable.</p>
+
+  1. [=Recalculate the untrusted network status of all fenced frame descendants=] given the
```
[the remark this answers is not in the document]
blu25: Only if the iframe teardown results in a child fenced frame also being torn down, but that will get caught by this anyway, so we don't need to add anything for iframes.

spec.bs
```diff
@@ -1951,28 +1981,80 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+  To <dfn>Recalculate the untrusted network status of all fenced frame descendants</dfn> given a
+  [=top-level traversable=] |topLevelTraversable|, run these steps:
+
+  1. [=Assert=]: this is running [=in parallel=].
+
+  1. Let |navigables| be |topLevelTraversable|'s [=navigable/active document=]'s
+     [=Document/inclusive descendant navigables=] with [=inclusive-dn-unfenced|unfenced=] set to
+     true.
+  
+  1. Let |navigablesWithNetworkChildren| be an empty [=set=].
```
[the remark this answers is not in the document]
blu25: Yes, this is correct. The very last step in the algorithm (step 4.7.2) ensures that `|navigablesWithNetworkChildren|` will always contain fenced navigables. I wasn't sure if calling out that the set is expected to only contain certain items was a valid thing to do when declaring a variable in an algorithm directly, but since it sounds like it is, I can make that change.

spec.bs
blu25: That's correct. It essentially ensures that if some nested fenced frame isn't ready for network cutoff, all of its ancestors are marked as not being ready either.

spec.bs
```diff
@@ -849,6 +857,11 @@ a reporting destination=] through the reference it kept, it will handle all of t
 stored in the [=list=]. If the destination is never finalized, then the pending events will never
 be sent.
 
+An <dfn for=fencedframetype>untrusted network status</dfn> is either "<dfn export for='untrusted
+network status'>`enabled`</dfn>", "<dfn export for='untrusted network status'>`disabled for this
+tree`</dfn>", or "<dfn export for='untrusted network status'>`disabled for this and descendant
```
domfarolino, line 862: How about the following renames:

 - "disabled for this tree" --> "disabled for this frame". Or is there another reason "tree" was chosen? My understanding is that this is an intermediate state where network is revoked just for this frame, but not for any subframes fenced/unfenced.
 - "disabled for this and descendant tree" --> "disabled for full subtree" (means that all frames, fenced/unfenced underneath the one in question, has network disabled)

What do you think?

----

Separately, I'd appreciate you adding a note underneath indicating that the middle state is not a final state; it is just an intermediate/limbo state that we're in until .... (non-normatively enumerate the conditions under which a frame has it switch from this middle state to the state).

spec.bs
```diff
@@ -2027,6 +2115,27 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+
+<div algorithm=network-status-document-abort>
+  Add a new step to the [=abort a document=] algorithm after step 4 that says:
+
+  5. [=Recalculate the untrusted network status of all fenced frame descendants=] given <var
+     ignore>document</var>'s [=node navigable=]'s [=navigable/top-level traversable=].
+</div>
+
+<div algorithm=network-status-navigate>
+  Modify the call to [=attempt to populate the history entry's document=] in the [=navigate=]
```
domfarolino, line 2134: which call? All calls? Is it definitely correct to do this after both navigation and travesal?

spec.bs
```diff
@@ -2027,6 +2115,27 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  algorithm. Update the [=completionSteps=] to include the following new step:
+
+  2. [=Recalculate the untrusted network status of all fenced frame descendants=] given <var
```
domfarolino, line 2137: Same parallel issue exists here as above, I think. Maybe we want to, on the same line, conditionally go in parallel if we're not already in parallel.

However I still have another question: why do we have to recalculate after this specific milestone?

spec.bs
```diff
@@ -1954,28 +1985,85 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        Otherwise, [=map/set=] |promises|[|global|] to the [=list=] « |p| ».
+
+     1. [=Recalculate the untrusted network status of all fenced frame descendants=] given
+        |global|'s [=Window/browsing context=]'s [=browsing context/top-level traversable=].
 
   1. Return |p|.
 </div>
 
+<div algorithm>
+  To <dfn>Recalculate the untrusted network status of all fenced frame descendants</dfn> given a
```
domfarolino, line 2012:
```suggestion
  To <dfn>recalculate the untrusted network status of all fenced frame descendants</dfn> given a
```

domfarolino (commented):
One more issue I spotted is that the https://wicg.github.io/fenced-frame/#revoke-network-for-a-partition-nonce algorithm uses "this" which it cannot do since it is called from an in parallel context. I think we should fix that in this PR>

spec.bs
blu25: "tree" was chosen simply because that's the naming convention we used in the implementation, which is scoped to fenced frame trees.

> My understanding is that this is an intermediate state where network is revoked just for this frame, but not for any subframes fenced/unfenced.

A child iframe will also have its network access revoked at the intermediate stage, so saying "disabled for this frame" would not be correct.

Would "disabled for this tree" and "disabled for all subtrees" be a better naming convention? Subtrees here is referring to any descendant fenced frame trees.

I've also added a note outlining what the states mean.

spec.bs
```diff
@@ -2027,6 +2115,27 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  nonce, if it exists.
+  (<a href="https://github.com/WICG/fenced-frame/issues/191">WICG/fenced-frame#191</a>)
+</div>
+
+The network revocation mechanism requires the following monkeypatches to the [[HTML]] Standard.
+
+<div algorithm=network-status-document-abort>
+  Add a new step to the [=abort a document=] algorithm after step 4 that says:
+
+  5. [=Recalculate the untrusted network status of all fenced frame descendants=] given <var
```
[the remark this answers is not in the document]
blu25: I'm not entirely sure why this algorithm was modified to only be allowed to run in parallel. There's nothing that will block the main thread, so this should be okay to run both in parallel or in the document event loop. Unless you have any objections or there's some piece I'm missing, I'm going to remove that line.

This is needed to account for the following corner case:
1. A parent FF embeds a child FF.
2. The child FF disables its network.
3. The parent FF initiates a navigation of the child to a new config. This will result in a successful navigation and the network status reverting back to full untrusted access.
4. The parent FF immediately after calls `disableUntrustedNetwork()`. The in-progress navigation will be aborted. (Note this behavior is not spec'd yet, but there is [an issue logged for it](https://github.com/WICG/fenced-frame/issues/173))

If we don't recalculate the network statuses after the child FF's navigation is aborted, the promise will get stuck thinking some child frame still has network access.

spec.bs
blu25: Reordering this line to make it clear it's just the last step in the `[=navigate=]` algorithm.

spec.bs
blu25: Right now, this won't do anything. This is really only going to be useful when https://github.com/WICG/fenced-frame/issues/173 is added to the spec. For now, I'm going to take this out, and it can be added back in once the stop in-progress navigation work is added.

blu25:
> One more issue I spotted is that the https://wicg.github.io/fenced-frame/#revoke-network-for-a-partition-nonce algorithm uses "this" which it cannot do since it is called from an in parallel context. I think we should fix that in this PR.

Modified to take in a `relevant settings object` instead of using `this`.

spec.bs
domfarolino: If "subtree" is really just a synonym for "unfenced", then I think the following names would make more sense "disable for this tree" and "disabled for this tree and fenced subtrees"

spec.bs
```diff
@@ -849,6 +857,17 @@ a reporting destination=] through the reference it kept, it will handle all of t
 stored in the [=list=]. If the destination is never finalized, then the pending events will never
 be sent.
 
+An <dfn for=fencedframetype>untrusted network status</dfn> is either "<dfn export for='untrusted
+network status'>`enabled`</dfn>", "<dfn export for='untrusted network status'>`disabled for this
+tree`</dfn>", or "<dfn export for='untrusted network status'>`disabled for all subtrees`</dfn>".
+
+Note: [=untrusted network status/Disabled for this tree=] is not the final network cutoff state. It
+is an intermediate state where the frame has had its network access revoked, but at least one
+subframe still has network access. It does not get special API access at this stage as any
```
domfarolino, line 866: When you say "but at least one subframe still has network access" do you mean normal child subframe, or is this only limited to sub-fenced-frame trees? I think you mean the latter, so I'd clarify that everything inside of the current tree (restricted to fences) suddenly gets network cut-off immediately.

spec.bs
blu25: Replaced the names with "disabled for this tree" and "disabled for this tree and fenced subtrees".

spec.bs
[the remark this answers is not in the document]
domfarolino: Wait how will it get caught by this? Check out the iframe destruction/removal path — I don't think it results in fenced frames being "removed" from the DOM, so I don't think we'd go through this path for fenced frames if their iframe parent is removed.

**Also**, note that the recalculate algorithm requires being in parallel, but the removal steps are not called from in parallel. That's a bug.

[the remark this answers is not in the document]
blu25: I guess I had assumed it was a cascading operation, but it looks like it's not. I'll add a separate `removed from a document` algorithm for iframes. I didn't see any place where this was being called for iframes to begin with, so it should be okay to add a new algorithm.

I'm not sure if there's some common path that would be better to put this in. I considered [the node removal algorithm](https://dom.spec.whatwg.org/#concept-node-remove), but that lives in the DOM spec, and I don't think that's supposed to get access to the concepts of traversables and node navigables. Let me know if there's a better spot you think this could live in.

I also updated the calls to be run in parallel.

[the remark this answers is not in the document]
domfarolino: It's already being called here: https://html.spec.whatwg.org/C#the-iframe-element:html-element-removing-steps. It's using the more appropriate removing steps hook, instead of the removed from document hook, which I have a goal of removing from the HTML spec anyways, so I don't think we should add more instances of it sadly.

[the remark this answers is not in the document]
blu25: I've updated this to be a monkeypatch of that algorithm.

spec.bs
```diff
@@ -1950,32 +1994,106 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        |currentNavigable| and |config|'s [=fenced frame config instance/untrusted network status=]
+        is [=untrusted network status/disabled for this tree=], false otherwise.
+
+        Note: A [=fenced navigable container/fenced navigable=] is added to
+        |navigablesWithNetworkChildren| when it is the unfenced ancestor of another fenced frame
+        that is determined to not be ready for network cutoff.
+
+     1. If |networkCutoffReady| is true:
+
+        1. Set |config|'s [=fenced frame config instance/untrusted network status=] to [=untrusted
```
domfarolino, line 2062: Let's add a little note below this mentioning that any APIs that are gated on this untrusted network status bit are now IMMEDIATELY usable after this change, even before the promises finish resolving. I find this a little weird, but there is no bug here and it at least matches the implementation, so that's fine. We should just document it for sure.

domfarolino (approved):

Issue #179, opened 2024-08-17
[Spec] Fix the navigation patch numberings + patches


blu25:
The navigation algorithm has changed since we last patched it. Our patches now reference steps that are now in different places, or steps that no longer exist since they were refactored into their own separate algorithm. We need to update our patches to match the current state of the navigation part of the spec.

Issue #178, opened 2024-08-16
Navigating from the default FencedFrameConfig(url) constructor doesn't install a fenced frame config.


blu25:
The default `FencedFrameConfig(url)` constructor does not install a fenced frame config in the spec. `FencedFrameConfig(url)` puts a `url` into the `url` slot, and then we can pass in [either a url or urn to the navigate algorithm](https://wicg.github.io/fenced-frame/#ref-for-fencedframeconfig-url). The navigate algorithm [only creates a fenced frame config struct](https://wicg.github.io/fenced-frame/#navigation-changes:~:text=If%20url%20is%20a%20urn%20uuid%20and%20navigable%20is%20a%20fenced%20navigable%3A) and sets the [target fenced frame config](https://wicg.github.io/fenced-frame/#ref-for-source-snapshot-params-target-fenced-frame-config%E2%91%A0) if the value passed in is a urn. So, if we pass in a url to that algorithm instead, no fenced frame config will be created, and no fenced frame config instance will be created either, so the fenced frame won't have access to window.fence.

This should create a fenced frame config + install a fenced frame config instance unconditionally, but that is not what is happening currently.

blu25:
Note that when this is implemented, the [effective sandbox flags](https://wicg.github.io/fenced-frame/#fenced-frame-config-effective-sandbox-flags) will need to be set to the default constructor value (see [kFencedFrameForcedSandboxFlags](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/common/frame/fenced_frame_sandbox_flags.h;l=14-24;drc=047c7dc4ee1ce908d7fea38ca063fa2f80f92c77) for the list of flags it will need to be set to).

PR #180, opened 2024-08-17
[Spec] Add sandbox attribute + handling

blu25:
This CL does the following:

- Specs out the "sandbox" attribute in the `fencedframe` element.
- Specs out the `fencedframe sandboxing flag set` to match the [iframe equivalent](https://html.spec.whatwg.org/multipage/browsers.html#iframe-sandboxing-flag-set).
- Specs out the behavior that ensures that a fenced frame's active sandboxing flag set is at least as restrictive as the fenced frame config's [effective sandboxing flags](https://wicg.github.io/fenced-frame/#fenced-frame-config-effective-sandbox-flags).
- Specs out the behavior that a fenced frame cannot load if the final sandboxing flag set is more restrictive than the [effective sandboxing flags](https://wicg.github.io/fenced-frame/#fenced-frame-config-effective-sandbox-flags).

Follow-up PRs will need to be done for the following:
- Setting the effective sandboxing flags for config-generating APIs in their respective repos.
- Setting the effective sandboxing flags for the default constructor case. See: https://github.com/WICG/fenced-frame/issues/178
- Fixing the numberings for the navigate patches, as the algorithm has changed significantly since we last monkey patched it. See: https://github.com/WICG/fenced-frame/issues/179

See issue: https://github.com/WICG/fenced-frame/issues/155

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/180.html" title="Last updated on Aug 26, 2024, 3:07 PM UTC (663dfc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/180/c799ffe...663dfc2.html" title="Last updated on Aug 26, 2024, 3:07 PM UTC (663dfc2)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -93,6 +93,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: queue a cross-origin embedder policy inheritance violation; url: queue-a-cross-origin-embedder-policy-inheritance-violation
       text: determine navigation params policy container; url: determining-navigation-params-policy-container
       text: cross-origin opener policy enforcement result; url: coop-enforcement-result
+      text: determine the creation sandboxing flags; url: determining-the-creation-sandboxing-flags
+      text: iframe sandboxing flag set; url: iframe-sandboxing-flag-set
       for: cross-origin opener policy enforcement result
         text: needs a browsing context group switch; url: coop-enforcement-bcg-switch
     urlPrefix: dom.html
@@ -106,6 +108,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: dom interface; url: concept-element-dom
       text: accessibility considerations; url: concept-element-accessibility-considerations
       text: represents; url: represents
+      text: is initial about:blank; url: is-initial-about-blank
     urlPrefix: common-dom-interfaces.html
       text: reflect; url: reflect
     urlPrefix: embedder-content-other.html
@@ -145,6 +148,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: navigable; url: navigation-params-navigable
         text: origin; url: navigation-params-origin
         text: COOP enforcement result; url: navigation-params-coop-enforcement-result
+        text: final sandboxing flag set; url: navigation-params-sandboxing
       for: history handling behavior
         text: replace; url: hh-replace
       for: document state
@@ -161,6 +165,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: step; url: she-step
       for: source snapshot params
         text: has transient activation; url: source-snapshot-params-activation
+      text: the navigation must be a replace; url: the-navigation-must-be-a-replace
     urlPrefix: interaction.html
       text: activation notification; url: activation-notification
       text: consume user activation; url: consume-user-activation
@@ -411,6 +416,7 @@ interface HTMLFencedFrameElement : HTMLElement {
   [CEReactions] attribute FencedFrameConfig? config;
   [CEReactions] attribute DOMString width;
   [CEReactions] attribute DOMString height;
+  [SameObject, PutForwards=value] readonly attribute DOMTokenList sandbox;
   [CEReactions] attribute DOMString allow;
 };
 </xmp>
@@ -424,6 +430,22 @@ Descendants of <{fencedframe}> elements represent nothing.
 Each <{fencedframe}> has a <dfn for=fencedframe>config</dfn>, which is either a
 {{FencedFrameConfig}} or null. It is initially null.
 
+Each <{fencedframe}> has a <dfn for=fencedframe>fencedframe sandboxing flag set</dfn>, which is a
+[=sandboxing flag set=]. Which flags in a [=fencedframe/fencedframe sandboxing flag set=] are set
+at any particular time is determined by the <{fencedframe}> element's <{fencedframe/sandbox}>
+attribute.
+
+<div algorithm=determine-sandbox-flags-patch>
+  Modify the [=determine the creation sandboxing flags=] algorithm. Rewrite the second step in the
+  union to be the following 2 steps:
+
+  * If |embedder| is an <{iframe}> element, then: the flags set on |embedder|'s [=iframe sandboxing
+    flag set=].
+
+  * If |embedder| is an <{fencedframe}> element, then: the flags set on |embedder|'s [=fencedframe/
+    fencedframe sandboxing flag set=].
+</div>
+
 <div algorithm=insertion>
   When a <{fencedframe}> element |element| is [=node is inserted into a document|inserted into a
   document=] whose [=Document/browsing context=] is non-null, run these steps:
@@ -433,7 +455,8 @@ Each <{fencedframe}> has a <dfn for=fencedframe>config</dfn>, which is either a
 
   1. Set |nested traversable|'s [=navigable/loading mode=] to "`fencedframe`".
 
-  1. <span class=XXX>Parse the sandbox attributes, once it exists</span>
+  1. If |element| has a <{fencedframe/sandbox}> attribute, then [=parse a sandboxing directive=]
+     given the attribute's value and |element|'s [=fencedframe/fencedframe sandboxing flag set=].
 
   Issue: It's not necessary to call the <a
   href=https://html.spec.whatwg.org/multipage/browsing-the-web.html#url-and-history-update-steps>URL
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -424,6 +428,22 @@ Descendants of <{fencedframe}> elements represent nothing.
...
+attribute.
+
+<div algorithm=determine-sandbox-flags-patch>
+  Modify the [=determine the creation sandboxing flags=] algorithm. Rewrite the second step in the
+  union to be the following 2 steps:
+
+  * If |embedder| is an <{iframe}> element, then: the flags set on |embedder|'s [=iframe sandboxing
+    flag set=].
+
+  * If |embedder| is an <{fencedframe}> element, then: the flags set on |embedder|'s [=fencedframe/
```
domfarolino, line 443:
```suggestion
  * If |embedder| is a <{fencedframe}> element, then: the flags set on |embedder|'s [=fencedframe/
```

spec.bs
```diff
@@ -492,8 +513,26 @@ The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, det
 in the <{fencedframe}>'s [=fenced navigable container/fenced navigable=] is initialized. Its value
 must be a [=serialized permissions policy=]. [[!PERMISSIONS-POLICY]]
 
-The IDL attribute <dfn attribute for=HTMLFencedFrameElement>allow</dfn> must [=reflect=] the
-respective content attribute of the same name.
+The <dfn element-attr for=fencedframe>sandbox</dfn> attribute, when specified, enables a set of
+extra restrictions on any content hosted by the <{fencedframe}>. Its value must be an [=unordered
+set of unique space-separated tokens=] that are [=ASCII case-insensitive=]. The allowed values are
+the same as the allowed values for the <{iframe}>'s <{iframe/sandbox}> attribute.
```
domfarolino, line 519: Please enumerate them here, linking to each dfn, like the iframe section does, instead of just delegating to it.

spec.bs
```diff
@@ -492,8 +513,26 @@ The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, det
...
+
+When an <{fencedframe}> element's <{fencedframe/sandbox}> attribute is set or changed while it has a
```
domfarolino, line 521: I think we should use the [attribute changed steps](https://dom.spec.whatwg.org/#concept-element-attributes-change-ext) extension to be more algorithmic. In fact, https://github.com/whatwg/html/pull/10188#discussion_r1719334729 has some good relevant advice on this.

domfarolino, line 521:
```suggestion
When a <{fencedframe}> element's <{fencedframe/sandbox}> attribute is set or changed while it has a
```

(Likely obsoleted by the above comment)

spec.bs
```diff
@@ -492,8 +513,26 @@ The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, det
...
+non-null [=fenced navigable container/fenced navigable=], the user agent must [=parse a sandboxing
+directive=] given the attribute's value and the <{fencedframe}> element's [=fencedframe/fencedframe
+sandboxing flag set=].
+
+When an <{fencedframe}> element's <{fencedframe/sandbox}> attribute is removed while it has a
```
domfarolino, line 526: Same as above, I think we can cover the removal case with the attribute changed steps too, right?

domfarolino, line 526:
```suggestion
When a <{fencedframe}> element's <{fencedframe/sandbox}> attribute is removed while it has a
```

Like the above, I think this comment is _probably_ obsoleted by the one immediately above.

spec.bs
```diff
@@ -492,8 +513,26 @@ The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, det
...
+non-null [=fenced navigable container/fenced navigable=], the user agent must empty the
+<{fencedframe}> element's [=fencedframe/fencedframe sandboxing flag set=].
```
domfarolino, line 528: Make this more algorithmic by delegating to infra algorithms to clear a list, assuming it's all a list under the hood.

spec.bs
```diff
@@ -2832,6 +2868,20 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
        navigable=], or the result of [=checking if unloading is user-canceled=] for |navigable|'s
        [=navigable/active document=]'s [=Document/inclusive descendant navigables=] otherwise.
 
+  Add a step after step 22.8.2 (Let |finalSandboxFlags| be the [=set/union=]...) that reads:
+
+  3. If |sourceSnapshotParams|' [=source snapshot params/target fenced frame config=] is not null,
```
domfarolino, line 2873:
```suggestion
  3. If |sourceSnapshotParams|'s [=source snapshot params/target fenced frame config=] is not null,
```

I believe in standards we generally prefer the trailing `s` in these cases, even though both are grammatically correct. Can you fix this and the many instances of this below / elsewhere in this PR?

spec.bs
```diff
@@ -2832,6 +2868,20 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+     and if |sourceSnapshotParams|' [=source snapshot params/target fenced frame config=]'s [=fenced
+     frame config/effective sandboxing flags=] is not null, then set |finalSandboxFlags| to the
+     [=set/union=] of |finalSandboxFlags| and |sourceSnapshotParams|' [=source snapshot
+     params/target fenced frame config=]'s [=fenced frame config/effective sandboxing flags=]'
+     [=effective sandboxing flags/value=].
+
+     Note: This ensures that the |finalSandboxFlags| are *at least* as restrictive as the [=fenced
+     frame config/effective sandboxing flags=] defined in the [=source snapshot params/target fenced
+     frame config=]. A separate check will make sure that |finalSandboxFlags| is not more
```
domfarolino, line 2882: Please describe exactly where (and ideally link to, if you can, just for readability and navigability) exactly the separate check is.

spec.bs
```diff
@@ -3359,6 +3410,34 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
+  "`Allowed`":
+
+  1. Let |navigable| be |navigationParams|'s [=navigation params/navigable=].
+
+  1. If |navigable| is not a [=fenced navigable container/fenced navigable=], then return
+     "`Allowed`".
+
+  1. Let |effectiveSandboxingFlags| be the |navigable|'s [=navigable/active browsing context=]'s
+     [=browsing context/fenced frame config instance=]'s [=fenced frame config instance/effective
+     sandboxing flags=].
```
domfarolino, line 3426: This concerns me a little bit, since I think we're pulling the _live_ list of flags here, right?

Could this set of sandboxing flags have changed from what it was observed while loading, to now? Should we be reaching for the snapshot of them at all?

domfarolino (commented):
Looking pretty good.

spec.bs
blu25: Added the full list with links.

spec.bs
blu25: Added an "attribute change steps" section for the fencedframe element with handlers for changing/removing the "sandbox" attribute.

blu25: Obsolete, resolving.

spec.bs
blu25: Obsolete, resolving.

blu25: Added a section to handle removal. I'm assuming that attribute removal will invoke `[=attribute change steps=]` with `|value|` = null, but correct me if that assumption is incorrect.

spec.bs
blu25: It's not as clear as I would like it to be, but the [sandboxing flags set](https://html.spec.whatwg.org/multipage/browsers.html#sandboxing-flag-set) definition says it's a set, but "set" is unlinked. I'm going to assume it's an [ordered set](https://infra.spec.whatwg.org/#ordered-set), which is just a [list](https://infra.spec.whatwg.org/#list) under the hood, which means I can reference the [empty](https://infra.spec.whatwg.org/#list-empty) algorithm for it.

spec.bs
blu25: Fixed 4 instances of `.*s|'` instead of `.*s|'s` in the spec file.

spec.bs
blu25: Added a link to the `should navigation response to navigation request be blocked by sandboxing flags?` algorithm.

spec.bs
blu25: The fenced frame config effective sandboxing flags should stay the same. The only place it will be set is when a config-generating API sets it when first creating the config. It's probably better to not be touching the live navigable for this though, so I'll change it to the `sourceSnapshotParams`'s `target fenced frame config`'s `effective sandboxing flags`.

domfarolino: Of if the `config` attribute was assigned mid-way through navigation, and we end up referencing the "new" config that's not associated with _this_ navigation anymore (although "this" navigation should be cancelled by some of the bail out points anyways, but there are not many of them). Anyways, I think the change you made is good.

spec.bs
domfarolino: That works, thanks!

spec.bs
```diff
@@ -492,8 +513,47 @@ The <dfn element-attr for=fencedframe>allow</dfn> attribute, when specified, det
...
+
+1. If |localName| is <{fencedframe/sandbox}>, then:
+
+   1. If |value| is null, then [=set/empty=] |element|'s [=fencedframe/fencedframe sandboxing flag
+      set=].
+
+   1. Otherwise, run [=parse a sandboxing directive=] given the |value| and |element|'s
+      [=fencedframe/fencedframe sandboxing flag set=].
+
+|element|, |localName|, |oldValue|, |value|, and |namespace|
```
domfarolino, line 555: What's up with this line haha?

domfarolino (commented):
Pretty much lgtm % one comment about what looks like a stray random line of text

spec.bs
blu25: That was very much a stray random line of text so that I could continuously build the spec while writing the algorithm without it complaining about variables only being used once.

domfarolino (approved):

spec.bs
```diff
@@ -3298,16 +3370,18 @@ algorithms to achieve the outcomes described in the above explanatory content.
...
-  before the step inside the [=queue a task|queued task=] starting with "If
-  |failure| is true, then:" that reads:
+  Modify [[HTML]]'s [=attempt to populate the history entry's document=] algorithm. Add the
+  following conditions to step 6.4 (Otherwise, if any of the following are true:) that say:
+  
+  * The result of [=should navigation response to navigation request be blocked by Permissions
+    Policy?=] given <var ignore>navigationParams</var> is "`Blocked`";
 
-  8. Otherwise, if the result of [=should navigation response to navigation request be blocked by
-    Permissions Policy?=] given <var ignore>navigationParams</var> is "`Blocked`", then set
```
yaramh24-alt: Not plocked

PR #181, opened 2024-08-19
[Spec] Clean up + fix outdated navigation references

blu25:
The [navigate](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate) algorithm has changed since we originally monkey patched it. This PR updates our monkey patch to reference the steps as they currently exist. This also cleans up the `Should navigation response to navigation request be blocked by Permissions Policy?` definition to match the formatting of other algorithms.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -3330,7 +3330,7 @@ algorithms to achieve the outcomes described in the above explanatory content.
 </div>
 
 <div algorithm>
-  <dfn>should navigation response to navigation request be blocked by Permissions Policy?</dfn>
+  <dfn>Should navigation response to navigation request be blocked by Permissions Policy?</dfn>
 
   Given a [=navigation params=] (|navigationParams|), this algorithm returns "`Blocked`" or
   "`Allowed`":
--- a/spec.bs
+++ b/spec.bs
@@ -106,6 +106,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: dom interface; url: concept-element-dom
       text: accessibility considerations; url: concept-element-accessibility-considerations
       text: represents; url: represents
+      text: is initial about:blank; url: is-initial-about-blank
     urlPrefix: common-dom-interfaces.html
       text: reflect; url: reflect
     urlPrefix: embedder-content-other.html
@@ -140,6 +141,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: attempt to populate the history entry's document; url: attempt-to-populate-the-history-entry's-document
       text: navigation params; url: navigation-params
       text: snapshot source snapshot params; url: snapshotting-source-snapshot-params
+      text: the navigation must be a replace; url: the-navigation-must-be-a-replace
       for: navigation params
         text: response; url: navigation-params-response
         text: navigable; url: navigation-params-navigable
@@ -2764,11 +2766,17 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
 </div>
 
-<div algorithm=navigate>
-  Modify the definition of [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an
-  optional [=string=] |sharedStorageContext| (default null).
+<div algorithm=navigation-must-replace-patch>
+  Modify [=the navigation must be a replace=] algorithm. Rewrite the definition to read:
 
-  Modify step 7 of [[HTML]]'s [=navigate=] algorithm to include the following condition:
+  [=The navigation must be a replace=], given a [=URL=] <var ignore>url</var> and a [=navigable=]
+  |navigable|, if any of the following are true:
+
+  Further rewrite the second condition to read:
+
+    * |navigable|'s [=navigable/active document=]'s [=is initial about:blank=] is true.
+
+  Add the new condition:
 
     * |navigable| is a [=fenced navigable container/fenced navigable=];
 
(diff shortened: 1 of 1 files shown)


domfarolino:
> This also cleans up the Should navigation response to navigation request be blocked by Permissions Policy? definition to match the formatting of other algorithms.

Other algorithms in our spec, or in HTML? The current declaration of that algorithm seems pretty good / clear, what's wrong with it?

spec.bs
```diff
@@ -2764,11 +2766,17 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
 </div>
 
-<div algorithm=navigate>
-  Modify the definition of [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an
-  optional [=string=] |sharedStorageContext| (default null).
+<div algorithm=navigation-must-replace-patch>
```
domfarolino, line 2769: Why rewrite so much of this existing algorithm? We can just modify more surgically to keep it accepting a Document, but modify the single "navigable is a fenced navigable" condition to go from Document -> [relevant global -> navigable](https://html.spec.whatwg.org/C#window-navigable) to get that navigable.

That seems less invasive right?

spec.bs
```diff
@@ -2780,8 +2788,19 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
     /fenced-frame/history-length-fenced-navigations-replace-do-not-contribute-to-joint.https.html
     /fenced-frame/history-length-outer-page-navigation-not-reflected-in-fenced.https.html
   </wpt>
+</div>
 
-  Modify step 8 of the same algorithm to include the following condition:
+<div algorithm=navigate>
+  Modify the definition of [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an
+  optional [=string=] |sharedStorageContext| (default null).
```
domfarolino, line 2795: While we're here, notice that `#navigate` has all of its optional non-positional params as _definitions_, so they are linkable. Let's do the same here. Let's make `|sharedStorageContext|` not just a variable but a definition, and then link to it in the one spot we reference it below, later in this algorithm.

spec.bs
```diff
@@ -2780,8 +2788,19 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+
+  Modify step 12 of [[HTML]]'s [=navigate=] algorithm to read:
+
+  12. If [=the navigation must be a replace=] given |url| and |navigable|, then set
```
domfarolino, line 2799: This modification is obsolete if we leave the "the navigation must be a replace" algorithm _declaration_ alone, right?

spec.bs
blu25: I was debating on whether it was cleaner to keep the algorithm less changed, or have it set up so that it would be slightly less steps to access things. I'll keep the definition as is and do the document -> global object -> navigable traversal for the new condition.

spec.bs
blu25: Turned `|sharedStorageContext|` into a `<dfn>`.

blu25:
> > This also cleans up the Should navigation response to navigation request be blocked by Permissions Policy? definition to match the formatting of other algorithms.
> 
> Other algorithms in our spec, or in HTML? The current declaration of that algorithm seems pretty good / clear, what's wrong with it?

Most algorithms in our spec don't start with "Create a new algorithm called X..." but instead either just define the algorithm's name or state "To do X given Y, run these steps:". I think it's better for this algorithm to try to match the latter for consistency.

spec.bs
```diff
@@ -108,6 +108,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: dom interface; url: concept-element-dom
       text: accessibility considerations; url: concept-element-accessibility-considerations
       text: represents; url: represents
+      text: is initial about:blank; url: is-initial-about-blank
```
domfarolino, line 111: Is this needed? I don't see it referenced in this PR.

spec.bs
```diff
@@ -2836,8 +2836,14 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
     /fenced-frame/history-length-fenced-navigations-replace-do-not-contribute-to-joint.https.html
     /fenced-frame/history-length-outer-page-navigation-not-reflected-in-fenced.https.html
   </wpt>
+</div>
 
-  Modify step 8 of the same algorithm to include the following condition:
+<div algorithm=navigate>
+  Modify the definition of [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an
```
domfarolino, line 2842:
```suggestion
  Modify [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an
```

(might need to rewrap)

spec.bs
```diff
@@ -2836,8 +2836,14 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+  optional [=string=] <dfn for=navigate>sharedStorageContext</dfn> (default null).
+
+  Modify step 13 of [[HTML]]'s [=navigate=] algorithm (If all of the following are true:) to include
```
domfarolino, line 2845:
```suggestion
  Modify step 13 of [[HTML]]'s [=navigate=] algorithm ("If all of the following are true":) to include
```

(might need to rewrap)

domfarolino (commented):
Looks like there are some new rebase conflicts, can you resolve them?

spec.bs
blu25: Not entirely sure why this is there, but this was the source of the merge conflict. I'll remove it

domfarolino (approved):

PR #182, opened 2024-08-19
Remove unused config size attributes

VergeA:
The IDL containerWidth, containerHeight, contentWidth, and contentHeight attributes not actually implemented in Chromuim, so we should remove them.

Additionally, the internal fenced frame config struct does not implement container size, only content size. So we should remove that as well.

***
<a href="https://pr-preview.s3.amazonaws.com/VergeA/fenced-frame/pull/182.html" title="Last updated on Aug 29, 2024, 5:37 PM UTC (7a51c0e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/182/c799ffe...VergeA:7a51c0e.html" title="Last updated on Aug 29, 2024, 5:37 PM UTC (7a51c0e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -666,9 +666,9 @@ boundary, which can allow for colluding parties to join cross-site data and buil
 user. To prevent that, the ad auction API [=construct a pending fenced frame config|constructs=] a
 [=fenced frame config=] whose underlying [=fenced frame config/mapped url|URL=] is opaque to the
 embedding context. The [=fenced frame config=] is also constructed with restrictions on what the
-[=fenced frame config/container size=] and [=fenced frame config/content size=] of the frame must be
-and what the [=fenced frame config/effective enabled permissions|permissions policy=] of the frame
-must be, as those can be used as fingerprinting vectors.
+[=fenced frame config/content size=] of the frame must be and what the [=fenced frame config/
+effective enabled permissions|permissions policy=] of the frame must be, as those can be used as 
+fingerprinting vectors.
 
 Displaying a personalized payment button:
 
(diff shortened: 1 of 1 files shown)


domfarolino:
Is this ready for review?

VergeA:
Yes, sorry! I don't think I have the proper permissions on the repo to add reviewers myself.

domfarolino (approved):
Very nice, thanks for this cleanup!

domfarolino:
Before I merge, can I ask @blu25 to take a quick look too? Just since you all are more familiar with the latest FF implementation than me at the moment.

VergeA:
Thanks for the review Dom! And yep, sounds good.

spec.bs
```diff
@@ -1324,12 +1315,6 @@ maps to an internal [=fenced frame config=] [=struct=].
   [Exposed=Window, Serializable]
   interface FencedFrameConfig {
     constructor(USVString url);
-    
-    readonly attribute FencedFrameConfigSize? containerWidth;
-    readonly attribute FencedFrameConfigSize? containerHeight;
-    readonly attribute FencedFrameConfigSize? contentWidth;
-    readonly attribute FencedFrameConfigSize? contentHeight;
```
blu25, line 1331: Is the `FencedFrameConfigSize` typedef being used anywhere at this point? It looks like no, so we should probably remove that as well. (Should also clean this up in the implementation while we're at it.)

Same with `FencedFrameConfigURL` I think.

VergeA: Done.

blu25 (approved):
Changes LGTM, thanks!

PR #183, opened 2024-08-19
[Spec] Fix navigation from the default constructor not installing a fenced frame config.

blu25:
The existing spec results in a fenced frame navigating directly to the `url` stored in a `FencedFrameConfig` IDL object when navigating to a config built with the `FencedFrameConfig(url)` default constructor. This results in a `fenced frame config` and `fenced frame config instance` not being created for the fenced frame, cutting off the internal document from `window.fence`.

This PR fixes that in the following ways:

- Modifies the `FencedFrameConfig(url)` constructor to also create a `fenced frame config` and add it to the navigable's `fenced frame config mapping`.
- Unconditionally passes in the `FencedFrameConfig`'s `urn` when navigating the internal document, ensuring that the navigation will find the associated `fenced frame config`, and instantiate a `fenced frame config instance` as expected.
- Removes the `url` field from the `FencedFrameConfig` in favor of using the `mapped url` in the `fenced frame config`/`fenced frame config instance`.
- Adds a method to store a new `FencedFrameConfig` directly in the finalized mapping without first putting it in the pending mapping.

This PR also adds a `default fenced frame effective sandboxing flags` constant, which is needed when building the `fenced frame config` in the `FencedFrameConfing(url)` constructor.

See issue: https://github.com/WICG/fenced-frame/issues/178

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/183.html" title="Last updated on Oct 5, 2024, 7:50 PM UTC (3675ec4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/183/00676a5...3675ec4.html" title="Last updated on Oct 5, 2024, 7:50 PM UTC (3675ec4)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1454,9 +1454,9 @@ Note: A config's [=fencedframeconfig/url=] is only null if a [=fencedframeconfig
   |value|, |serialized|, and |forStorage| are:
 
   1. If |forStorage| is true, then throw a {{DataCloneError}} {{DOMException}}.
-
-  1. Set |serialized|.\[[Url]] to |value|'s [=fencedframeconfig/url=].
   
+  1. Set |serialized|.\[[Url]] to |value|'s [=fencedframeconfig/url=].
+
   1. Set |serialized|.\[[Urn]] to |value|'s [=fencedframeconfig/urn=].
 
   1. Set |serialized|.\[[SharedStorageContext]] to |value|'s [=fencedframeconfig/
@@ -1480,7 +1480,7 @@ Note: A config's [=fencedframeconfig/url=] is only null if a [=fencedframeconfig
 <div algorithm="FencedFrameConfig deserializer">
   Their [=deserialization steps=], given |serialized|, |value|, and <var ignore> targetRealm</var>
   are:
-  
+
   1. Initialize |value|'s [=fencedframeconfig/url=] to |serialized|.\[[Url]].
   
   1. Initialize |value|'s [=fencedframeconfig/urn=] to |serialized|.\[[Urn]].
--- a/spec.bs
+++ b/spec.bs
@@ -2892,8 +2892,9 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
       1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
   
-    1. If |navigable| is a [=fenced navigable container/fenced navigable=], and
-       <var ignore>sourceDocument</var>'s [=node navigable=] is not |navigable|:
+    1. If |navigable| is a [=fenced navigable container/fenced navigable=] and
+       <var ignore>sourceDocument</var>'s [=node navigable=] is in |navigable|'s
+       [=navigable/active document=]'s [=Document/ancestor navigables=]:
 
       1. Let |config| be a new [=fenced frame config=] with the following [=struct/items=]:
 
--- a/spec.bs
+++ b/spec.bs
@@ -2436,18 +2436,6 @@ in the [[#nested-traversables-intro]].
   1. Return |navigables|.
 </div>
 
-<div algorithm="inclusive-ancestor-navigables-patch">
-  Modify the [=Document/inclusive ancestor navigables=] algorithm to take a new optional
-  [=boolean=] argument <var><dfn lt="inclusive-an-unfenced">unfenced</dfn></var> that defaults to
-  false.
-
-  Further rewrite step 1 of this algorithm to:
-
-  1. Let <var ignore>navigables</var> be <var ignore>document</var>'s [=Document/ancestor
-     navigables=] with [=an-unfenced|unfenced=] set to
-     <var>[=inclusive-an-unfenced|unfenced=]</var>.
-</div>
-
 <div algorithm="ancestor-navigables-patch">
   Modify the [=Document/ancestor navigables=] algorithm to take a new optional [=boolean=]
   argument <dfn lt="an-unfenced">unfenced</dfn> that defaults to false, and rewrite the algorithm
--- a/spec.bs
+++ b/spec.bs
@@ -2893,12 +2893,14 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
     1. If |url| is a [=urn uuid=] and |navigable| is a [=fenced navigable container/fenced
        navigable=]:
 
-       Issue: If a fenced frame generates a FencedFrameConfig using a config-generating API, and
-       then correctly guesses the urn:uuid of that config, it can currently navigate itself to that
-       config, even though this is meant to only allow embedders to navigate fenced frames to
-       configs. This algorithm should be patched to be able to take in a FencedFrameConfig and use
-       that as the check to determine if this path is followed. See:
-       [issue #194](https://github.com/WICG/fenced-frame/issues/194)
+       Issue: The above condition is not as tight as it needs to be. For example, if a
+       <{fencedframe}> generates a {{FencedFrameConfig}} using a config-generating API, and then
+       correctly guesses the config's [=fencedframeconfig/urn|urn:uuid=], it can theoretically
+       navigate itself to that config by passing the guessed urn into the navigate algorithm as a
+       [=URL=], via something like the {{Window/location}} API. This is bad, because the purpose of
+       a {{FencedFrameConfig}} is to ensure that only an embedder can navigate a <{fencedframe}> to
+       the resource represented by the config, by using the config object directly. See <a
+       href=https://github.com/WICG/fenced-frame/issues/194>#194</a> for thoughts on fixing this.
 
       1. Let |config| be the result of [=fenced frame config mapping/finding a
          config=] in <var ignore>sourceDocument</var>'s [=node navigable=]'s [=navigable/traversable
--- a/spec.bs
+++ b/spec.bs
@@ -2879,7 +2879,10 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
       1. Set |config|'s [=fenced frame config/embedder shared storage context=] to
          |sharedStorageContext|.
 
-      1. Set <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/target fenced frame
+      1. Assert: |sourceSnapshotParams|'s [=source snapshot params/target fenced frame config=] is
+         null.
+
+      1. Set |sourceSnapshotParams|'s [=source snapshot params/target fenced frame
          config=] to |config|.
 
       1. [=Assert=] |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=] is a
@@ -2889,7 +2892,7 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
       1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
   
-    1. If |url| is a [=url=], |navigable| is a [=fenced navigable container/fenced navigable=], and
+    1. If |navigable| is a [=fenced navigable container/fenced navigable=], and
        <var ignore>sourceDocument</var>'s [=node navigable=] is not |navigable|:
 
       1. Let |config| be a new [=fenced frame config=] with the following [=struct/items=]:
@@ -2914,9 +2917,12 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
         : [=fenced frame config/effective enabled permissions=]
         :: null
+      
+      1. Assert: |sourceSnapshotParams|'s [=source snapshot params/target fenced frame config=] is
+         null.
 
-      1. Set <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/target fenced frame
-         config=] to |config|.
+      1. Set |sourceSnapshotParams|'s [=source snapshot params/target fenced frame config=] to
+         |config|.
 
   <wpt>
     /fenced-frame/frame-navigation.https.html
--- a/spec.bs
+++ b/spec.bs
@@ -465,18 +465,19 @@ The <dfn attribute for=HTMLFencedFrameElement>config</dfn> IDL attribute getter 
        Note: This holds because when the element has been removed from the DOM, its removal steps
        immediately destroy the [=fenced navigable container/fenced navigable=].
 
-  1. Let |navigationUrn| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/urn=].
+  1. Let |navigation url or urn| be the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] if
+     the given {{FencedFrameConfig}}'s [=fencedframeconfig/url=] is not null, and the given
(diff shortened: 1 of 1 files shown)


domfarolino:
I assumed the changes described here would match the implementation but that doesn't seem to be the case in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/html_fenced_frame_element.cc;l=518-535;drc=82dff63dbf9db05e9274e11d9128af7b9f51ceaa, or the constructor which doesn't store or create an underlying config in default mode, right?

blu25:
> I assumed the changes described here would match the implementation but that doesn't seem to be the case in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/html_fenced_frame_element.cc;l=518-535;drc=82dff63dbf9db05e9274e11d9128af7b9f51ceaa, or the constructor which doesn't store or create an underlying config in default mode, right?

There's no underlying config being added to the [urn/config map](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_url_mapping.h;l=188;drc=81f4f2c7936cc451e4071b9b7b25d86040d869c7), but the navigation request is creating a [new fenced frame properties](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=1705-1708;drc=78865bbb3da76c31548130e87e20ce1357ac3c8b) object that gets installed in the resulting navigated document. So we do need a `fenced frame config` object to exist and be accessible by the inner document by the time navigation ends. I do agree that this should be done without touching the pending/finalized mapping, since those aren't necessary.

I think the way to have this match the implementation is to have the navigation algorithm directly create a new `fenced frame config` and put it into the `target fenced frame config` in the `source snapshot params` for embedder-initiated navigations. That will ensure that the config object is in place for this navigation type, and we can keep the mapping untouched.

spec.bs
```diff
@@ -2877,6 +2888,35 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
       1. Set |url| to |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
 
       1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
+  
+    1. If |url| is a [=url=], |navigable| is a [=fenced navigable container/fenced navigable=], and
```
domfarolino, line 2892:
```suggestion
    1. If |url| is a [=URL=], |navigable| is a [=fenced navigable container/fenced navigable=], and
```

domfarolino, line 2892: Actually, I think `|url|` is _always_ a URL here, right? URN UUIDs are URLs themselves, so I think this portion of the condition is redundant?

domfarolino:
> I think the way to have this match the implementation is to have the navigation algorithm directly create a new fenced frame config and put it into the target fenced frame config in the source snapshot params for embedder-initiated navigations. That will ensure that the config object is in place for this navigation type, and we can keep the mapping untouched.

Yep this is exactly what I was thinking too.

spec.bs
```diff
@@ -2877,6 +2888,35 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
...
+           : [=effective sandboxing flags/value=]
+           :: The [=fencedframetype/default fenced frame effective sandboxing flags=].
+
+           : [=effective sandboxing flags/visibility=]
+           :: [=visibility/opaque=]
+
+        : [=fenced frame config/effective enabled permissions=]
+        :: null
+
+      1. Set <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/target fenced frame
```
domfarolino, line 2918: At this point, we're guaranteed that the target fenced frame config is null before this step, and that we're not clobbering an existing config right? Maybe add an assert?

domfarolino (commented):
Generally LGTM with small nits

spec.bs
blu25: `|url|` can be a urn:uuid if `|navigable|` is not a fenced navigable, but that will get caught here anyway, so it should be fine to remove.

spec.bs
blu25: Done. Also added an equivalent assert for the urn:uuid case.

spec.bs
```diff
@@ -2877,6 +2891,38 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
       1. Set |url| to |config|'s [=fenced frame config/mapped url=]'s [=mapped url/value=].
 
       1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
+  
+    1. If |navigable| is a [=fenced navigable container/fenced navigable=], and
+       <var ignore>sourceDocument</var>'s [=node navigable=] is not |navigable|:
```
domfarolino, line 2896: This is to limit the condition to embedder-initiated navigations, right? But this wouldn't catch navigations initiated by a deep descendant _inside_ a fenced frame, that navigates `window.top`, right? That would pass this condition, which I think is not what we want, right?

Separately, can you point me to where, in the implementation, this check is taking place? (Just so I know what I'm talking about).

blu25: > That would pass this condition, which I think is not what we want, right?

Yes, that would be allowed under the current algorithm state, which we don't want. Should we also be concerned about the "If url is a [urn uuid](https://www.ietf.org/rfc/rfc4122.txt#urn-uuid) and navigable is a [fenced navigable](https://github.com/WICG/fenced-frame/pull/183#fenced-navigable-container-fenced-navigable):" part as well? If a descendant navigable somehow knew a valid `urn:uuid` in the fenced frame mapping, could it also navigate the fenced frame root using that `urn:uuid`? I'm not sure if there are other safeguards in place to prevent this from happening, or if we should add a "embedder-initiated" check there as well.

> Separately, can you point me to where, in the implementation, this check is taking place? (Just so I know what I'm talking about).

The impl equivalent takes place [in the navigation request file](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=1712-1715;drc=f522344e45882da4c7f7cb1b3a0a7bd747d654bb). It's checking for [is_embedder_initiated_fenced_frame_navigation](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;drc=f522344e45882da4c7f7cb1b3a0a7bd747d654bb;bpv=1;bpt=1;l=1663) which is set by the [`FencedFrame::Navigate()` method](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame.cc;l=182;drc=bd15364a662c095e05f2ded9fb7c578e46281fb0). That can only be invoked from the `HTMLFencedFrameElement`.

If we want to perfectly match that, we could add an optional boolean to the [`Navigate` algorithm](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate) for `is embedder initiated fenced frame navigation`.

Alternatively, we could just check for whether `sourceDocument's node navigable` is in the `|navigable|`'s [`active document`](https://html.spec.whatwg.org/multipage/document-sequences.html#nav-document)'s [`ancestor navigables`](https://html.spec.whatwg.org/multipage/document-sequences.html#ancestor-navigables) list. I'll do this method, but if you feel like we should do the other one let me know.

domfarolino: > I'll do this method, but if you feel like we should do the other one let me know.

I think that less-invasive approach makes sense. Plus I think it solves the earlier question you asked, right?

> Should we also be concerned about the "If url is a [urn uuid](https://www.ietf.org/rfc/rfc4122.txt#urn-uuid) and navigable is a https://github.com/WICG/fenced-frame/pull/183#fenced-navigable-container-fenced-navigable:" part as well? If a descendant navigable somehow knew a valid urn:uuid in the fenced frame mapping, could it also navigate the fenced frame root using that urn:uuid?

That is, your approach stops descendant navigables from doing this, right? I guess the spec could accidentally allow some descendant navigable to still navigate to a URN that it somehow guesses correctly (since it should be buried inside of a `FencedFrameConfig` object), and still cause this outcome which isn't great. Maybe as a follow-up we can make the navigate algorithm take a `FencedFrameConfig` object, to make it clear that the ability to call navigate with a URL is tied to the fenced frame navigation path only. Let me know if you think that sounds like the right approach, and then maybe you could file a follow-up issue?

blu25: > That is, your approach stops descendant navigables from doing this, right?

Correct. I think as the spec currently stands, if a fenced frame can generate a FencedFrameConfig, figure out what the urn:uuid for it is, it can navigate **itself** to that config via `location.href = urn_i_generated`, which I don't think we want.

I'll also update the other branch to do the ancestor check as well.

While I was doing this, I realized that we don't actually have unfenced versions of the ancestor navigables algorithms. We do have them for the descendant navigables algorithms. I've updated this PR to add those patches in and call the unfenced versions of them in these 2 spots.

> Maybe as a follow-up we can make the navigate algorithm take a FencedFrameConfig object

I think that would be good. Then the thing I described above with a fenced frame navigating itself to a urn:uuid wouldn't be possible. I created https://github.com/WICG/fenced-frame/issues/194 to track this.

domfarolino: > Correct. I think as the spec currently stands, if a fenced frame can generate a FencedFrameConfig, figure out what the urn:uuid for it is, it can navigate itself to that config via location.href = urn_i_generated, which I don't think we want.

This is weird. I assumed the given spec's limitation (which you've captured in #194) only made it possible for an embedder to navigate a fenced frame if it "guesses" the `urn uuid` inside the config that it already has, but it seems like the spec might also be vulnerable to a fenced frame generating a config and navigating _itself_ to that config's urn (assuming it can guess the urn of course)? I guess that's because [here in the spec](https://wicg.github.io/fenced-frame/#ref-for-traversable-navigable-fenced-frame-config-mapping%E2%91%A2) we only check the source document's traversable navigable's fenced frame config mapping.

My understanding is that the implementation does not have this limitation though, right? That's for fenced frame navigations, the [implementation just blindly consults its _embedder_'s frame tree's mapping](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=2525-2533;drc=e66b343b5554785a32ad988bfe5c3c524f5e1857). Do we know why the spec blindly looks at _sourceDocument_ — which might not be the embedder, but might actually be the fenced frame itself — when consulting the mapping?

I'm not sure this is true, because when we compare urn UUIDs, we always compare the given one with ones in the map of the embedder, not a frame's own map, right? (At least I think that's how the implementation works... we always look at the embedder's frame tree, and I feel like we do the same in spec?

spec.bs
```diff
@@ -2863,8 +2902,10 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   Insert these steps immediately after step 20, the step that goes [=in parallel=], so that what
   follows are the first steps that run [=in parallel=] in the patched algorithm:
 
-    1. If |url| is a [=urn uuid=] and |navigable| is a [=fenced navigable container/fenced
-       navigable=]:
+    1. If |url| is a [=urn uuid=], |navigable| is a [=fenced navigable container/fenced navigable=],
+       and <var ignore>sourceDocument</var>'s [=node navigable=] is in |navigable|'s
+       [=navigable/active document=]'s [=Document/ancestor navigables=] with [=an-unfenced|
+       unfenced=] set to true:
```
domfarolino, line 2908: How about this: instead of adding this condition to patch up the concern found in #194, let's leave it open to the bugs described there, and briefly describe why this step is not sufficient and that #194 tracks this, in an `Issue:` indented below this?

spec.bs
```diff
@@ -2892,9 +2933,9 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 
       1. Run steps in |config|'s [=fenced frame config/on navigate callback=].
   
-    1. If |navigable| is a [=fenced navigable container/fenced navigable=] and
-       <var ignore>sourceDocument</var>'s [=node navigable=] is in |navigable|'s
-       [=navigable/active document=]'s [=Document/ancestor navigables=]:
+    1. If |navigable| is a [=fenced navigable container/fenced navigable=] and <var
+       ignore>sourceDocument</var>'s [=node navigable=] is in |navigable|'s [=navigable/active
+       document=]'s [=Document/ancestor navigables=] with [=an-unfenced|unfenced=] set to true:
```
domfarolino, line 2938: I think this part (i.e., this usage of "ancestor navigables") is fine though, since it doesn't have to do with gating the access of the fenced frame URN config mapping. Do you agree?

spec.bs
```diff
@@ -2436,6 +2436,45 @@ in the [[#nested-traversables-intro]].
   1. Return |navigables|.
 </div>
 
+<div algorithm="inclusive-ancestor-navigables-patch">
+  Modify the [=Document/inclusive ancestor navigables=] algorithm to take a new optional
```
domfarolino, line 2440: We don't call this anywhere, right?

blu25: At the moment, no. I think it's worth keeping this in place so it will always correctly invoke "ancestor navigables", and we won't need to remember to patch it later if we need to use it.

spec.bs
blu25: I think this is fine. This is also needed because we only want to create a fenced frame config if the navigation is initiated by the fenced frame's embedder. If a fenced frame's document navigates itself, that should just re-use the existing config.

spec.bs
blu25: > Do we know why the spec blindly looks at sourceDocument — which might not be the embedder, but might actually be the fenced frame itself — when consulting the mapping?

If I were to take a guess, maybe it's because we needed access to a document at that point, and `sourceDocument` at the time felt like it was enough of an analog. At this point we can just as easily get navigable's unfenced parent's active document, right?

> I'm not sure this is true, because when we compare urn UUIDs, we always compare the given one with ones in the map of the embedder, not a frame's own map, right? (At least I think that's how the implementation works... we always look at the embedder's frame tree, and I feel like we do the same in spec?

I believe that's what's happening in the implementation. The urn lives in a mapping that lives in the embedder (at this point, the fenced frame might not even have been created yet). The spec should do that as well.

spec.bs
domfarolino: Eh sorry, personally I don't think we should clutter the blame history with this if we're not using it. Let's introduce it when we need it, alright?

spec.bs
domfarolino: > At this point we can just as easily get navigable's unfenced parent's active document, right?

I think, so? Assuming that doesn't break URN iframes, which I think it should be fine since the URN appears in the parent's (fenced, though)'s config mapping. Anyways, we don't have to take care of that in this PR. It can be done separately.

blu25: Sounds good. I've linked this comment thread in https://github.com/WICG/fenced-frame/issues/194 as a reference, and that can be taken care of as part of that effort.

spec.bs
blu25: Ok. I'll remove it.

domfarolino (approved):

PR #203, opened 2025-01-08
[Spec] Allow setting automatic beacon data from cross-origin subframes.

blu25:
This PR modifies the automatic beacon algorithms to support data being set from frames that are cross-origin to the fenced frame config's mapped URL. More specifically, this makes the following modifications:

- `setReportEventDataForAutomaticBeacons()` is modified to allow data to be set from cross-origin subframes, but only if the data's `crossOriginExposed` parameter is set to `true`.
- A new `snapshot automatic beacon mapping` type is introduced that contains 2 mappings. The first mapping is the same as the existing `automatic beacon data map`. The second mapping only contains automatic beacon data whose `crossOriginExposed` parameter is set to true. This is needed since, as part of this change, cross-origin subframes that trigger automatic beacons will now use the first cross-origin exposed data it finds up the frame tree, versus the first data it finds regardless of whether it is cross-origin exposed or not (which it wouldn't be able to use).
- The `attempt to send an automatic beacon` algorithm now checks either the "all data mapping" or the "cross-origin only mapping" based on whether the automatic beacon initiator is cross-origin to the fenced frame config's mapped URL. This will ensure that if an ancestor has usable data, it won't accidentally grab data from a closer ancestor that can't be used. This also has a side effect of letting us remove the `|should send beacon with data|` variable, as the same behavior is now obtained by simply checking the cross-origin only data mapping.
- The `get the automatic beacon data mapping to use` is renamed and modified to read data into both the regular mapping and the cross-origin only mapping.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/203.html" title="Last updated on Jan 22, 2025, 3:27 PM UTC (0c72724)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/203/ffd3d63...0c72724.html" title="Last updated on Jan 22, 2025, 3:27 PM UTC (0c72724)">Diff</a>


spec.bs
```diff
@@ -2173,6 +2171,11 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
     /fenced-frame/automatic-beacon-cross-origin-no-opt-in.https.html
     /fenced-frame/automatic-beacon-use-ancestor-data.https.html
   </wpt>
+
+  TODO: Add the following tests:
```
domfarolino, line 2175: This is blocking merge right?

spec.bs
```diff
@@ -2371,9 +2372,23 @@ An <dfn for=fencedframetype>automatic beacon data</dfn> is a [=struct=] with the
...
 </dl>
 
-Each {{Document}} object has an associated <dfn for=Document>automatic beacon data map</dfn>, which
-is a [=map=] whose [=map/keys=] are [=fencedframetype/automatic beacon event type=]s and whose
-[=map/values=] are null or an [=fencedframetype/automatic beacon data=]
+An <dfn for=fencedframetype>automatic beacon data mapping</dfn> is a [=map=] whose [=map/keys=] are
+[=fencedframetype/automatic beacon event type=]s and whose [=map/values=] are null or an
+[=fencedframetype/automatic beacon data=].
+
+An <dfn for=fencedframetype>snapshot automatic beacon mapping</dfn> is a [=struct=] with the
```
VergeA, line 2379: I think this dfn and the one above it are named too similarly. Can we call this top-level structure something else? Maybe "automatic beacon data store" or something?

spec.bs
```diff
@@ -2239,22 +2239,22 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
...
+     beacon mapping=]'s [=snapshot automatic beacon mapping/data mapping=] if |isCrossOrigin| is
+     false, |sourceSnapshotParams|'s [=source snapshot params/snapshot automatic beacon mapping=]'s
+     [=snapshot automatic beacon mapping/cross-origin exposed data mapping=] otherwise.
 
-  1. If |beacon data| is null and |has header opt in| is false, abort these steps.
+  1. Let |beaconData| be |beaconMapping|[|eventType|].
 
-  1. Let |is cross origin| be true if |sourceOrigin| is not [=same origin=] with |config|'s [=fenced
-     frame config instance/mapped url=]'s [=url/origin=], false otherwise.
+  1. Let |has header opt in| be |sourceSnapshotParams|'s [=source snapshot params/automatic beacons
```
VergeA, line 2252: Can we just call this |automaticBeaconsAllowed|? I don't know if the opt-in header is relevant here, because we're a few levels of indirection away from where it's defined, and we're really just interested in the "allowed" boolean anyway.

Maybe just add a note that this comes from an opt-in header?

spec.bs
blu25: Renamed to "snapshot automatic beacon store".

spec.bs
blu25: Renamed the variable and added a note explaining that beacons can still be sent even if set to false.

VergeA (approved):
Looks good now, thanks!

spec.bs
```diff
@@ -2239,22 +2239,26 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
...
-     allowed=].
+  1. Let |beaconData| be |beaconMapping|[|eventType|].
 
-  1. If |beacon data| is null and |has header opt in| is false, abort these steps.
+  1. Let |automaticBeaconsAllowed| be |sourceSnapshotParams|'s [=source snapshot params/automatic
+     beacons allowed=].
 
-  1. Let |is cross origin| be true if |sourceOrigin| is not [=same origin=] with |config|'s [=fenced
-     frame config instance/mapped url=]'s [=url/origin=], false otherwise.
+     Note: This tracks whether there was an opt in via the [:Allow-Fenced-Frame-Automatic-Beacons:]
```
domfarolino, line 2255:
```suggestion
     Note: This tracks whether there was an opt-in via the [:Allow-Fenced-Frame-Automatic-Beacons:]
```

spec.bs
```diff
@@ -2239,22 +2239,26 @@ event|event-level beacon=] when a fenced frame initiates a successful [=navigate
...
+     header. Automatic beacons can still be sent out while this value is false for either
+     same-origin documents or documents that opt in via setting data.
```
domfarolino, line 2257: same

domfarolino (approved):
I'm happy to defer to @VergeA's review here since the annals of the cross-origin automatic beacon work is slightly beyond me at this point, but from looking at the PR nothing major sticks out.

PR #204, opened 2025-01-10
Add spec for APIs gated on network revocation.

VergeA:
After untrusted network access has been disabled in a fenced frame via disableUntrustedNetwork(), script running in the frame will gain access to new powerful APIs. This patch provides an algorithm that other specs can use when implementing methods that should only succeed in a fenced frame with network access revoked.

Currently, the only method that will rely on this algorithm is SharedStorage.get(), which is implemented in the Shared Storage draft spec. I'm planning to update that spec separately after this PR merges.

***
<a href="https://pr-preview.s3.amazonaws.com/VergeA/fenced-frame/pull/204.html" title="Last updated on Jan 14, 2025, 7:16 PM UTC (9fa9fbc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/204/ffd3d63...VergeA:9fa9fbc.html" title="Last updated on Jan 14, 2025, 7:16 PM UTC (9fa9fbc)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2196,7 +2196,7 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
  * CSP directive of <code>fenced-frame-src</code>
  * Features as HTML's <code>&lt;fencedframe&gt;</code>
 
-<h3 id=methods-gated-on-network-revocation>Methods gated on network revocation</h3>
+<h3 id=gating-methods-on-network-revocation>Gating methods on network revocation</h3>
 
 *This first introductory paragraph is non-normative.*
 
--- a/spec.bs
+++ b/spec.bs
@@ -762,11 +762,11 @@ can freely flow in and out without risk of the credit card information being joi
 data. Because of that, the fenced frame can be constructed directly from the web platform using the
 {{FencedFrameConfig}} constructor without compromising privacy. The button at this point has no
 personalized data in it since it can't access the credit card data yet. The {{Document}} can only
-read that credit card data once it turns off all network access, preventing the data from flowing
-out of the fenced frame and preventing it from being joined with cross-site data to build a user
-profile. Once it does that, the button will then display the last 4 digits of the user's credit card
-number, as it is saved in the browser, inside the first-party storage partition for the ecommerce
-platform's origin.
+read that credit card data once it turns off all network access via
+{{Fence/disableUntrustedNetwork()}}, preventing the data from flowing out of the fenced frame and
+preventing it from being joined with cross-site data to build a user profile. Once it does that,
+the button will then display the last 4 digits of the user's credit card number, as it is saved in
+the browser, inside the first-party storage partition for the ecommerce platform's origin.
 
 <h4 id=fenced-frame-config-struct>The [=fenced frame config=] [=struct=]</h4>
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2196,6 +2196,34 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
...
 
+<h3 id=methods-gated-on-network-revocation>Methods gated on network revocation</h3>
+
+*This first introductory paragraph is non-normative.*
+
+After a fenced frame has fully disabled untrusted network access, meaning the {{Promise}} returned
+by {{Fence/disableUntrustedNetwork()}} has [=resolved=], certain powerful interface methods will
+become available to script which executes inside of the fenced frame. These methods are defined in
+other specifications, which will use the below algorithm to determine if invocation can occur
+successfully. Currently, the only method which is gated behind revocation of untrusted network
```
blu25, line 2207: Should we rephrase this to say "An example of a method which is gated behind revocation..."? That way, we won't need to go back and reword this if another use case is added.

blu25 (commented):
Algorithm looks good. Just one question on the introductory paragraph.

spec.bs
VergeA: In the case of new methods being added, I was imagining this becoming a non-normative list, with links to the relevant specs containing all the methods the fenced frame script can now access. But for now, I didn't think it made sense to make a list with only one element.

If that list would be cumbersome to maintain, then I'm fine with your suggestion, but I'm also fine with just listing all methods we know about explicitly.

blu25: I feel like it would be annoying to maintain. It would also be easy to forget to update the list. So I'd prefer adding the one example and then letting the [existing spec linking infrastructure](https://dontcallmedom.github.io/webdex/f.html#fetch%40%40fetch%25%25dfn) take care of showing people where it's being used.

VergeA: SGTM, updated this sentence with your suggestion and re-wrapped.

blu25 (approved):

spec.bs
```diff
@@ -2196,6 +2196,34 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
...
+After a fenced frame has fully disabled untrusted network access, meaning the {{Promise}} returned
+by {{Fence/disableUntrustedNetwork()}} has [=resolved=], certain powerful interface methods will
+become available to script which executes inside of the fenced frame. These methods are defined in
+other specifications, which will use the below algorithm to determine if invocation can occur
+successfully. One example of a method which is gated behind revocation of untrusted network access
+is {{SharedStorage/get()}} when invoked outside of a {{SharedStorageWorklet}}. This method is
+defined in the [[Shared-Storage]] draft specification.
+
+<div algorithm>
+  To <dfn export>determine if a navigable has fully revoked network</dfn> given a [=navigable=]
```
domfarolino, line 2212: Out of curiosity, since we're exporting this, is there a corresponding PR made in the Shared Storage spec yet?

spec.bs
```diff
@@ -2196,6 +2196,34 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
  * CSP directive of <code>fenced-frame-src</code>
  * Features as HTML's <code>&lt;fencedframe&gt;</code>
 
+<h3 id=methods-gated-on-network-revocation>Methods gated on network revocation</h3>
```
domfarolino, line 2199: I think this is misleading since we're not introducing the methods here. Let's rename this (and its `id`) to something like "Gating methods on network revocation"

spec.bs
```diff
@@ -2196,6 +2196,34 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
...
+other specifications, which will use the below algorithm to determine if invocation can occur
+successfully. One example of a method which is gated behind revocation of untrusted network access
+is {{SharedStorage/get()}} when invoked outside of a {{SharedStorageWorklet}}. This method is
+defined in the [[Shared-Storage]] draft specification.
+
+<div algorithm>
+  To <dfn export>determine if a navigable has fully revoked network</dfn> given a [=navigable=] 
+  |navigable|:
+  
+  1. If |navigable|'s [=navigable/traversable navigable=] is not a [=fenced navigable
```
domfarolino, line 2215: Will this work for URN iframes, where the traversable navigable is not a fenced navigable? Or does the network revocation stuff not apply to them at all?

VergeA: Nope, see corresponding Chromium impl [here](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame_config.h;drc=c0883e36f0f65273f002c2ca8b7e9474256e00e4;l=498) that explictly prevents disabling network for URN iframes.

domfarolino (approved):

PR #206, opened 2025-01-10
[Spec] Network revocation patches for WebSocket and WebTransport APIs

blu25:
This ensures that the following things happen for both `WebTransport` and `WebSocket` objects:

- All open connections associated with a context are closed when untrusted network access is revoked.
- New connections cannot be created after network access is revoked.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/206.html" title="Last updated on Feb 19, 2025, 2:27 AM UTC (997d803)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/206/82a9f09...997d803.html" title="Last updated on Feb 19, 2025, 2:27 AM UTC (997d803)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -269,6 +269,18 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
 spec: turtledove; urlPrefix: https://wicg.github.io/turtledove/
   type: dfn
     text: construct a pending fenced frame config; url: construct-a-pending-fenced-frame-config
+spec: RFC6455; urlPrefix: https://datatracker.ietf.org/doc/html/rfc6455
+  type: dfn
+    text: fail the WebSocket connection; url: #section-7.1.7
+spec: WebSocket; urlPrefix: https://websockets.spec.whatwg.org/
+  type: dfn
+    text: establish a WebSocket connection; url: #concept-websocket-establish
+spec: WebTransport; urlPrefix: https://w3c.github.io/webtransport
+  type: dfn
+    for: WebTransport
+      text: cleanup; url: #webtransport-cleanup
+      text: queue a network task; url: #webtransport-queue-a-network-task
+    text: initialize WebTransport over HTTP; url: #initialize-webtransport-over-http
 </pre>
 
 <style>
@@ -2123,6 +2135,16 @@ Issue: This will require a RFC to add a test-only function to the WPT web driver
   1. [=set/Append=] |nonce| to the user agent's [=network revocation nonce set=].
 
   1. [=fetch group/terminated|Terminate=] |settings|'s [=fetch/fetch group=].
+
+  1. [=list/For each=] {{WebSocket}} object |webSocket| whose [=relevant settings object=] is
+     |settings|, run [=fail the WebSocket connection=] given |webSocket|.
+
+  1. [=list/For each=] {{WebTransport}} object |webTransport| whose [=relevant settings object=] is
+     |settings|, [=WebTransport/cleanup=] |webTransport| with a newly [=exception/create|created=]
+     {{WebTransportError}} whose {{WebTransportErrorOptions/source}} is `"session"`.
+
+     Note: Not passing in a {{WebTransportCloseInfo}} ensures that the {{WebTransport}} object is
+     set to the `"failed"` state rather than the `"closed"` state.
 </div>
 
 <div algorithm>
@@ -2161,7 +2183,41 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   (<a href="https://github.com/WICG/fenced-frame/issues/191">WICG/fenced-frame#191</a>)
 </div>
 
-The network revocation mechanism requires the following monkeypatches to the [[HTML]] Standard.
+<h3 id=disable-websocket>WebSocket monkeypatches for network revocation</h3>
+
+The network revocation mechanism requires the following monkeypatch to the [[WebSockets]]
+Standard.
+
+<div algorithm=establish-websocket-patch>
+  Modify the [=establish a WebSocket connection=] algorithm. Add a new step after step 10 that
+  reads:
+
+  11. If the result of running [=determine if a navigable has revoked network for itself=] given
+      <var ignore>client</var>'s [=environment settings object/global object=]'s
+      [=Window/navigable=] is true, [=fail the WebSocket connection=].
+</div>
+
+<h3 id=disable-webtransport>WebTransport monkeypatches for network revocation</h3>
+
+The network revocation mechanism requires the following monkeypatch to the [[WebTransport]]
+Standard.
+
+<div algorithm=initialize-webtransport-patch>
+  Modify the [=initialize WebTransport over HTTP=] algorithm. Rewrite step 5 (keeping all substeps
+  unchanged) to read:
+
+  5. If any of the following conditions are true:
+     
+     - Running [=should request be blocked by Content Security Policy?=] with |request| returns
+       **"Blocked"**;
+     - Running [=should request be blocked due to a bad port=] with |request| returns **"blocked"**;
+     - Running [=determine if a navigable has revoked network for itself=] given <var
+       ignore>client</var>'s [=environment settings object/global object=]'s [=Window/navigable=]
+       returns true;
+     
+     then abort the remaining steps and [=WebTransport/queue a network task=] with
+     <var ignore>transport</var> to run these steps:
+</div>
 
 <h3 id=new-request-destination>New [=request=] [=request/destination=]</h3>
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2123,6 +2135,16 @@ Issue: This will require a RFC to add a test-only function to the WPT web driver
   1. [=set/Append=] |nonce| to the user agent's [=network revocation nonce set=].
 
   1. [=fetch group/terminated|Terminate=] |settings|'s [=fetch/fetch group=].
+
+  1. [=list/For each=] {{WebSocket}} object |webSocket| whose [=relevant settings object=] is
+     |settings|, run [=fail the WebSocket connection=] given |webSocket|.
```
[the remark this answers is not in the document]
blu25: I had reached out to Mike Taylor about this since I was also unsure about how to grab the objects. Based on [existing precedent in the spec](https://html.spec.whatwg.org/#unloading-document-cleanup-steps) (fair warning: this is going to load the **entire** HTML spec, so it might take a minute for your computer to unfreeze), it's totally valid to say "`for each X object whose relevant * object is |theObject|`", so that's what I decided to use.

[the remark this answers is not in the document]
VergeA: TIL! Will go ahead and resolve this.

spec.bs
```diff
@@ -2198,16 +2254,21 @@ table](https://fetch.spec.whatwg.org/#destination-table) to illustrate that <{fe
...
-other specifications, which will use the below algorithm to determine if invocation can occur
+other specifications, which will use the first algorithm below to determine if invocation can occur
 successfully. One example of a method which is gated behind revocation of untrusted network access
 is {{SharedStorage/get()}} when invoked outside of a {{SharedStorageWorklet}}. This method is
 defined in the [[Shared-Storage]] draft specification.
 
+In the intermediate state where {{Fence/disableUntrustedNetwork()}} has been called but the returned
+{{Promise}} has not [=resolved=] yet, there are restrictions on features that involve making
+requests over the network. The second algorithm below is used to determine whether those features
+are allowed.
```
VergeA, line 2270: I think it might be good to mention that in this state, nested fenced frame trees may still have network access, and these restrictions will not apply to them until they also disable network. Hence why the algorithm below relies on the keyword "itself."

VergeA:
LGTM % 1 comment

spec.bs
blu25: Updated this section to call that out.

VergeA (approved):
Looks great, thanks!

spec.bs
```diff
@@ -2123,6 +2135,16 @@ Issue: This will require a RFC to add a test-only function to the WPT web driver
   1. [=set/Append=] |nonce| to the user agent's [=network revocation nonce set=].
 
   1. [=fetch group/terminated|Terminate=] |settings|'s [=fetch/fetch group=].
+
+  1. [=list/For each=] {{WebSocket}} object |webSocket| whose [=relevant settings object=] is
```
domfarolino, line 2139: This algorithm is called in parallel so I don't really think it's valid for us to be inspecting "each" WebIDL `WebSocket` object of a given settings object, and running algorithms on it that (a) HTML currently runs [as a task on the Document's event loop](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-documents:unloading-document-cleanup-steps:~:text=Assert%3A%20this%20is%20running%20as%20part%20of%20a%20task%20queued%20on%20oldDocument%27s%20relevant%20agent%27s%20event%20loop.), and (b) the WebSocket spec itself [only runs as proper event-loop-bound networking tasks](https://w3c.github.io/webtransport/#ref-for-webtransport-queue-a-network-task%E2%91%A0%E2%91%A5). So we should probably (a) figure out a better way to restructure this, and (b) add a step at the beginning of this algorithm to assert that this is running in parallel, to make it easier to catch this kind of stuff in the future.

blu25: I'm going to move this to the `disableUntrustedNetwork()` algorithm itself, before the parallel steps. I think that should work.

For WebSockets, the document unloading steps you linked calls the [make disappear](https://websockets.spec.whatwg.org/#make-disappear) algorithm, which directly calls [fail the websocket connection](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.7) without first moving into a different event loop.

For WebTransport, the [cleanup](https://w3c.github.io/webtransport/#webtransport-cleanup) algorithm is invoked by [close()](https://w3c.github.io/webtransport/#dom-webtransport-close), and it doesn't drop into a different event loop either.

Based on that, I think it should be fine to put it directly in the non-parallel section of `disableUntrustedNetwork()`.

domfarolino: How does this compare to the implementation?

blu25: The implementation checks for whether the nonce stored internally corresponds to one of the network revoked nonces. Nonces in the spec are [not well defined at the moment](https://fetch.spec.whatwg.org/#network-partition-key:~:text=Let%20secondKey%20be%20null%20or%20an%20implementation%2Ddefined%20value.). The assumption I'm making here is that there's a 1:1 ratio of nonces to fenced frame trees/windows. If we want to explicitly check the nonce value here, that will get more tricky.

I think we can check the nonce for WebTransport, since a [WebTransport session](https://www.w3.org/TR/webtransport/#protocol-webtransport-session) references an [underlying connection](https://www.w3.org/TR/webtransport/#underlying-connection), and from there you can just grab its [associated key's](https://fetch.spec.whatwg.org/#connection-key) second tuple value.

WebSockets are more tricky. The underlying connection it uses is [its own separate thing that's not shareable with the fetch connection object](https://websockets.spec.whatwg.org/#concept-websocket-connection-obtain:~:text=Although%20structured%20a%20little%20differently%2C%20carrying%20different%20properties%2C%20and%20therefore%20not%20shareable%2C%20a%20WebSocket%20connection%20is%20very%20close%20to%20identical%20to%20an%20%22ordinary%22%20connection.). They do have a [nonce being generated](https://websockets.spec.whatwg.org/#concept-websocket-connection-obtain:~:text=Let%20keyValue%20be%20a%20nonce%20consisting%20of%20a%20randomly%20selected%2016%2Dbyte%20value%20that%20has%20been%20forgiving%2Dbase64%2Dencoded%20and%20isomorphic%20encoded.) during the connection process, but I'm wasn't sure if that's actually one that would be usable for what we're trying to do.

As an alternative, I opted to not use nonces in the spec. If you think that is too much of a deviation let me know and I can try to think of a better way to do this. If that is the case, this will probably be either blocked on the existing partition nonce work, or will require extra nonce work that will be blocked on that work.

domfarolino: I think I'm more concerned about the timing of the revocation, and less about the nonces involved. I just want to make sure we have something in the implementation observable equivalent to the lines this PR is adding to `disableUntrustedNetwork()`.

blu25: The equivalent closings in the impl happen in [`NetworkContext::RevokeNetworkForNonces`](https://source.chromium.org/chromium/chromium/src/+/main:services/network/network_context.cc;l=3339-3346;drc=e93478ec3ffbb7b398ae484d7284c0ee046d48cf). That is ultimately invoked by [`RenderFrameHostImpl::DisableUntrustedNetworkInFencedFrame`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/render_frame_host_impl.cc;l=10272;drc=3deaac799dc39f0ba14200463fd31bba49eb72a6), which is called directly from the renderer (via [`Fence::disableUntrustedNetwork`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/fenced_frame/fence.cc;l=370;drc=a3be094f402eecf71710e03aa7e4423fb224bd74).

In the spec, the stuff whose equivalent is in `RevokeNetworkForNonces` is happening in parallel. But the intention of this is to close all network channels as soon as possible after `disableUntrustedNetwork()` is called. IIUC the main reason why this part of the impl is in parallel is because of the multiple IPC calls that need to happen. If the spec requires that the WebSocket/WebTransport calls happen on the document's event loop, I think it should functionally be the same.

domfarolino: As discussed offline, the script-observable effects of "making disappear" any sockets are asynchronous anyways (given the "hook" in https://websockets.spec.whatwg.org/#ref-for-section-7.1.4%E2%91%A3 and the forced task queueing there), and that forced asyncness gives us the necessary window required to let the implementation do _its_ async IPC process hopping before the `close` event is fired at the socket. So I think we're OK here ultimately.

spec.bs
```diff
@@ -269,6 +269,18 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
...
+  type: dfn
+    text: fail the WebSocket connection; url: #section-7.1.7
+spec: WebSocket; urlPrefix: https://websockets.spec.whatwg.org/
+  type: dfn
+    text: establish a WebSocket connection; url: #concept-websocket-establish
+spec: WebTransport; urlPrefix: https://w3c.github.io/webtransport
+  type: dfn
+    for: WebTransport
+      text: cleanup; url: #webtransport-cleanup
+      text: queue a network task; url: #webtransport-queue-a-network-task
```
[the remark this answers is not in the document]
blu25: I think we should keep this in place, mainly because the WebTransport spec uses ["queue a network task"](https://w3c.github.io/webtransport/#webtransport-queue-a-network-task) in the algorithm we're patching as well as a bunch of other spots, so I'd rather keep it as is.

spec.bs
```diff
@@ -2139,7 +2163,18 @@ Issue: This will require a RFC to add a test-only function to the WPT web driver
   1. Return <b>allowed</b>.
 </div>
 
-<h3 id=disable-fetch>Fetch monkeypatches for network revocation</h3>
+<h3 id=disable-monkeypatches>Monkeypatches for network revocation</h3>
+
+*This introductory section is non-normative.*
+
+The network revocation mechanism requires various patches to various standards. These standards
```
domfarolino, line 2170: How about this instead, if you agree with it:

The network revocation mechanism in this section requires patching standards that make network requests. Before our patches, these standards make network requests (such as initiate WebSocket connections) while unaware of the network revocation *status* of the context that ultimately initiated the request. Our patches add checks to the algorithm in those standards to consider the network revocation status of the initiating context before the request is made.

WDYT? I think it's more clear and less hand wavey.

blu25: That sounds good. I've updated this section to have that wording.

PR #208, opened 2025-01-16
Add demo section to local unpartitioned access explainer.

VergeA:
This section contains command line flags to enable the local unpartitioned data access functionality in Chromium, and links to the demo sites we created for TPAC. This will make it easier for interested folks to test out the personalized payments use case on Chrome Canary.


shivanigithub (approved):
lgtm, thanks!

PR #209, opened 2025-01-28
Check that document is fully active for window.fence methods

blu25:
This PR makes the following changes

- Add fully active document checks in `window.fence` methods as they appear in the associated implementation.
- Move the `{{SecurityError}}` check in `disableUntrustedNetwork()` to the newly added active document check.
- Have all instances of `throw` link to the [throw definition](https://webidl.spec.whatwg.org/#dfn-throw).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/209.html" title="Last updated on Feb 4, 2025, 7:15 PM UTC (13f22b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/209/1c78202...13f22b3.html" title="Last updated on Feb 4, 2025, 7:15 PM UTC (13f22b3)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1759,8 +1759,10 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 <div algorithm>
   The <dfn method for=Fence>reportEvent(|event|)</dfn> method steps are:
 
-  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then [=exception/throw=] a
-     {{SecurityError}} {{DOMException}}.
+  1. Let |document| be [=this=]'s [=relevant global object=]'s [=associated Document=].
+
+  1. If |document| is not [=Document/fully active=], then [=exception/throw=] a {{SecurityError}}
+     {{DOMException}}.
 
   1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
      [=browsing context/fenced frame config instance=].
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1759,6 +1759,9 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 <div algorithm>
   The <dfn method for=Fence>reportEvent(|event|)</dfn> method steps are:
 
+  1. If [=this=]'s {{Document}} is not [=Document/fully active=], then [=exception/throw=] a
```
VergeA, line 1762: Does `this` have a `Document`? I know it's referenced as such in the `notifyEvent` algorithm today, but elsewhere I see us accessing `[=this=]'s [=relevant global object=]'s [=associated Document=]`. So if those are referring to the same entity, probably good to keep them consistent.

VergeA (commented):
One question, otherwise looks good.

spec.bs
blu25: I think you're right on that. I've changed all these instances to using relevant global object's associated document.

VergeA (approved):

spec.bs
```diff
@@ -1680,11 +1680,11 @@ partial interface Navigator {
   1. Let |instance| be [=this=]'s [=relevant global object=]'s [=Window/browsing context=]'s
      [=browsing context/fenced frame config instance=].
 
-  1. If |instance| is null, then throw a {{DOMException}}.
+  1. If |instance| is null, then [=exception/throw=] a {{DOMException}}.
```
domfarolino, line 1683: Now that I look at this, we have to specify the specific type of DOMException. Can you make this match the implementation?

domfarolino (approved):

Issue #205, opened 2025-01-10
[Spec] Promise returned by disableUntrustedNetwork call is not resolved with ongoing navigation


xiaochen-z:
See [RenderFrameHostImpl::CalculateUntrustedNetworkStatus()](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/render_frame_host_impl.cc;bpv=1;bpt=1;l=10520?q=%22FrameTreeNode%20and%20it%20does%20not%20have%20ongoing%20navigation.%22&ss=chromium%2Fchromium%2Fsrc&gsn=CalculateUntrustedNetworkStatus&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fchromium.googlesource.com%2Fcodesearch%2Fchromium%2Fsrc%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dcontent%2Fbrowser%2Frenderer_host%2Frender_frame_host_impl.cc%23lyJwquzYhw0dDV1zI33ym6SBUllG5CgxfvA_ap5xKFc).

If any descendant frames have ongoing navigations, the promise returned by disableUntrustedNetwork call at the parent frame cannot be resolved.

Note if the frame itself has ongoing navigations, it is allowed to resolve the promise. At the time navigation commits, untrusted network status will be recalculated. If all descendant frames' navigations are done, the promise will then be resolved.

Issue #192, opened 2024-09-12
[Spec] Specify way to exempt URLs from network revocation


blu25:
This will require a change to the WPT web driver + an RFC opened for it.

domfarolino:
Can you describe why this is important / why we filed this? It isn't for web-exposed behavior right? Testing-only? (A brief description of how this came about / what PR prompted this would be good).

VergeA:
I can provide some more context:

Writing WPTs for fenced frames is hard, because we can't directly communicate data between the fenced frame and the test harness. This is by design, because fenced frames don't have access to communication channels like postMessage() that are commonly used in iframe WPTs. To get around this, we use a RemoteContext executor to send RPCs between the harness and the fenced frame over the network. This allow the test harness to observe behavior that occurs in the fenced frame, and assert on that behavior.

When a fenced frame disable network access, that will include requests sent to the RemoteContext URL used for testing. Meaning, we'll lose our last channel to observe fenced frame behavior in tests. The new `window.internals` API we added to Chrome was intended to exempt RemoteContext URLs from network revocation, so that we can still write tests against fenced frames.

In order to allow our WPTs to be truly cross-platform, we'd need to standardize an interface by which all browsers can "exempt URLs from network revocation" in tests. This is where the WPT web driver + RFC come in.

PR #210, opened 2025-02-10
Spec descendant frames' ongoing navigations block disableUntrustedNetwork promise from being resolved

xiaochen-z:
fix: #205

This PR only implements the change in algorithm "recalculate the untrusted network status of all fenced frame descendants".

The part that this algorithm should be called whenever a navigation commits will be done in a follow-up PR. I'm still investigating how to spec this. See previous [comment](https://github.com/WICG/fenced-frame/pull/176#discussion_r1881105124) on this.

***
<a href="https://pr-preview.s3.amazonaws.com/xiaochen-z/fenced-frame/pull/210.html" title="Last updated on Feb 25, 2025, 2:37 PM UTC (ced0b67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/210/fa4ec2c...xiaochen-z:ced0b67.html" title="Last updated on Feb 25, 2025, 2:37 PM UTC (ced0b67)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2069,12 +2069,10 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
      
         1. If |currentNavigable|' [=navigable/ongoing navigation=] is a [=navigation ID=]:
         
-            1. Let |ancestorNavigables| be |currentNavigable|'s [=navigable/active document=]'s 
-               [=Document/ancestor navigables=] with [=an-unfenced|unfenced=] set to false.
-               
-            1. If |ancestorNavigables| is [=list/is not empty=] and the last element of 
-               |ancestorNavigables| is a [=fenced navigable container/fenced navigable=], then 
-               [=set/append=] that element to |navigablesWithNetworkChildren|.
+            1. Let |ancestorFencedRoot| be |currentNavigable|'s [=navigable/traversable navigable=].
+
+            1. If |ancestorFencedRoot| is a [=fenced navigable container/fenced navigable=],
+               [=set/append=] |ancestorFencedRoot| to |navigablesWithNetworkChildren|.
         
         1. [=iteration/continue=].
 
--- a/spec.bs
+++ b/spec.bs
@@ -145,6 +145,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: historyHandling; url: navigation-hh
       text: referrerPolicy; url: navigation-referrer-policy
       text: attempt to populate the history entry's document; url: attempt-to-populate-the-history-entry's-document
+      text: navigation ID; url: navigation-id
       text: navigation params; url: navigation-params
       text: snapshot source snapshot params; url: snapshotting-source-snapshot-params
       text: the navigation must be a replace; url: the-navigation-must-be-a-replace
@@ -2064,14 +2065,25 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
      1. Let |currentNavigable| be the result of [=stack/pop|popping=] from |navigables|.
 
-     1. If |currentNavigable| is not a [=fenced navigable container/fenced navigable=], then
-        [=iteration/continue=].
+     1. If |currentNavigable| is not a [=fenced navigable container/fenced navigable=]:
+     
+        1. If |currentNavigable|' [=navigable/ongoing navigation=] is a [=Navigation Id=]:
+        
+            1. Let |ancestorNavigables| be |currentNavigable|'s [=navigable/active document=]'s 
+               [=Document/ancestor navigables=] with [=an-unfenced|unfenced=] set to false.
+               
+            1. If |ancestorNavigables| is [=list/is not empty=] and the last element of 
+               |ancestorNavigables| is a [=fenced navigable container/fenced navigable=], then 
+               [=set/append=] that element to |navigablesWithNetworkChildren|.
+        
+        1. [=iteration/continue=].
 
      1. Let |config| be |currentNavigable|'s [=navigable/active browsing context=]'s [=browsing
         context/fenced frame config instance=].
 
      1. If |config|'s [=fenced frame config instance/untrusted network status=] is [=untrusted
-        network status/disabled for this tree and fenced subtrees=], then [=iteration/continue=].
+        network status/disabled for this tree and fenced subtrees=] and |currentNavigable|' 
+        [=navigable/ongoing navigation=] is not a [=Navigation Id=], then [=iteration/continue=].
 
      1. Let |networkCutoffReady| be true if |navigablesWithNetworkChildren| does not [=set/contain=]
         |currentNavigable| and |config|'s [=fenced frame config instance/untrusted network status=]
@@ -2099,7 +2111,9 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
         1. [=map/Clear=] |config|'s [=fenced frame config instance/on network disabled promises=].
 
-     1. Otherwise:
+     1. If |networkCutoffReady| is false or |currentNavigable|' [=navigable/ongoing navigation=] is 
+        a [=Navigation Id=]:
+        
         1. Let |ancestorFencedRoot| be |currentNavigable|'s [=traversable navigable/unfenced
            parent=]'s [=navigable/traversable navigable=].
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2064,14 +2065,23 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
      1. Let |currentNavigable| be the result of [=stack/pop|popping=] from |navigables|.
 
-     1. If |currentNavigable| is not a [=fenced navigable container/fenced navigable=], then
-        [=iteration/continue=].
+     1. If |currentNavigable| is not a [=fenced navigable container/fenced navigable=]:
+     
+        1. If |currentNavigable|' [=navigable/ongoing navigation=] is a [=navigation ID=]:
```
blu25, line 2070: Do we need to care if this is set to [`"traversal"`](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-history-step:~:text=for%20navigable%20to-,%22traversal%22,-.)?

spec.bs
```diff
@@ -2099,7 +2109,9 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
         1. [=map/Clear=] |config|'s [=fenced frame config instance/on network disabled promises=].
 
-     1. Otherwise:
+     1. If |networkCutoffReady| is false or |currentNavigable|' [=navigable/ongoing navigation=] is 
+        a [=navigation ID=]:
```
blu25, line 2113: We're checking if `|currentNavigable|`'s ongoing navigation is a navigation ID 3 times in this loop. It should probably be pulled out into its own variable.

xiaochen-z: That's a good idea, I've made it a variable.

spec.bs
xiaochen-z: The HTML spec has this line:

> Any attempts to navigate a [navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable) that is currently [traversing](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-traverse-history-step) are ignored.

So I assume `ongoing navigation is a navigation ID` means there is an ongoing navigation. Otherwise if it is "traversal" or null, there is no ongoing navigation.

blu25: It's set to "traversal" during the [history step](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-history-step). I'm not sure if we need to account for the case where a fenced frame somehow gets into some history navigation, and then an ancestor fenced frame disables its network while the child frame is still in the middle of the history navigation. (I'm also not sure if a fenced frame can get into this state.) In other words, would it be safer to just say "the ongoing navigation must be null" so we account for actual navigations as well as history navigations?

xiaochen-z: Changed to `|ongoingNavigation| is not null` and `|ongoingNavigation| is null`. Removed navigation ID reference.

blu25 (approved):
LGTM, thanks!

xiaochen-z:
+@domfarolino for review, thank you.

spec.bs
```diff
@@ -2099,7 +2110,8 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
 
         1. [=map/Clear=] |config|'s [=fenced frame config instance/on network disabled promises=].
 
-     1. Otherwise:
+     1. If |networkCutoffReady| is false or |ongoingNavigation| is not null:
```
domfarolino, line 2113:
```suggestion
     1. If |ongoingNavigation| is not null:
```

Isn't this the same, since at this point we already know that `|networkCuttoffReady|` is false?

domfarolino (commented):
Pretty much lgtm. This covers traversals as well btw, which I think is the goal too? Does that match the implementation? And are there WPTs?

spec.bs
xiaochen-z: The above if branch does not early return. Also the line `Otherwise:` is removed. So at this point `|networkCuttoffReady|` can be true or false here.

Even if we add the early return above, the condition `|networkCutoffReady| is false` is still needed. As the suggested edit, IIUC when `|networkCutoffReady|` is false but `|ongoingNavigation|` is null, the branch will not be entered, however it should.

xiaochen-z:
Thanks!

> This covers traversals as well btw, which I think is the goal too? Does that match the implementation? 

From what I read from the spec, when `ongoing navigation` is "traversal", it means the navigable is traversing its history. For example, FF going back to a previously navigated URL. The FF may end up with a different page and having its untrusted network status reset to enabled.

This is the race condition that the PR targets: When a parent frame tries to disable untrusted networks, it checks the child frame and finds it has untrusted network disabled. Without checking descendant frames' ongoing navigations, the parent frame goes ahead marking its untrusted network also disabled. But if the child frame has an ongoing navigation, including the history traversal, it resets the child frame's untrusted network to enabled as the frame navigates to a different page.

The implementation checks this via [FrameTreeNode::HasNavigation()](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/frame_tree_node.cc;drc=798b98d70313e6a55bcf9cc85bc7ca7d42ca6d23;bpv=1;bpt=1;l=969?gsn=HasNavigation&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fchromium.googlesource.com%2Fcodesearch%2Fchromium%2Fsrc%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dcontent%2Fbrowser%2Frenderer_host%2Fframe_tree_node.cc%23r5U9RBP1DD4eTeCRnEua5QnpHy13mpIjbZOvTo2U6g0), which is set during normal navigation as well as history traversal. So the spec matches the implementation.

> And are there WPTs?

No. I'm adding WPTs for this. Tracking in bug: [crbug/397377177](https://issues.chromium.org/u/1/issues/397377177).

xiaochen-z:
WPTs added in https://chromium-review.googlesource.com/c/chromium/src/+/6280958. However they were added to wpt_internal/ where other network revocation WPTs live. I suppose it is not possible to link internal WPTs in the spec?

blu25:
> WPTs added in https://chromium-review.googlesource.com/c/chromium/src/+/6280958. However they were added to wpt_internal/ where other network revocation WPTs live. I suppose it is not possible to link internal WPTs in the spec?

On a separate PR, @domfarolino had suggested offline to add a [non-normative section that links to the internal WPT directory](https://github.com/WICG/fenced-frame/pull/206/commits/fd1e038ee95d516377be167278ef438eae9a27e4) with an explanation of why they're internal. Not sure if it's worth doing here or just adding the WPTs later once we move them to the external directory.

spec.bs
domfarolino: Ah I see.

domfarolino:
I think that'd be good.

xiaochen-z:
Thanks! There is an [issue block](https://github.com/WICG/fenced-frame/pull/210/files#diff-6f5a1d8263b0b0c42e2716ba5750e3652e359532647ac934c1c70086ae3ceddaL2132) below already stated that the internal WPTs will be upstreamed and listed there once the required test-only function is added to the WPT web driver.

domfarolino:
Sure but maybe this PR can link to the specific test that's relevant to the changes made in this patch.

spec.bs
```diff
@@ -2121,12 +2132,17 @@ Several APIs specific to fenced frames are defined on the {{Fence}} interface.
...
+        
         1. Let |ancestorFencedRoot| be |currentNavigable|'s [=traversable navigable/unfenced
            parent=]'s [=navigable/traversable navigable=].
 
         1. If |ancestorFencedRoot| is a [=fenced navigable container/fenced navigable=],
            [=set/append=] |ancestorFencedRoot| to |navigablesWithNetworkChildren|.
+           
+  Note: The Chromium-internal web platform test is available at <a
+  href="https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/wpt_internal/fenced_frame/disable-untrusted-network-ongoing-navigation.https.html">third_party/blink/web_tests/wpt_internal/fenced_frame/disable-untrusted-network-ongoing-navigation.https.html</a>.
+
```
domfarolino, line 2145: Please also add some text noting that this will be upstreamed with the associated issues preventing this are resolved. I know that kinda exists elsewhere but let's just allude to it here so people reading the spec don't think this is the expected end result.

xiaochen-z: Add that this will be upstreamed once #192 is resolved.

Build failure will be resolved once #212 is fixed, which is unrelated to this PR.

bikeshed allows issue linking using a much simpler syntax: https://speced.github.io/bikeshed/#remote-issues. However, it does not work on my end. Tracking in TODO: #213

PR #216, opened 2025-02-26
[Spec] Network revocation patches for Reporting API and Signed Exchange Reports

blu25:
This spec patch does the following:

- Modifies the "generate and queue a report" to early return if invoked from documents within a fenced frame tree with its network access revoked.
- Modifies the Webpackage signed exchange reporting API to only queue a report if the request's associated partition nonce is not one of the revoked nonces.
  - Note that this spec is on hold and most likely not going to become a standard, but we are still adding this patch in to cover all our bases.
- Modifies the "must be blocked due to a revoked partition nonce" algorithm to look for the nonce and URL in the request object, rather than expecting one to be provided.
  - This should be fine because the algorithm that [pulls the nonce from the request object](https://fetch.spec.whatwg.org/#request-determine-the-network-partition-key) ultimately gets that from the [request's associated environment](https://wicg.github.io/anonymous-iframe/#spec-network-partition-key). That nonce is set [when the environment is first created](https://wicg.github.io/anonymous-iframe/#spec-navigation-partition-nonce), and we have a patch to [conditionally set it to the fenced frame's network revocation nonce](https://github.com/WICG/fenced-frame/pull/149).
- Makes the "nonce" addition to the [network partition key definition in the credentialless iframes spec](https://wicg.github.io/anonymous-iframe/#spec-network-partition-key) more well defined.

https://github.com/WICG/fenced-frame/pull/149 should be merged in before this PR so that the logic that sets the environment's partition nonce to the fenced frame nonce is in place.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/pull/216.html" title="Last updated on Mar 25, 2025, 1:38 PM UTC (e099656)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/216/01a8212...e099656.html" title="Last updated on Mar 25, 2025, 1:38 PM UTC (e099656)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -221,6 +221,8 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: queue a cross-origin embedder policy CORP violation report; url: queue-a-cross-origin-embedder-policy-corp-violation-report
     text: should request be blocked due to a bad port; url: block-bad-port
+    for: request
+      text: determine the network partition key; url: request-determine-the-network-partition-key
     for: response
       text: has-cross-origin-redirects; url: response-has-cross-origin-redirects
 spec: mixed-content; urlPrefix: https://w3c.github.io/webappsec-mixed-content/
@@ -2178,7 +2180,8 @@ Once that web driver changes is made, existing Chromium-internal web platform te
   To determine whether fetching a [=request=] |request| <dfn>must be blocked due to
   a revoked partition nonce</dfn>, run these steps:
 
-  1. Let |nonce| be the result of running [=determine the network partition key=] on |request|.
+  1. Let |nonce| be the result of running [=request/determine the network partition key=] on
+     |request|.
 
   1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], and if
      |request|'s [=request/URL=] [=set/exists=] in it, return <b>allowed</b>.
--- a/spec.bs
+++ b/spec.bs
@@ -4425,7 +4425,7 @@ Add the following algorithm:
 </div>
 
 <div algorithm=network-partition-key-patch>
-  Rewrite the [=network partition key=] redefinition to read:
+  Rewrite the [=network partition key=] <a href=https://wicg.github.io/anonymous-iframe/#spec-network-partition-key>redefinition</a> to read:
 
   A [=network partition key=] is a [=tuple=] consisting of:
 
--- a/spec.bs
+++ b/spec.bs
@@ -4429,9 +4429,9 @@ Add the following algorithm:
 
   A [=network partition key=] is a [=tuple=] consisting of:
 
-  - a [=site=].
-  - null or an [=implementation-defined=] value.
-  - a <dfn for="network partition key">nonce</dfn> (null or an [=implementation-defined=] value).
+  * a [=site=].
+  * null or an [=implementation-defined=] value.
+  * a <dfn for="network partition key">nonce</dfn> (null or an [=implementation-defined=] value).
 </div>
 
 <div algorithm="create navigation params by fetching">
--- a/spec.bs
+++ b/spec.bs
@@ -269,6 +269,9 @@ spec: attribution-reporting; urlPrefix: https://wicg.github.io/attribution-repor
 spec: turtledove; urlPrefix: https://wicg.github.io/turtledove/
   type: dfn
     text: construct a pending fenced frame config; url: construct-a-pending-fenced-frame-config
+spec: webpackage; urlPrefix: https://wicg.github.io/webpackage/loading.html
+  type: dfn
+    text: wait and queue a report for; url: wait-and-queue-a-report-for
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2172,12 +2177,14 @@ Once that web driver changes is made, existing Chromium-internal web platform te
 </div>
 
 <div algorithm>
-  To determine whether fetching a [=request=] <var ignore>request</var> <dfn>must be blocked due to
-  a revoked partition nonce</dfn> using a [=fenced frame config instance/partition nonce=] |nonce|
-  and a [=URL=] |requestURL|, run these steps:
+  To determine whether fetching a [=request=] |request| <dfn>must be blocked due to
+  a revoked partition nonce</dfn>, run these steps:
+
+  1. Let |nonce| be the result of running [=request/determine the network partition key=] on
```
VergeA, line 2183: Does this algorithm actually return a nonce? In the fetch spec, it looks like it returns a [tuple](https://fetch.spec.whatwg.org/#network-partition-keys), but there's a `secondKey` that can be an implementation-defined value. Is that what you're referring to here?

I did a quick search of the FF spec and I don't see any place where we patch in a nonce.

VergeA: Ah I just double checked and saw that the credentialless iframe spec includes the nonce in the partition key. I still think you need to pull it out of the tuple here, tho.

blu25: Done. The tuple was not well defined in the credentialless iframe spec (what is a 'nonce'?) so I added a patch to redefine it here.

VergeA (approved):
Looking good now, thanks!

spec.bs
```diff
@@ -2209,12 +2214,16 @@ Once that web driver changes is made, existing Chromium-internal web platform te
 </div>
 
 <div algorithm>
-  To determine whether fetching a [=request=] <var ignore>request</var> <dfn>must be blocked due to
-  a revoked partition nonce</dfn> using a [=fenced frame config instance/partition nonce=] |nonce|
-  and a [=URL=] |requestURL|, run these steps:
+  To determine whether fetching a [=request=] |request| <dfn>must be blocked due to
+  a revoked partition nonce</dfn>, run these steps:
```
domfarolino, line 2218: I'm guessing this can be unwrapped more optimally on the previous line now? Try running `specfmt`, on the latest version.

spec.bs
```diff
@@ -2288,6 +2297,29 @@ Standard.
...
+
+The network revocation mechanism requires the following monkeypatch to the [[Webpackage]]
+Standard.
+
+<div algorithm=sxg-patch>
+  Modify the [monkeypatch](https://wicg.github.io/webpackage/loading.html#mp-http-fetch) to [=HTTP
+  fetch=]. Rewrite step 4 under the **"b2" or "b3"** switch case to read:
+
+  4. If the result of determining whether <var ignore>request</var> [=must be blocked due to a
+      revoked partition nonce=] is false, then [=in parallel=], [=wait and queue a report for=] <var
```
domfarolino, line 2319: Indent is wrong here. specfmt's latest version should work here.

spec.bs
```diff
@@ -4395,6 +4427,16 @@ Add the following algorithm:
...
 </div>
 
+<div algorithm=network-partition-key-patch>
+  Rewrite the [=network partition key=] redefinition to read:
+
+  A [=network partition key=] is a [=tuple=] consisting of:
+
+  - a [=site=].
+  - null or an [=implementation-defined=] value.
+  - A <dfn for="network partition key">nonce</dfn> (null or an [=implementation-defined=] value).
```
domfarolino, line 4437: Be consistent with capitals in this list. Also: are there other specs that monkeypatch the same definition to add a nonce? If so, we should name this "fenced frame nonce". Or maybe we should rename this regardless, to future-proof it.

spec.bs
blu25: Ran specfmt on the file.

spec.bs
blu25: I'll make it lowercase to match the convention we use in the fenced frame spec and to match the case of "null".

The [iframe credentialless spec](https://wicg.github.io/anonymous-iframe/#spec-network-partition-key) introduces and uses this nonce, but none of the items in its network partition key definition are named. I'm just introducing a name to make it easier to get in the spec. Really this is a fenced frame or iframe credentialless nonce. Alternatively, I could also keep it unnamed, but then I'd need to call `|network partition key|[2]` to get access to it, but that seems harder to read.

If we want to name all 3 fields, it would probably be `site`/`top level site`, `second key`, and `nonce`/`credentialless or fenced frame nonce` respectively.

Let me know what you'd prefer.

domfarolino: Explicit naming sounds good! Let's just link to https://wicg.github.io/anonymous-iframe/#spec-network-partition-key from "redefinition" from clarity, just with a normal anchor tag.

domfarolino (approved):
LGTM % the one open comment

spec.bs
blu25: Kept the explicit naming with the nonce, and added a link.

PR #219, opened 2025-03-07
link dns prefetch network revocation patch

### LINKED ISSUE #217: [Spec] link dns-prefetch network revocation patch
See implementation: [NetworkContext::ResolveHost](https://source.chromium.org/chromium/chromium/src/+/main:services/network/network_context.cc;bpv=1;bpt=1;l=1974?q=resolvehost%20network_context.cc&ss=chromium%2Fchromium%2Fsrc&gsn=ResolveHost&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fchromium.googlesource.com%2Fcodesearch%2Fchromium%2Fsrc%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dservices%2Fnetwork%2Fnetwork_context.cc%23OPWI2PFkPbWoChkaRb_PHxSnYAJdbd1CCPRvOdRozBk)

link dns-prefetch is blocked if the request's nonce is marked as network revoked.

xiaochen-z:
Fix #217.

***
<a href="https://pr-preview.s3.amazonaws.com/xiaochen-z/fenced-frame/pull/219.html" title="Last updated on Mar 27, 2025, 6:30 PM UTC (9cba8a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/219/9e3e56c...xiaochen-z:9cba8a9.html" title="Last updated on Mar 27, 2025, 6:30 PM UTC (9cba8a9)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2264,7 +2264,8 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   
   1. Let |nonce| be |key|'s [=network partition key/nonce=].
   
-  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network revocation exemption map=][|nonce|][|url|] does not exist, then:
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
+     revocation exemption map=][|nonce|][|url|] does not exist, then:
      
       1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return.
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -2277,7 +2277,7 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
      revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
      
-    1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return
+     1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return
        failure.
 </div>
 
--- a/spec.bs
+++ b/spec.bs
@@ -2272,7 +2272,7 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   
   Insert these steps immediately after step 2:
   
-  1. Let |nonce| be |key|'s [=network partition key/nonce=].
+  1. Let |nonce| be <var ignore>key</var>'s [=network partition key/nonce=].
   
   1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
      revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
--- a/spec.bs
+++ b/spec.bs
@@ -2245,7 +2245,8 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   
   1. Let |nonce| be |key|'s [=network partition key/nonce=].
   
-  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=] and |url| does not [=set/exist=] in it, then:
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=] and |url| does
+     not [=set/exist=] in it, then:
      
       1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, return.
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -2265,7 +2265,7 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   1. Let |nonce| be |key|'s [=network partition key/nonce=].
   
   1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
-     revocation exemption map=][|nonce|][|url|] does not exist, then:
+     revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
      
       1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return.
 </div>
--- a/spec.bs
+++ b/spec.bs
@@ -2258,7 +2258,7 @@ The network revocation mechanism requires the following monkeypatches to the [[F
 </div>
 
 <div algorithm=resolve-an-origin-patch>
-  Modify the [=resolve an origin=] algorithm to include an extra parameter: an [=url=] |url|.
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -2256,6 +2257,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   (<a href="https://github.com/WICG/fenced-frame/issues/191">WICG/fenced-frame#191</a>)
 </div>
 
+<div algorithm=resolve-an-origin-patch>
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: an [=url=] |url|.
```
domfarolino, line 2261:
```suggestion
  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
```

spec.bs
```diff
@@ -2256,6 +2257,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  
+  Insert these steps immediately after step 2:
+  
+  1. Let |nonce| be |key|'s [=network partition key/nonce=].
+  
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=] and |url| does
+     not [=set/exist=] in it, then:
```
domfarolino, line 2268: I would make this more clear and verbose by saying if NREM[nonce] exists, but NREM[nonce][url] does not exist, then....

spec.bs
```diff
@@ -2256,6 +2257,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+     
+      1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, return.
```
domfarolino, line 2270:
```suggestion
      1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return.
```

spec.bs
```diff
@@ -2256,6 +2257,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
+  
+  Insert these steps immediately after step 2:
+  
+  1. Let |nonce| be |key|'s [=network partition key/nonce=].
+  
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
+     revocation exemption map=][|nonce|][|url|] does not exist, then:
+     
+      1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return.
```
domfarolino, line 2270: This step is indented one space too deep.

spec.bs
```diff
@@ -2256,6 +2257,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
 
+<div algorithm=resolve-an-origin-patch>
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
+  
+  Insert these steps immediately after step 2:
+  
+  1. Let |nonce| be |key|'s [=network partition key/nonce=].
+  
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
+     revocation exemption map=][|nonce|][|url|] does not exist, then:
```
domfarolino, line 2268:
```suggestion
     revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
```

spec.bs
```diff
@@ -2266,6 +2267,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
+  
+  Insert these steps immediately after step 2:
+  
+  1. Let |nonce| be <var ignore>key</var>'s [=network partition key/nonce=].
+  
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
+     revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
+     
+    1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return.
```
blu25, line 2280: The [caller of `resolve an origin`](https://fetch.spec.whatwg.org/#ref-for-resolve-an-origin%E2%91%A1) is expecting some return value, so we can probably use "failure" since it already exists in the algorithm and the caller is set up to handle it.

```suggestion
    1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return failure.
```

spec.bs
```diff
@@ -2266,6 +2267,19 @@ The network revocation mechanism requires the following monkeypatches to the [[F
   (<a href="https://github.com/WICG/fenced-frame/issues/191">WICG/fenced-frame#191</a>)
 </div>
 
+<div algorithm=resolve-an-origin-patch>
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
+  
+  Insert these steps immediately after step 2:
```
blu25, line 2273: Putting this after step 2 means that IP addresses and localhost will not be subject to the nonce revocation checks. Is that intended?

xiaochen-z: Yeah this is the place where I'm not sure about.

For IP addresses and localhost, the host resolution is done locally: `HostResolverManager::ResolveLocally()`. So they are not making any network requests. And the spec also seems to say so. Ideally they should be allowed to resolve regardless of the untrusted network status.

However, the local resolution (and other host resolution logics) is done in `net/` layer, where untrusted network status info is unavailable. So in the implementation, the untrusted network status check is put in the upstream `NetworkContext::ResolveHost()`. That means IP addresses and localhost, if not exempted, will just be blocked from host resolution, even if they actually do not need to make any request to DNS.

I think for now it is best to match the spec with our implementation. I have updated the insertion to before step 1.

spec.bs
```diff
@@ -2266,6 +2267,20 @@ The network revocation mechanism requires the following monkeypatches to the [[F
...
+  Modify the [=resolve an origin=] algorithm to include an extra parameter: a [=URL=] |url|.
+  
+  Insert these steps before step 1:
+  
+  1. Let |nonce| be <var ignore>key</var>'s [=network partition key/nonce=].
+  
+  1. If the user agent's [=network revocation exemption map=][|nonce|] [=map/exists=], but [=network
+     revocation exemption map=][|nonce|][|url|] does not [=map/exist=], then:
+     
+    1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return
```
blu25, line 2280:
```suggestion
     1. If the user agent's [=network revocation nonce set=] [=set/contains=] |nonce|, then return
```
Add 1 extra space so the `1.` lines up with the start of the text on the line above.

blu25 (approved):
LGTM % formatting nit

PR #222, opened 2025-03-12
Embedder initiated navigation network revocation patch

### LINKED ISSUE #221: [Spec] Embedder-initiated navigation is not allowed when network is revoked
See implementation: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=1314-1330?q=%22Navigations%20cannot%20be%20initiated%20from%20a%22&ss=chromium%2Fchromium%2Fsrc

The behavior is if you have a nested iframe inside a fenced frame. The fenced frame disables its untrusted network, then attempts to start an embedder-initiated navigation of the nested iframe. The check linked above will be hit and the navigation will not be inititated.

xiaochen-z:
Fix #221.

***
<a href="https://pr-preview.s3.amazonaws.com/xiaochen-z/fenced-frame/pull/222.html" title="Last updated on Mar 24, 2025, 6:44 PM UTC (616e774)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/fenced-frame/222/c8ae373...xiaochen-z:616e774.html" title="Last updated on Mar 24, 2025, 6:44 PM UTC (616e774)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -3422,7 +3422,8 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
   
   Add a step after step 5 of [[HTML]]'s [=navigate=] algorithm that reads:
   
-  6. If the result of running [=determine if a navigable has revoked network for itself=] given <var ignore>sourceDocument</var>'s [=node navigable=] is true, then return.
+  6. If the result of running [=determine if a navigable has revoked network for itself=] given <var
+     ignore>sourceDocument</var>'s [=node navigable=] is true, then return.
 
   Modify step 13 of [[HTML]]'s [=navigate=] algorithm ("If all of the following are true:") to
   include the following condition:


spec.bs
```diff
@@ -3419,6 +3419,11 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 <div algorithm=navigate>
   Modify [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an optional [=string=]
   <dfn for=navigate>sharedStorageContext</dfn> (default null).
+  
+  Add a step after step 5 of [[HTML]]'s [=navigate=] algorithm that reads:
+  
+  6. If the result of running [=determine if a navigable has revoked network for itself=] given <var
```
domfarolino, line 3425: The decision here is made on the main thread then, right? Where as in the Chromium implementation, it is made in parallel (i.e., the browser process). Is there an observable difference?

I guess this is consistent with what we do when an embedder whose network has been revoked tries to create and navigate a fenced frame that _it_ owns: https://wicg.github.io/fenced-frame/#ref-for-fenced-frame-config-instance-untrusted-network-status. I'm curious, how does _that_ compare to the Chromium implementation? Does the spec currently diverge from the Chromium implementation in that case?

xiaochen-z: Sorry for the confusion. The CL linked in the issue is not what this spec is for. I've updated issue description.

> The decision here is made on the main thread then, right? Where as in the Chromium implementation, it is made in parallel (i.e., the browser process). Is there an observable difference?

Yes, you are right. The [step 5 in the navigate algorithm](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate) "allowed by sandboxing to navigate" is implemented in [renderer/core/frame/local_frame.cc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/local_frame.cc;l=2021?q=%22%23allowed-to-navigate%22&ss=chromium%2Fchromium%2Fsrc). But what we try to spec is implemented in [content/browser/renderer_host/navigation_request.cc](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=1314-1330?q=%22Navigations%20cannot%20be%20initiated%20from%20a%22&ss=chromium%2Fchromium%2Fsrc).

In the end the navigation is not initiated if the network is revoked so I don't think there is an observable difference. Should the added step be labelled as run in parallel?

> how does that compare to the Chromium implementation?

This spec:
* Scenario: nested iframe in a fenced frame, embedder initiates navigation of the nested iframe
* Check implementation: [content/browser/renderer_host/navigation_request.cc](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=1314-1330?q=%22Navigations%20cannot%20be%20initiated%20from%20a%22&ss=chromium%2Fchromium%2Fsrc)
* Behavior: Navigation will not be initiated. Console error: "Navigations cannot be initiated from a fenced frame after its network has been disabled."

The case you referenced:
* Scenario: nested fenced frame in a fenced frame, embedder initiates navigation of the nested fenced frame
* Check implementation: [content/browser/fenced_frame/fenced_frame.cc](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/fenced_frame/fenced_frame.cc;l=112-135?q=fenced_frame.cc&ss=chromium%2Fchromium%2Fsrc)
* Behavior: Navigation will not be initiated. Console error: "Embedder-initiated navigations of fenced frames are not allowed after the embedder's network has been disabled."

domfarolino: > In the end the navigation is not initiated if the network is revoked so I don't think there is an observable difference. Should the added step be labelled as run in parallel?

I think I agree that it is not observably different, so that should be OK.

And OK thanks for clarifying the scenarios, I think that makes sense.

xiaochen-z: Thanks. Marked as run in parallel.

spec.bs
```diff
@@ -3419,6 +3419,11 @@ CORP violation report=] algorithm, as leaving it unfenced may cause a privacy le
 <div algorithm=navigate>
   Modify [[HTML]]'s [=navigate=] algorithm to include an extra parameter: an optional [=string=]
   <dfn for=navigate>sharedStorageContext</dfn> (default null).
+  
+  Add a step after step 5 of [[HTML]]'s [=navigate=] algorithm that reads:
+  
+  6. [=In parallel=], if the result of running [=determine if a navigable has revoked network for
+     itself=] given <var ignore>sourceDocument</var>'s [=node navigable=] is true, then return.
```
domfarolino, line 3426: This isn't doing what you want it to do. If you return from in parallel, you just return from in parallel. It doesn't impact anything in the caller explicitly. I think in https://github.com/WICG/fenced-frame/pull/222#discussion_r1997283365 I meant that you can just not run it in parallel if there is no observable difference. Either that, or it might be preferred to wait until `#navigate` goes in parallel and then run this check. I think it's probably fine here though, just without parallelism.

xiaochen-z: Got it. Dropped `In parallel`.

domfarolino (approved):

